Issue #11, opened 2017-11-05
Missing dependencies


EiffL:
The setup.py script doesn't list all dependencies, in particular, I'm running into the following:
```
Traceback (most recent call last):
  File "/opt/desc/run.py", line 43, in <module>
    Stage.main()
  File "/opt/lsst/software/stack/python/miniconda3-4.2.12/lib/python3.5/site-packages/descpipe/stage.py", line 19, in main
    stage.run()
  File "/opt/desc/run.py", line 26, in run
    import desc.wlmassmap.mocks.extract_footprint
  File "/opt/lsst/software/stack/python/miniconda3-4.2.12/lib/python3.5/site-packages/desc/wlmassmap/mocks/extract_footprint.py", line 3, in <module>
    from GCRCatalogs import load_catalog
  File "/opt/lsst/software/stack/python/miniconda3-4.2.12/lib/python3.5/site-packages/GCRCatalogs/__init__.py", line 7, in <module>
    from .dc1 import *
  File "/opt/lsst/software/stack/python/miniconda3-4.2.12/lib/python3.5/site-packages/GCRCatalogs/dc1.py", line 8, in <module>
    import pymssql
ImportError: No module named 'pymssql'

```

PR #12, opened 2017-11-05
Added missing dependencies in setup.py

EiffL:
This pull request is intended to fix issue #11


yymao:
#11 and #12 are fixed.

Issue #13, opened 2017-11-06
remove `*_any` quantities


yymao:
Currently we have quantities like `magerr_u_any`, which were introduced so that users who do not care using the exactly same quantity among different catalogs would be able to grab the quantity without worrying which quantity exists in which catalog.

Since [GCR 0.2.1](https://github.com/yymao/generic-catalog-reader/releases/tag/v0.2.1), a new (and arguable better) way to do this is to use the `first_available()` method. 

So, instead of doing:

    gc1.get_quantities(['magerr_u_any', 'magerr_g_any'])
    gc2.get_quantities(['magerr_u_any', 'magerr_g_any'])

One can now do:

    mag_u_acceptable = ['magerr_u_lsst', 'magerr_u_sdss', 'magerr_u_des']
    mag_g_acceptable = ['magerr_g_lsst', 'magerr_g_sdss', 'magerr_g_des']
    gc1.get_quantities([gc1.first_available(mag_u_acceptable), gc1.first_available(mag_g_acceptable)])
    gc2.get_quantities([gc2.first_available(mag_u_acceptable), gc2.first_available(mag_g_acceptable)])

It is true that the code in the new version would be *longer*. But it is more explicit and does not require the reader to define those `*_any` quantities. 

@j-dr @evevkovacs @dkorytov @EiffL What do you think?  Any comments/suggestions?  If that looks good I'll start to remove `*_any` quantities in the existing readers.

evevkovacs:
I think this new method is better. It allows the user to define what the possibilities can be. Not to add more work, but I am assuming that the list of acceptable quantities is prioritized? ie above it asks for lsst first, then failing that sdss, then failing that des?

yymao:
Yes, hence the name `first_available`. This way, the user can decide the fallback sequence they need.

EiffL:
I like it, my only worry would be that you don't know exactly which magnitude is being used, a printed or log message would be a nice feedback to the user in case the first choice is not available.

yymao:
@EiffL, that's a good point. The question is whether this function should always print that information out, since the user can always add a line to check? Thoughts?

evevkovacs:
The user specifies and ordered list (that's why I asked), so by looking at 
the quantities list it is easy to see what's available. That said, a 
message would be a nice convenience.


On Mon, 6 Nov 2017, Francois Lanusse wrote:

> Date: Mon, 6 Nov 2017 13:46:36
[...]

PR #14, opened 2017-11-06
remove '*_any' quantities

yymao:
As discussed in #13, this PR removes `*_any` quantities in the readers.

@j-dr @dkorytov @evevkovacs @EiffL can you take a look?


yymao:
@evevkovacs we also need to update the scheme google doc accordingly.

evevkovacs:
The DESC Schema google doc is updated to remove the _any variables. I took a new snapshot and updated the catalog announcement page with the new snapshot.

j-dr:
Looks good to me.

PR #19, opened 2017-11-14
Sed fitting

danielsf:
- Add the code necessary to turn Galacticus fluxes in to CatSim SEDs and normalizations

- Update `phosim_descqa.py` to take all of its parameters from the protoDC2 catalog

- Add hack to overwrite anomalously large dust extinction values in protoDC2 catalog


yymao:
@danielsf thanks!

PR #21, opened 2017-11-15
Sed fitting refactor

danielsf:
refactors the CatSim interface into importable modules


yymao:
@danielsf Can we make a submodule and call it, say `CatSimInterface` (or other name you like better) and move the three scripts into it?
```
GCRCatalog/
    __init__.py
    ...
    CatSimInterface/
          __init__.py
          SedFitter.py
          DatabaseEmulator.py
          CatalogClasses.py
```

And in `GCRCatalog/CatSimInterface/__init__.py` we do:
```python
from .SedFitter import *
from .DatabaseEmulator import *
from .CatalogClasses import *
```

Then in `GCRCatalog/__init__.py` we just do:
```python
from .CatSimInterface import *
```

How do you think?

danielsf:
Yeah.  That sounds reasonable

danielsf:
I am going to call the module `GCRCatSimInterface` just in case other `CatSimInterface`s show up in peoples `$PYTHONPATH`

yymao:
@danielsf that's ok. i don't have strong options. I was thinking that people would either do 
```python
from GCRCatalogs import CatSimInterface
```
or 
```python
from GCRCatalogs.CatSimInterface import *
```
And hence it should be ok to not repeat "GCR". But either way is ok.

yymao:
@danielsf oh, I guess you are making the GCGCatSimInterface a top level package, right?

danielsf:
Yeah... I misread your proposed structure.

I don't have a strong opinion.

I need to run, now.  Let me know what you want the final structure to be and I will implement it later tonight/tomorrow morning.

yymao:
@danielsf No problem. This is fine but I have some other minor comments. Will post them here.

GCRCatSimInterface/CatalogClasses.py
```diff
@@ -0,0 +1,135 @@
+import os
+import numpy as np
+from GCRCatSimInterface import sed_from_galacticus_mags
```
yymao: Should be:
```python
from .SedFitter import sed_from_galacticus_mags
```

examples/phosim_descqa.py
```diff
@@ -36,205 +36,16 @@
...
-        Need to leave this method here to overload the get_phoSimMagNorm
-        in the base PhoSim InstanceCatalog classes
-        """
-        return self.column_by_name('fittedMagNorm')
-
-
 ############################
 # actually write the catalog
 
+from GCRCatSimInterface import PhoSimDESCQA
```
yymao: should move all import statements to the top

GCRCatSimInterface/CatalogClasses.py
```diff
@@ -0,0 +1,135 @@
...
+                                np.logical_or(rv_list<1.0, rv_list>5.0)))
+        rv_list[offensive_rv] = self._dust_rng.random_sample(len(offensive_rv[0]))*4.0+1.0
+
+        return np.array([av_list, rv_list])
+
+    @compound('sedFilename', 'fittedMagNorm')
+    def get_fittedSedAndNorm(self):
+
+        if not hasattr(self, '_disk_flux_names'):
+            catsim_mag_file = os.path.join(getPackageDir('gcr_catalogs'),
```
yymao: we should move `CatSimSupport` into `GCRCatSimInterface`, and then here we use `getPackageDir('GCRCatSimInterface')`

danielsf:
I think I have responded to all of your comments.  Let me know if anything still doesn't look right.

yymao:
@danielsf thanks! All looks good. Two additional comments: 

In `setup.py`, line 32 and 34 should be modified to:
```
packages=['GCRCatalogs', 'GCRCatSimInterface'],
package_data={'GCRCatalogs': ['catalog_configs/*.yaml'], 'GCRCatSimInterface': ['data/*']},
```

In `GCRCatSimInterface/CatalogClasses.py` Line 87-89: I thought it should be the following? 
```
            catsim_mag_file = os.path.join(getPackageDir('GCRCatSimInterface'),
                                           'data', 'CatSimMagGrid.txt')
```

I'll be offline soon. Will merge during the day (probably a better idea anyway) when these are fixed.

danielsf:
`getPackageDir` is a specific LSST utility to find the directory for a package that EUPS knows about.  I have configured `gcr-catalogs` so that it can be declared to EUPS (that is what the `ups/gcr_catalogs.table` file is for).  `GCRCatSimInterface` is one level above that, so calling the LSST `getPackageDir` method on it will not return anything.

That being said, I assumed that, since anyone running the CatSim interface will need to have the LSST stack setup anyway, it was alright to assume we would be declaring `gcr-catalogs` to EUPS and using it as though it were an LSST stack package.  Feel free to push back against that assumption.

I will fix `setup.py`

PR #28, opened 2017-11-29
Check redshift

danielsf:
This pull request changes the catalog classes to use `true_redshift` rather than `redshift` when calculating the distance modulus to apply to SED fits.

It also adds some scripts for generating plots to validate the SED fitting code.


yymao:
Should the validation scripts sit inside `GCRCatSimInterface/validation/` or `examples/validation/`? (This is a question. I am ok with either choice, but want to check if this has be considered)

danielsf:
I did not think about that.  I put them in `GCRCatSimInterface` because they only pertain to the CatSim interface.  Wherever they go, I think it should be clear that the only pertain to the CatSim interface.  We could put them in `/examples/CatSimInterface/` (I guess I could also move the `phosim_descqa.py` script in there, too).

yymao:
@danielsf that sounds like a good way to do it!

PR #37, opened 2017-12-04
added sky_area to buzzard reader

j-dr:
I've included sky_area in the buzzard configs and set it as an attribute in the reader.


yymao:
@j-dr looks good. Can you add `sky_area` to `buzzard_high-res_v1.1.yaml` too? We also need to figure out the sky area of the test catalog (has only 1 healpixel..)

j-dr:
Okay, just added. I changed the default pixel from 0 to 42 because 0 is on the edge so I don't know the exact area and 42 is definitely a whole healpixel.

yymao:
@j-dr How accurate are these numbers? Here's the numbers I got (in sq deq) --- 10335 (full buzzard), 413 (high-res), 53.9 (test). See https://portal.nersc.gov/project/lsst/descqa/v2/?run=2017-12-04_27&test=SkyArea (using nside = 256)

If I use nside=128, the number changes a bit: 10360 (buzzard), 427 (high-res), 55.8 (test) https://portal.nersc.gov/project/lsst/descqa/v2/?run=2017-12-04_26&test=SkyArea

Should we update the numbers in the yaml config files?

j-dr:
Ah, yes, Buzzard is a quarter sky so 10,313.25 degrees. I calculated the number for highres a while ago, and I forget how, but it doesn't look like the calculation here is converged. Can you go to nside=2048 for highres?

PR #38, opened 2017-12-04
add reader and config for protodc2 tidal addon catalog

yymao:
This PR adds a reader and a config file for the protoDC2 tidal addon catalog @EiffL created.

### DIFF
--- a/GCRCatalogs/alphaq_addon.py
+++ b/GCRCatalogs/alphaq_addon.py
@@ -39,8 +39,10 @@ class AlphaQTidalCatalog(BaseGenericCatalog):
         with h5py.File(self._filename, 'r') as fh:
             data = fh['tidal'].value
             for name, (dt, _) in data.dtype.fields.items():
-                for indices in product(*map(range, dt.shape)):
-                    native_quantities.add((name + '/' + '/'.join(map(str, indices))).strip('/'))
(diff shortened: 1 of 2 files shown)


EiffL:
Yes that will work, sorry I could have done myself but I was initially thinking of modifying the AlphaQ reader to read in an auxilliary file and expose the tidal field along with the other information

yymao:
@EiffL if the arrays have been sorted in the same order as the original catalog, I would modify the original reader too. But in this case, I figure it is better to use `pandas.merge`. Make sense?

yymao:
@EiffL I only exposed `eigvals` and `eigvects` --- are there other fields I should expose?

yymao:
This PR has been tested and can be merged. Here's an example of how to use it and pandas to add the info to the main catalog.

```python
import pandas as pd
import GCRCatalogs

gc = GCRCatalogs.load_catalog('protoDC2')
gc_tidal = GCRCatalogs.load_catalog('protoDC2_addon_tidal')

data = pd.DataFrame(gc.get_quantities(['galaxy_id', 'ra', 'dec'])).set_index('galaxy_id')
data_tidal = pd.DataFrame(gc_tidal.get_quantities(gc_tidal.list_all_quantities())).set_index('galaxy_id').join(data)
```

PR #55, opened 2017-12-13
Low memory gcr catsim interface

yymao:
Made several improvements to `GCRCatSimInterface.DatabaseEmulator` to use less memory and also speeding up a few things. 

(ping @TomGlanzman @jchiang87 @danielsf for their attention)


yymao:
This PR is an attempt to fix the issue raised in https://github.com/LSSTDESC/DC2_Repo/issues/33#issuecomment-351511088

yymao:
With @TomGlanzman's help, I was able to test this PR and the code works. It also uses less memory according to Tom (but Tom can give us more detail). In short, this temporary solution for protoDC2 seems to be OK. 

While Tom can already use this PR to continue his test, I'll still wait utill @danielsf has a chance to look at my changes before I merge this PR.

katrinheitmann:
Thanks very much for dealing with this so quickly!!

PR #57, opened 2017-12-24
Create SCHEMA.md

yymao:
Originally we maintain a schema documentation in [this Google Doc](https://docs.google.com/document/d/1rUsImkBkjjw82Xa_-3a8VMV6K9aYJ8mXioaRhz0JoqI/edit) but it becomes difficult to keep track of the schema (as people updated the schema and forgot to update the readers and vice versa) and inconsistent names have appeared (as expected).  This PR includes the schema into this repo so that we can track any future changes. 

[Click here to preview the schema](https://github.com/LSSTDESC/gcr-catalogs/blob/add-schema/GCRCatalogs/SCHEMA.md)

@dkorytov @evevkovacs @rmandelb @danielsf @j-dr: please take a look and see if there's any changes needed. I believe I cover everything needed by the image simulation (except for SEDs). 

I thank @EiffL for his help on this during holidays...

### DIFF
--- a/GCRCatalogs/SCHEMA.md
+++ b/GCRCatalogs/SCHEMA.md
@@ -1,0 +1,62 @@
+# Schema of GCR Catalogs as used in LSST DESC
+
+## Metadata
+
+Attribute name | Type | Definition
+--- | --- | ---
+`version` | `str` | catalog version
+`cosmology` | `astropy.cosmology` | cosmology of the mock catalog
+`halo_mass_def` | `str` | halo mass definition, e.g., `vir`, `200m`, `200c`
+`lightcone` | `bool` | whether or not the catalog is a light cone catalog
+
+## Galaxy properties
+
+- Label names are generally in `lowercase_separated_by_underscores` format, except in a few cases an upper case letter is needed (e.g., `Mag_true_Y_lsst_z0`).
+- Label names generally start with the name of the physical quantity, and are followed by specifications (i.e., use `size_disk_true` not `true_disk_size`). Some exceptions are `galaxy_id`, `halo_id`, `halo_mass`.
+- A quantity with `_true` usually means before the lensing effect is taken into account.
+- Not all quantities listed below are available in all catalogs. Use `list_all_quantities()` to find available quantities.
+- In addition to the quantities listed below, there are also *native quantities*, which are quantities whose label names and/or units have not been homogenized and may change in the future. Nevertheless, one can still access native quantities via GCRCatalogs. Use `list_all_native_quantities()` to find available native quantities.
+
+Quantity Label | Unit | Definition
+--- | --- | ---
+`galaxy_id` | - | Unique integer identifier
+`ra` | degree | Right ascension, lensed
+`ra_true` | degree | Right ascension, not lensed
+`dec` | degree | Declination, lensed
+`dec_true` | degree | Declination, not lensed
+`redshift_true` | - | Cosmological redshift
+`redshift` | - | Cosmological redshift with line-of-sight motion
+`Mag_true_<band>_<filter>_z<z>` | - | Absolute magnitude, not lensed, in `<band>` with  `<filter>` and k-corrected to `<z>`, e.g., `Mag_true_Y_lsst_z0`, `Mag_true_g_des_z01`
+`mag_<band>_<filter>_lsst` | - | Apparent magnitude, lensed, in `<band>` with  `<filter>`, e.g., `mag_Y_lsst`, `mag_g_des`
+`magerr_<band>_<filter>_lsst` | - | Error in apparent magnitude in `<band>` with  `<filter>`, e.g., `magerr_Y_lsst`, `magerr_g_des`
+`sersic` | - | Sersic index of galaxy light profile
+`sersic_disk` | - | Sersic index of disk light profile
+`sersic_bulge` | - | Sersic index of bulge light profile
+`A_v` | - | Extinction in V-band
+`R_v` | - | Ratio of total to selective extinction in B and V bands
+`size` | arcsec | Galaxy half-light radius (of major axis), lensed
+`size_true` | arcsec | Galaxy half-light radius (of major axis), not lensed
+`size_disk_true` | arcsec | Disk half-light radius (of major axis), not lensed
+`size_bulge_true` | arcsec | Bulge half-light radius (of major axis), not lensed
+`ellipticity_1` | - | Ellipticity component 1, lensed
+`ellipticity_2` | - | Ellipticity component 2, lensed
+`ellipticity_1_true` | - | Ellipticity component 1, for galaxy, not lensed
+`ellipticity_2_true` | - | Ellipticity component 2, for galaxy, not lensed
+`ellipticity_1_disk_true` | - | Ellipticity component 1, for disk, not lensed
+`ellipticity_2_disk_true` | - | Ellipticity component 2, for disk, not lensed
+`ellipticity_1_bulge_true` | - | Ellipticity component 1, for bulge, not lensed
+`ellipticity_2_bulge_true` | - | Ellipticity component 2, for bulge, not lensed
+`shear_1` | - | Shear (gamma) component 1 *(sign convention to be supplied)*
+`shear_2` | - | Shear (gamma) component 2 *(sign convention to be supplied)*
+`convergence` | - | Convergence (kappa)
+`magnification` | - | Magnification
+`halo_id` | - | Unique ID of the main halo that contains the galaxy
+`halo_mass` | M<sub>sun</sub> | Halo mass of the main halo that contains the galaxy
+`stellar_mass` | M<sub>sun</sub> | Total stellar mass of the galaxy
+`is_central` | *(bool)* | If the galaxy is the central galaxy of the main halo
+`position_x` | Mpc | 3D position (x coordinate)
+`position_y` | Mpc | 3D position (y coordinate)
+`position_z` | Mpc | 3D position (z coordinate)
+`velocity_x` | km/s | 3D velocity (x component)
+`velocity_y` | km/s | 3D velocity (y component)
+`velocity_z` | km/s | 3D velocity (z component)
--- a/GCRCatalogs/SCHEMA.md
+++ b/GCRCatalogs/SCHEMA.md
@@ -26,7 +26,7 @@ Quantity Label | Unit | Definition
 `dec_true` | degree | Declination, not lensed
 `redshift_true` | - | Cosmological redshift
 `redshift` | - | Cosmological redshift with line-of-sight motion
-`Mag_true_<band>_<filter>_z<z>` | - | Absolute magnitude, not lensed, in `<band>` with  `<filter>` and k-corrected to `<z>`, e.g., `Mag_true_Y_lsst_z0`, `Mag_true_g_des_z01`
+`Mag_true_<band>_<filter>_z<z>` | - | Absolute magnitude, not lensed, in `<band>` with  `<filter>` and k-corrected to `<z>`, e.g., `Mag_true_Y_lsst_z0`, `Mag_true_g_des_z01`. In the case when postfix is `_z0`, it means rest-frame absolute magnitude.
 `mag_<band>_<filter>_lsst` | - | Apparent magnitude, lensed, in `<band>` with  `<filter>`, e.g., `mag_Y_lsst`, `mag_g_des`
 `magerr_<band>_<filter>_lsst` | - | Error in apparent magnitude in `<band>` with  `<filter>`, e.g., `magerr_Y_lsst`, `magerr_g_des`
 `sersic` | - | Sersic index of galaxy light profile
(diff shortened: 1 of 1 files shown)


yymao:
Any comments, @dkorytov @evevkovacs @rmandelb @danielsf @j-dr? If not I would like to merge this PR and add a link to this new schema from the Google Doc.

danielsf:
Looks good.  What do you mean by "sign convention to be supplied" in the shear components?  Is that a decision we have to make, or are we letting each catalog define their own sign convention, as long as they specify it in the schema.  I am not sure I am comfortable with letting each catalog choose their own sign convention.

yymao:
I meant we (as a group) should decide the convention and require each catalog to do necessary conversion its reader. It says "to be supplied" because I don't think we have decided about the convention yet.

evevkovacs:
Here are my comments:
1) I think we also need  Mag_true_<band>_<filter>, for the rest=frame magnitudes that have not been kcorrected ( analagous to Mag_true_<band>_<filter>_z<z>, but without the z)
2) The shear sign convention should be documented somewhere. The easiest place is in the meta-data.
3) What happened to separate id's for the disk and bulge? Are these optional?
4) What about the galaxy SED? CatSim needs this.

yymao:
@evevkovacs 
1. Isn't the rest frame abs. mag. different from `Mag_true_lsst_z0`? 
2. See the discussion above. We should decide the convention as a group. 
3. These are directly generated from `galaxy_id` in the GCR-CatSim interface. 
4. Right now CatSim is using the native quantities in protoDC2. We can certainly also decide how to specify SEDs in this GCR schema If we want to. Do we have an agreement on how to specify SEDs?

evevkovacs:
Yes, rest frame abs. mag.  IS different from Mag_true_lsst_z0. I thought that we needed rest frame abs. mag. without any k corrections.  Why are the k-corrected values needed? Does CatSim require them?

evevkovacs:
We don't have an agreement on the SED specification.  Do any other catalogs have them?

yymao:
@j-dr mentioned that he might add SEDs to Buzzard.

j-dr:
This all looks good to me. I'd be happy to add buzzard SEDs, we just need to decide on units resolution I guess.

dkorytov:
It all looks good to me as well.

yymao:
@danielsf @evevkovacs @abensonca what are your opinions about SED specifications?

evevkovacs:
The PZ group also needs SEDs. They have a set of their own templates and provided me with the guidelines for what we did in Galacticus which was to sample the SED with a set of narrow-band top-hat filters. They still have to check if this is adequate for their needs. I think that the specification could be either a set of filter measurements or a template number in some standard template library.

yymao:
@evevkovacs @danielsf do you have any suggestion on naming scheme for the SEDs?

danielsf:
Are you asking about a naming scheme for the columns specifying the SEDs, or the names of the SEDs themselves?  If the latter, won't that entail us choosing a specific standard DESC library of SEDs?  I suppose we could put a field in the catalog metadata where the catalog generator specifies which library of SEDs they used and where it can be found.  Then the column of SED names could just be a column of strings that is customized to each specific library.

I feel like I may have misinterpreted the question and am roaming off-topic.  Sorry.

yymao:
@danielsf does the "standard DESC library of SEDs" you mentioned already exist (sorry for my ignorance on this). If so, can you point me to its specification? I agree that if there's already a standard SED library we should just use them in the GCR catalogs.

danielsf:
No, it does not.  I was trying to say that, before we can specify a naming scheme for individual SEDs, we have to pick a standard SED library, which is its own difficult discussion.

There is a standard list of SEDs used by the project simulations team.  You can get it with

```
 curl -O "https://lsst-web.ncsa.illinois.edu/sim-data/sed_library/seds_170124.tar.gz"
```

and then looking in the `galaxySED` sub directory of the resulting directory.  We, however, should not feel like we have to use this library.  PhoSim and ImSim can use any library of ASCII files for SEDs.

yymao:
@danielsf ah, ok. So this is actually a different way to specify the SEDs than what is currently being done in protoDC2 (which are luminosities in narrow-band filters). 

I don't have preference on which to use. Is there any technical reasons that we should choose one over the other?

danielsf:
Like I said: I may have misinterpreted the original question.  I am in favor of just doing what we did in protoDC2 and specifying a standard set of tophat bandpasses which we use to specify an SED.  That should maximize our flexibility going forward.

yymao:
OK. Then we still need a naming scheme on the tophat filtered luminosities. Maybe just `sed_<start>_<width>_<component>`

danielsf:
I am happy with that (let's just be sure to specify the units of `start` and `width`; some of us still think in nanometers ;)

evevkovacs:
Yes, this sounds good. Angstroms is what protoDC2 uses

yymao:
@danielsf @evevkovacs I've added SEDs to the [schema](https://github.com/LSSTDESC/gcr-catalogs/blob/add-schema/GCRCatalogs/SCHEMA.md), please take a look. 

How about the rest frame magnitude issue? Are we agreeing that rest frame means z=0? @j-dr @evevkovacs

evevkovacs:
Sorry, but I find this very confusing. The document states: Absolute magnitude, not lensed, in <band> with <filter> and k-corrected to <z>, e.g., Mag_true_Y_lsst_z0, Mag_true_g_des_z01. To me, an absolute magnitude means that you measure the SED over the wavelength range of the filter. No k-corrections are needed. Why are k corrections even mentioned in the definition?

danielsf:
Looks good to me.  Two comments about the units:

1) Does the unit 4.4659e13 W/Hz have some other name of which I am not thinking at the moment?

2) Do we need to specify whether this is the total luminosity (through all 4pi steradians), the observed luminosity, or the rest frame luminosity?  Alternatively, we can explicitly say that the normalization does not matter and that these should only be used for color-color fitting of SEDs.

abensonca:
The 4.4659e13 W/Hz units is based on the AB magnitude system. SDSS refer to them as "maggy"s for fluxes. So, I guess you could call it an "absolute maggy".

yymao:
@evevkovacs I believe k-corrected to z=0.1 means the absolute magnitude is calculated as if we were at z=0.1 (in which case the "rest frame" would be at z=0.1). 

@j-dr @abensonca is this correct interpretation? 

If so, then "rest frame" *does* equal to k-corrected to z=0, no?

abensonca:
Yes - this all gets very confusing.... It's typical to k-correct to some redshift close to the median redshift of a survey, as that minimizes the size of the corrections being made (which, observationally at least, are uncertain). So, in this case it's exactly as @yymao says: the absolute magnitude is that which you would measure in that filter if the galaxy were transported to z=0.1.

To me at least, "rest frame" always means the absolute magnitude which you would measure in a filter if the galaxy were transported to z=0.

j-dr:
Yes, what @abensonca said.

yymao:
Thanks @j-dr @abensonca. I added a sentence in the schema: "In the case when postfix is `_z0`, it means rest-frame absolute magnitude." 

I also added "rest-frame" for the SED luminosity definition as @danielsf suggested. 

I believe I've addressed all issues. Any other comments? 

@evevkovacs, let me know if you are happy with this version. If so I'll merge it.

evevkovacs:
@yymao Yes, I think that is very clear now.

evevkovacs:
Since Scott has clarified that the only shape parameters we need for image generation are
majorAxis::disk
minorAxis::disk
majorAxis::bulge
minorAxis::bulge

can I ask why the galaxy-component ellipticities appear in the schema at all? It would appear that they are redundant. Isn't it sufficient to just provide the total ellipticity,  ellipticity1, ellipticity2 , lensed and unlensed?  Can we drop the bulge and disk ellipticities?

yymao:
@evevkovacs As we have discussed, we should include all names that we want to unified in the schema, not just the ones used by CatSim. The schema is **not** a reference for what the required quantities are, but to inform the catalog provider/user that if a quantity exists, what its name should be.

aphearin:
@yymao - I think the point is that there is no use-case we could think of to separately include disk and bulge ellipticity. While it's true that "the quantity exists", since it serves no downstream purpose then it seems there is no value in including it.

yymao:
Yes, if no catalog is using these quantities, then your point stands. But these quantities are in the protoDC2 catalogs...

aphearin:
Not sure I follow. Is there a problem with removing a quantity that was in protoDC2 that need not have ever been included?

evevkovacs:
The only reason they are in protoDC2 is way back when, I got a suggested schema which had galaxy properties for all the components. This is our chance to rethink that list and ask if it really makes sense to include all those quantities.

yymao:
We can remove these quantities from the protoDC2 catalog --- we should check if anyone is using them (CatSim is not the *only* user of gcr-catalogs), but I agree that the answer is probably no one is using them. 

But before these quantities are removed from the catalog, they should be in the schema. I thought Eve was suggesting to remove these quantities from the schema, not from the catalog, since this thread is about the schema. 

When these quantities are removed from the catalog, and we agree that there is unlikely to have future use of these quantities, then we can also remove them from the schema. Since we do not require catalogs to have all the quantities in the schema, I don't think this last task is a high priority task anyway.

evevkovacs:
I was suggesting to remove them both from protoDC2 and from the schema.  Since we are trying to finalize the schema, I couldn't see the point of having quantities that no one will use. I agree it's not the highest priority, but carrying around useless quantities is not desirable either. What is a rational way to determine if anyone is likely to use these quantities?

yymao:
@evevkovacs For this particular case, I think as long as we decided it is ok to remove them from the catalog, we can remove them from the schema. These particular entries are there only because protoDC2 has them, and I agree that if they have no use, we should remove them. 

Since updating the catalog requires a new PR, I suggest that in that PR for updating the catalog, we also remove these entries in `SCHEMA.md`

evevkovacs:
@dkorytov If you are adding Rv  and Av, you can also remove the bulge and disk ellipticities from the reader.

dkorytov:
If we do remove the component, they still can be accessed as native quantities if a user would be interested. The data is still there.

yymao:
@dkorytov @evevkovacs but if we keep them in the catalog, why would we want to remove them from the schema?

evevkovacs:
Personally, I don't think they should be kept in the catalog. Salman has argued against having them and he has a point. The quantities make little sense observationally, because an observer would only be interested in the total ellipticity, and how well that could be recovered.

rmandelb:
@evevkovacs - it's also true that we can't observe the per-component major/minor axis lengths.  So I don't see Salman's argument about unobservable per-component ellipticities as being that relevant.  I mean, we're never going to see/measure any aspects of the individual components of the galaxy.  

To me the relevant issue is duplication of information.  You can specify an ellipse for the purpose of image generation with 3 numbers.   So for example, a set of numbers that includes (major axis length, minor axis length, position angle) is sufficient.  A set of numbers like (e1, e2, average half-light radius) would also be sufficient.  But if you have (major axis length, minor axis length, position angle, e1, e2) then that's a bunch of extra information.  It makes more sense to me that we'd keep a sufficient set of 3 numbers for the purpose of image generation, and if people really want (e1, e2) they can convert it.

yymao:
Again, the purpose of the SCHEMA document is to make sure that people specify quantity names in the same way, but *not* to require or even suggest what quantities should be included in the catalog. 

Since there are many different ways to specify ellipticities, I think it is actually important to include all different ways in the schema, so that people can find a reference on what name means what.

If we really think some quantities should not even appear in the catalogs, we should first remove them from the catalogs, not from the schema, as I mentioned above.

As for duplicate information, we might want to check the mathematical relationships among these quantities in the readiness test. This is something I am thinking to implement but haven't gotten time yet.

evevkovacs:
@yymao  @katrinheitmann  @rmandelb I'm going to push back a bit on Yao's comments:  First, the previous version of the schema (google doc) had a somewhat different purpose. Not only did it standardize the quantity names, it also documented which catalogs had which quantities and it flagged the quantities that are required by PhoSim.  I agree that the inventory of catalog quantities is much better left to DESCQA, but the inventory of quantities required by PhoSim is useful to have and I think we need to collect it somewhere. (We have pinged Scott several times with detailed questions that could have been answered by such a list).  That said, such a list could be elsewhere, although I think it is easiest  and most convenient to just star the needed quantities in the current schema.
 Second, (and historically) the schema started out as a minimal set of quantities, and that's closer to the way I have been thinking about it.  I see 2 ways to go here:
1) The schema lists standardized names for every possible quantity.  In this case, the current spec.  is probably incomplete. We could take the approach that if catalog providers want to add a quantity to their reader and there isn't a name or equivalent duplicate available, then they apply to this repo for a name addition.
2) We keep the schema to a "minimal set". In this case, if there are duplicate ways to compute a given quantity, then the catalog provider does this in their reader and converts the catalog native quantities to quantities and definitions as provided in the schema. 
If you think about it 2) is an extension of the policy we have taken with units. The native quantities are left up to the provider, but the units and definitions have been "standardized" in the schema/GCR.  Ideally, there shouldn't be duplicates. There may be "derived" quantities. For these we should check the relationship to make sure that the provided quantities are self consistent and match the expected definitions (eg ellipticity defines in terms of major and minor axes).  But I don't think we should support, for example, multiple ways of computing the ellipticity. That defeats the purpose of the GCR.

yymao:
@evevkovacs While I don't fully agree with your point about the historical purpose of the schema, the first way to go you proposed is closer to what I have in mind. Below I detail some more thoughts on this:

I fully agree with Eve that it would be useful to have a document listing what quantities CatSim uses for PhoSim/imSim image generation. However, I think it is more apporiate for this document to appear in the [GCR-CatSim interface repo](https://github.com/LSSTDESC/sims_GCRCatSimInterface). (cc @danielsf) 

As Eve already mentioned, listing available quantities in each of the catalogs should not be done manually. And we have this already set up in DESCQA. 

Our philosophy has always been maintaining flexibility whenever we can. In other words, the purpose of the schema is *not* to require or to suggest how to specify certain information. The idea of the schema is that if the catalog has a column whose name has been specified in the schema, then the meaning and the definition of that column needs to conform to the schema. That is why I think we should have those ellipticities one in the schema. The catalogs can of course choose to not provide them (because of duplications or whatever reason), but if they do, they must follow the schema's definition. (That is, they cannot have a column called `ellipticity_true_disk` but has a different meaning, but they can have no such column. 

Does that make sense?

PR #61, opened 2017-12-29
fix a mismatch of size_minor in quantity translation

yymao:
@jchiang87 you can use this PR to try again.


jchiang87:
The galaxy major/minor axis entries in the resulting instance catalogs look ok now.  I'm running a quick test image with imsim as an additional sanity check.

jchiang87:
fwiw, here is the imsim output for the center of the central sensor in the LSST focal plane using an instance catalog created with this fix:
gcr_catalogs_fix_450581
cf. https://github.com/LSSTDESC/DC2_Repo/issues/69#issuecomment-353762198

yymao:
Thanks @jchiang87. Does it mean now #56 is considered fixed?

jchiang87:
I think so.

PR #64, opened 2018-01-15
protoDC2 addon catalog with random walk parameters for knots of galaxy formation

EiffL:
This pull request adds a new generic way of specifying an addon catalog for protoDC2 that extends the values  of the  parent catalog (with new columns), and adds random walk paramters (n_knots, knots_flux_ratio) that define how many knots of star formation to use  for each galaxy, and what fraction  of the disk luminosity should be transfered to the  knots.


yymao:
@EiffL thanks for the PR! Do you have any examples (notebooks, scripts, etc) of using this addon catalog?

EiffL:
Sure, it's very easy:
```
import GCRCatalogs
addon = GCRCatalogs.load_catalog('protoDC2_addon_knots')
quantities  = addon.get_quantities(['n_knots', 'knots_flux_ratio', 'redshift'])
```

If you look at the configuration file/code what happens is that the parent catalog loads first, and the addon adds the columns from the specified group in the specified file

yymao:
@EiffL just to clarify --- this "addon" catalog is not *only* addon in the sense that it loads the original catalog for the user, right? 

This is fine but we might need a different name to avoid confusion. My original idea of an addon catalog only contains the addon part. This is more like a catalog upgrade?

EiffL:
Sure, no problem if you want to rename it, but I guess, this kind of
extension was what I was thinking an add-on corresponded to.

yymao:
@EiffL the other option is to modify the tidal addon so it behaves the same. (As long as it is consistent I don't mind what addon actually means...)

EiffL:
Sure, I think it's nice to have all the information in the same catalog for
the user.

EiffL:
@yymao Created the issue to keep track of the update needed for the tidal catalog. Are  you ok for merging this  pull request ?

PR #69, opened 2018-01-19
add standardized SED labels to protoDC2 reader

### LINKED ISSUE #67: Make SED labels conform to the schema in protoDC2 reader
Now that we have agreed on how to specified SED labels in the [schema](https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/SCHEMA.md), we should make SED labels in the protoDC2 reader conform to the schema.

yymao:
This PR adds  standardized SED labels in the protoDC2 reader. This PR fixes #67.


yymao:
@dkorytov maybe we can use this PR to also add `A_v` and `R_v` as @evevkovacs suggested?  If you want to open a new branch that's fine too, but in that case it's probably better to merge this one first.

dkorytov:
The schema defines the observed magnitudes as lensed. The observed magnitudes in protoDC2 aren't lensed. I don't think it will be that hard to add in magnification...let me talk to Patricia.

yymao:
Indeed..., we could have `mag_true_<band>_<filter>` in the schema.

yymao:
So I think we have a few things to address that we may want to include here, or in a new PR after this one is merged:

- this PR
- Issue #70
- Add lensed observed mag, and/or add `mag_true_<band>_<filter>` in the schema
- Remove unused columns (component-wise ellipticity) in the catalog and schema

PR #73, opened 2018-01-25
Add stellar masses for disk/bulge (fixes #72)

### LINKED ISSUE #72: add `stellar_mass_disk` and `stellar_mass_bulge`
@dkorytov mentioned that we should use stellar mass to check if the disk/bulge component exists. Hence, I think we should add `stellar_mass_disk` and `stellar_mass_bulge` to the protoDC2 reader as well as the schema, so that the user does not need to know the native labels.

yymao:
This PR adds stellar masses for disk/bulge to both the protoDC2 reader and the schema, and fixes #72. 

It also fixes a few PEP8 issue so that it can pass travis CI build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ before_script:
 
 script:
 - nosetests tests/test_import.py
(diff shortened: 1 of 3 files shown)


yymao:
Thanks @dkorytov

Issue #70, opened 2018-01-19
Add Av and Rv to the reader for protoDC2


evevkovacs:
Av and Rv are required by the schema and should be supplied by the reader. These quantities are needed for image sims and should be subjected to readiness testing.

yymao:
Thanks @evevkovacs. 

@danielsf also pointed out a [code reference](https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/master/python/desc/sims/GCRCatSimInterface/CatalogClasses.py#L37) which I think would be useful when we implement this in the protoDC2 reader.

dkorytov:
So I've implemented the Av & Rv quantities. There are infinities in both of them and Rv has NANs when the denominator is zero with an non-zero numerator. 

Here's a distribution of Av (2nd figure is just a zoom in ) 
![image](https://user-images.githubusercontent.com/25048061/35240018-c1b0b702-ff77-11e7-8017-8ae04d45d1e1.png)

Rv
![image](https://user-images.githubusercontent.com/25048061/35240123-0cb943f4-ff78-11e7-8076-3802e959b75e.png)

Do these look reasonable? I don't have any feel for Av/Rv.

dkorytov:
@danielsf, I forgot to mention you in the above post.

evevkovacs:
The distribution for Av looks reasonable I think. The infinities are probably due to a failure in the dust model extrapolations. Andrew B worked hard on these, so maybe not so much room for further improvement.  As long as it is just a handful, it should be acceptable. Rv for the Milky Way is 3.1, so I think Rv looks ok too, but @danielsf should comment.

danielsf:
@dkorytov For my own sanity: what are

1) the total number of galaxies with Rv<0
2) the total number of galaxies with |Rv| > 5
3) the total number of galaxies in the sample (so I can see (1) and (2) as fractions)

Thanks

dkorytov:
@danielsf here it is:
Rv<0 / total = 3097/18265 = 0.1696
|Rv| > 5 / total = 3285/18265 = 0.1799

The sample isn't too large because I'm using a small test catalog on my laptop.

danielsf:
[...]

PR #74, opened 2018-01-25
Adding Av & Rv

dkorytov:
I'm a bit confused why this PR has 41 commits... Okay, I see. My changes where Incorporated though a PR of Yao's fork of my repo. Is this correct way to do this?


GCRCatalogs/alphaq.py
```diff
@@ -110,21 +110,30 @@ def _collect_native_quantities(name, obj):
...
-            'ellipticity_2_bulge_true': 'morphology/spheroidEllipticity2',
             'size_true': (
                 lambda size1, size2, lum1, lum2: ((size1*lum1)+(size2*lum2))/(lum1+lum2),
                 'morphology/diskMajorAxisArcsec',
-                'morphology/spheroidMinorAxisArcsec',
+                'morphology/spheroidMajorAxisArcsec',
                 'LSST_filters/diskLuminositiesStellar:LSST_r:rest',
                 'LSST_filters/spheroidLuminositiesStellar:LSST_r:rest',
             ),
+            'Av' : (lambda lum_v, lum_v_dust: (-2.5*(np.log10(lum_v_dust/lum_v))),
```
yymao: We use `A_v` and `R_v` in [our schema](https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/SCHEMA.md)

GCRCatalogs/alphaq.py
```diff
@@ -110,21 +110,30 @@ def _collect_native_quantities(name, obj):
...
+                    'otherLuminosities/totalLuminositiesStellar:V:rest',
+                    'otherLuminosities/totalLuminositiesStellar:V:rest:dustAtlas',
+              
+            ),
+            'Rv' : (lambda lum_v, lum_v_dust, lum_b, lum_b_dust :
```
yymao: Use `R_v`

GCRCatalogs/alphaq.py
```diff
@@ -110,21 +110,30 @@ def _collect_native_quantities(name, obj):
             'sersic_disk':              'morphology/diskSersicIndex',
             'sersic_bulge':             'morphology/spheroidSersicIndex',
             'ellipticity_true':         'morphology/totalEllipticity',
-            'ellipticity_disk_true':    'morphology/diskEllipticity',
```
yymao: We should probably leave these alone for now? I am not sure if we have reached agreement on this... and we should also check if anyone is using them

GCRCatalogs/alphaq.py
```diff
@@ -110,21 +110,30 @@ def _collect_native_quantities(name, obj):
...
                 'LSST_filters/diskLuminositiesStellar:LSST_r:rest',
                 'LSST_filters/spheroidLuminositiesStellar:LSST_r:rest',
             ),
+            'Av' : (lambda lum_v, lum_v_dust: (-2.5*(np.log10(lum_v_dust/lum_v))),
+                    'otherLuminosities/totalLuminositiesStellar:V:rest',
+                    'otherLuminosities/totalLuminositiesStellar:V:rest:dustAtlas',
+              
+            ),
+            'Rv' : (lambda lum_v, lum_v_dust, lum_b, lum_b_dust :
+                    np.choose(((lum_v_dust/lum_v==1.0) & (lum_b_dust/lum_b/lum_v_dust*lum_v==1.0)).astype(int),
```
yymao: I don't think this code does what it should be? I could be wrong but in any case it's not easy to read.. I suggest create a function for it, like

```python
def _calc_Rv(lum_v, lum_v_dust, lum_b, lum_b_dust):
    v = lum_v_dust/lum_v
    b = lum_b_dust/lum_b
    Rv = np.log10(v) / np.log10(b/v)
    Rv[np.isnan(Rv)] = 1.0
    return Rv
```

yymao:
Note: this PR fixes #70

GCRCatalogs/alphaq.py
dkorytov: I'll put back disk/bulge ellip, ellip1/2.

GCRCatalogs/alphaq.py
dkorytov: I was trying to keep it as a pure lambda function. I wasn't sure if messy lambda function was better than a free floating function. If you think a function is better, lets do that. 

I don't want to relying on NANs to handle the case of v=1 & b=1. A NAN can sneak it's way in some other way and we would just overwrite it instead of realizing something is off. The reason why I had (lum_b_dust/lum_b/lum_v_dust*lum_v ==1) is because there were rounding errors in the evaluation where v = 1, b != 1 but b/v = 1.

yymao: @dkorytov thanks for the comment. Let's separate out a few things here ---

- lambda function vs. regular function: I don't really mind which one we use, and lambda function is indeed convenient for this kind of use. In this particular case, since there are two intermediate quantities (`v` and `b`) that are used more than once, I thought it would be more efficient to use a regular function in which we can define intermediate variables, but not in lambda function. 

- checking nan vs.  checking `(v == 1) & (b == 1)`: I agree with you point and we can use the latter. 

- the other main issue here is actually the use of `np.choose`, I think you might have meant `np.where`? In any case, using a boolean mask to slice and then set values is probably the most efficient and readable way. 

So, one way to implement would be: 
```python
def _calc_Rv(lum_v, lum_v_dust, lum_b, lum_b_dust):
    v = lum_v_dust/lum_v
    b = lum_b_dust/lum_b
    b /= v
    Rv = np.log10(v) / np.log10(b)
    Rv[(v == 1) & (b == 1)] = 1.0
    return Rv
```

dkorytov: I agree with all points. I forgot about `np.where`.  `np.choose` is a messier solution.

dkorytov: There are still NANs when b = 0 and v = 0, which happens when all the v/b band are completely quenched by dust. Should we add another condition to avoid this known NAN? Or just let the NANs be until we figure out what to do with v = b = 0 issue.

yymao: In the case that `lum_v_dust/lum_v = lum_b_dust/lum_b = 0`, shouldn't we set `A_v = np.inf` and `R_v = 1`?  

But @danielsf should comment on this.

dkorytov: It depends at the rate `lum_v_dust/lum_v` and `lum_b_dust/lum_b` go to zero. If they have different rates, we can get Rv = inf/real number = +/- inf. If it's the same rate, then it should equal 1.

danielsf: I think I would prefer setting `Av=Rv=NaN` in these cases.  That way, on the CatSim side, we can just detect these objects with `np.where(np.isnan())`, as opposed to `np.where(np.logical_or(np.isnan(), np.isinf())))`.

Do I infer correctly that the long term goal is to prevent `lum_v_dust` and `lum_b_dust` from ever actually going to zero (or, at least, minimizing the frequency with which this happens)?  Or do we think this corresponds to some actual physical effect?

evevkovacs:
@danielsf  Yes, you infer correctly that they are outlier values coming from a couple of different sources: either a failure in the extrapolations of the dust model or very small galaxy-component sizes. Our plan is to vet these values for the next version of protoDC2

dkorytov:
So is everybody fine with setting Rv to `NAN` if we get complete extinction? 

For having no dust extinction, should we leave Av as `np.inf` and Rv as `1.0`?

dkorytov:
Never mind about Av = inf when there is no dust extinction. I'm going to set all the bad values to NANs if everybody is fine with that.

GCRCatalogs/alphaq.py
```diff
@@ -110,21 +123,34 @@ def _collect_native_quantities(name, obj):
...
+            'ellipticity_2_bulge_true':       'morphology/spheroidEllipticity2',
             'size_true': (
                 lambda size1, size2, lum1, lum2: ((size1*lum1)+(size2*lum2))/(lum1+lum2),
                 'morphology/diskMajorAxisArcsec',
-                'morphology/spheroidMinorAxisArcsec',
+                'morphology/spheroidMajorAxisArcsec',
                 'LSST_filters/diskLuminositiesStellar:LSST_r:rest',
                 'LSST_filters/spheroidLuminositiesStellar:LSST_r:rest',
             ),
+            'A_v' : (lambda lum_v, lum_v_dust: _calc_Av(lum_v,lum_v_dust),
```
yymao: No need for `lambda`
```python
'A_v': (_calc_Av, 
        'otherLuminosities/totalLuminositiesStellar:V:rest',
        'otherLuminosities/totalLuminositiesStellar:V:rest:dustAtlas',
),
```

GCRCatalogs/alphaq.py
```diff
@@ -110,21 +123,34 @@ def _collect_native_quantities(name, obj):
...
+                    'otherLuminosities/totalLuminositiesStellar:V:rest',
+                    'otherLuminosities/totalLuminositiesStellar:V:rest:dustAtlas',
+              
+            ),
+            'R_v' : (lambda lum_v, lum_v_dust, lum_b, lum_b_dust :
```
yymao: Similarly, no need for `lambda`
```
'R_v' : (_calc_Rv, [...]
```

GCRCatalogs/alphaq.py
```diff
@@ -110,21 +123,33 @@ def _collect_native_quantities(name, obj):
...
                 'morphology/diskMajorAxisArcsec',
-                'morphology/spheroidMinorAxisArcsec',
+                'morphology/spheroidMajorAxisArcsec',
                 'LSST_filters/diskLuminositiesStellar:LSST_r:rest',
                 'LSST_filters/spheroidLuminositiesStellar:LSST_r:rest',
             ),
+            'A_v' : (_calc_Av,
+                    'otherLuminosities/totalLuminositiesStellar:V:rest',
+                    'otherLuminosities/totalLuminositiesStellar:V:rest:dustAtlas',
+
```
yymao: remove unnecessary empty line

GCRCatalogs/alphaq.py
```diff
@@ -110,21 +123,33 @@ def _collect_native_quantities(name, obj):
...
+                    'otherLuminosities/totalLuminositiesStellar:V:rest',
+                    'otherLuminosities/totalLuminositiesStellar:V:rest:dustAtlas',
+              
+            ),
+            'R_v' : (_calc_Rv,
+                    'otherLuminosities/totalLuminositiesStellar:V:rest',
+                    'otherLuminosities/totalLuminositiesStellar:V:rest:dustAtlas',
+                    'otherLuminosities/totalLuminositiesStellar:B:rest',
+                    'otherLuminosities/totalLuminositiesStellar:B:rest:dustAtlas',
+
```
yymao: remove unnecessary empty line

yymao:
@dkorytov I fixed the empty lines (hope you don't mind). 

The reader is ready to be merged, but I want to double check again that these `A_v` and `R_v` values are reasonable. [See the test here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-01-28_3&test=readiness_protoDC2&right=2018-01-28_3/readiness_protoDC2/protoDC2/SUMMARY.html), in particular, notice that `R_v` ranges from -2.301e+15 to 1.499e+08 with an standard deviation of  7.286e+11. Are these numbers expected?

dkorytov:
That's quite alright about fixing the lines. :)

I think @danielsf would have the most informed opinion on Rv & Av. These are the same distribution as before but there are now no infinities:
screenshot from 2018-01-28 18-14-40
screenshot from 2018-01-28 18-14-44

yymao:
Hmm, @dkorytov the ranges we got seem different. For `A_v` I got -0.0108 to 808.3, and for `R_v` I got -2.301e+15 to 1.499e+08. Any idea what might be causing this difference?

danielsf:
As I said in the issue conversation

https://github.com/LSSTDESC/gcr-catalogs/issues/70

I'm not really happy with that Rv distribution, but I think the conclusion was "this is a simulation; there are always going to be unphysical galaxies; we should just try to catch them before we simulate the images/do any catalog analysis."

Feel free to tell me that I misinterpreted that previous discussion.

evevkovacs:
@danielsf @dkorytov  In the next version of DC2, we will be resampling from Galacticus galaxies and choosing only those deemed to be "good". Right now, we are removing galaxies with extreme values of extinction (=0 to machine precision). I propose that we try this scheme and see how our Rv distribution looks in the new catalog. If the distribution is still unsatisfactory, we can make a more stringent cut.

dkorytov:
I'm using a small catalog on my laptop to generate these plots, so I'm not surprised that the ranges are bit different. My ranges should always be within the full catalog range.

yymao:
OK. But do the numbers  -2.301e+15 and 1.499e+08 make sense? Don't they seem too large to physical?

yymao:
@danielsf sorry for keeping pinging during this busy time. I just want to make sure the `A_v` and `R_v` values we feed to CatSim are acceptable, for the purpose of Run 1.1. (I know that people have said that things will be fixed in the future catalogs, but I want to know what should we do with the current catalog.) 
I see two options here:
1. Keep `A_v` and `R_v` as they are now (which mean there are some unphysical values). 
2. Find those unphysical values and set them to NaN. 

@danielsf do you have a preference? 

Also, I should note that the current GCR-CatSim interface is not using `A_v` and `R_v` in the GCR. Hence, if we go with option 2, we also need to address LSSTDESC/sims_GCRCatSimInterface#9.

danielsf:
If we do nothing, this code

https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/master/python/desc/sims/GCRCatSimInterface/CatalogClasses.py#L87

will replace unphysical values of Av and Rv with random values.  That is how `sims_GCRCatSimInterface` works now, without using the `Av` and `Rv` values you are implementing in the reader.

I think that is the safest thing to do at this late date.

yymao:
Thanks, @danielsf. In that case, I'll consider this done and just note that we will *not* use `A_v` and `R_v` directly from the GCR catalog in Run 1.1.

PR #75, opened 2018-01-27
Convergence Fix

dkorytov:
The value in the catalog is not right. But we can recompute the value using the shears and magnification.


dkorytov:
On the edge of the catalog, magnification is zero. In that case I'm setting the magnification to 1 so that we get a reasonable convergence. 

The plot below shows the ra & dec positions of galaxies with zero magnitude. 
screenshot from 2018-01-26 18-21-49

GCRCatalogs/alphaq.py
```diff
@@ -94,7 +101,10 @@ def _collect_native_quantities(name, obj):
             'redshift_true': 'redshiftHubble',
             'shear_1':       'shear1',
             'shear_2':       'shear2',
-            'convergence':   'convergence',
+            'convergence':   (lambda mag, shear1, shear2 : _calc_conv(mag, shear1, shear2),
```
yymao: No need for `lambda`
```python
'convergence':   (_calc_conv, 'magnification', 'shear1', 'shear2'),
```

dkorytov: Done.

GCRCatalogs/alphaq.py
```diff
@@ -105,7 +105,7 @@ def _collect_native_quantities(name, obj):
                               'magnification',
                               'shear1',
                               'shear2'),
-            'magnification': (lambda mag: np.where(mag<0.1,np.ones_like(mag),mag),
+            'magnification': (lambda mag: np.where(mag<0.1, np.ones_like(mag), mag),
```
yymao: `np.where` can broadcast, so no need to first allocate memory for 1's. 
```python
lambda mag: np.where(mag < 0.1, 1.0, mag),
```
BTW, this is changing all magnification less than 0.1 to 1. Is this intended?

GCRCatalogs/alphaq.py
```diff
@@ -26,6 +26,13 @@ def md5(fname, chunk_size=65536):
             hash_md5.update(chunk)
     return hash_md5.hexdigest()
 
+def _calc_conv(mag, shear1, shear2):
+    slct = mag < 0.2
```
yymao: here you select mag < 0.2, but you use mag < 0.1 when modifying magnification itself. Just to make sure this is intended.

GCRCatalogs/alphaq.py
evevkovacs: Yes, it is intended. These small values (<0.1) are unphysical and have been traced to edge effects in the lensing code (all galaxies with these small values fall clode to the boundary of the cut-out for the catalog. It is reasonable to set these v

yymao: Ok, I'm just making sure since this change also implies that magnification cannot be less than 0.1.

GCRCatalogs/alphaq.py
dkorytov: Good catch!

PR #77, opened 2018-01-29
add minor-axis size and bulge-to-total ratio to protoDC2

### LINKED ISSUE #76: Add size_minor to schema and gcr reader
@yymao  @dkorytov For the ellipticity test, I need to recompute the ellipticity using the axis ratio for the galaxy. (because the data defintion  differs from the schema definition). But I just checked and only have size_minor_bulge_true, size_minor_disk_true, size_true, size_bulge_true, size_disk_true. I need  size_minor_true, which is not available. For protoDC2, this quantity needs to be added (I think Dan was using the r-band luminosity weighted sum of the components ) I can do this by hand for the time being, but the axis ratio needs to be in the schema as well I think.

yymao:
This PR fixes #76. 

This PR adds `size_minor_true` to protoDC2 using `size_true` and `ellipticity_true`. 

This PR also adds `bulge_to_total_ratio_i` to protoDC2 and an entry `bulge_to_total_ratio_<band>` to the schema. (I'm not sure if this is the best way; ideas welcome). 

(cc'ing @EiffL)

### DIFF
--- a/GCRCatalogs/SCHEMA.md
+++ b/GCRCatalogs/SCHEMA.md
@@ -65,7 +65,7 @@ Quantity Label | Unit | Definition
 `stellar_mass` | M<sub>sun</sub> | Total stellar mass of the galaxy
 `stellar_mass_disk` | M<sub>sun</sub> | Stellar mass of the disk component
 `stellar_mass_bulge` | M<sub>sun</sub> | Stellar mass of the bulge component
-`bulge_to_total_ratio_i` | - | bulge-to-total luminosity ratio in *i* band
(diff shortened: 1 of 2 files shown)


evevkovacs:
Schema change is fine.  B/T in i-band will be good for a start. If other validation tests are developed with different needs, we can adjust. 
I think there's a bug in the calculation of the weighted size. I commented.

yymao:
@evevkovacs Thanks! Yes, the bug has been fixed by Dan in #74 but it has not been merged yet. I've fixed it again here.

PR #83, opened 2018-02-21
add dust parameters for disk/bulge in protoDC2/schema

yymao:
I found that the values of dust parameters (`A_v` and `R_v`) for disk and for bulge are acutally different from each other, and also different from the dust parameters for the total light. 

I am not sure if this behavior is intended in protoDC2 (@evevkovacs @abensonca?), but it means that we need to define different quantities for disk/bulge dust parameters. 

This PR adds `A_v_disk`, `A_v_bulge`, `R_v_disk`, `R_v_bulge` to the protoDC2 reader and the GCR schema. 

Related issues: https://github.com/LSSTDESC/sims_GCRCatSimInterface/issues/9 & https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/32 (cc @danielsf)


abensonca:
Yes, that's expected. The disk and spheroid light are distributed differently relative to the dust so they are extinguished differently. The total light Av and Rv will be some weighed mean of the disk and spheroid.

evevkovacs:
Yes, as Andrew comments above, A_v and R_v are determined from the luminosities in the V and B filters before and after dust corrections. There's no reason to expect them to be the same for the bulge and disk components.

yymao:
Thanks @abensonca and @evevkovacs. Unless there's objection, I'll merge this on Friday (cc @dkorytov).

PR #92, opened 2018-02-24
add lsst mag to buzzard

yymao:
This PR adds the lsst mags to buzzard catalog. The previous PR #87 did not add the quantities to the reader.


yymao:
BTW, @j-dr, it seems that magerr for lsst are all zeros, is that a bug? (cc @cosmicshear)

j-dr:
No, it's not. We just haven't made lsst magnitude errors for buzzard. It's
probably something that will be useful though, so happy to do it. I'll
probably make them for the full 10 year survey.

yymao:
OK. I've removed `mag` and `magerr` since their native values are not populated. @j-dr

PR #94, opened 2018-03-20
ProtoDC2 v3

dkorytov:
The new GCR for protoDC2 v3. The only real change is halo_id is no specified by hostHaloTag because we don't have hostIndex in the new version.


GCRCatalogs/alphaq.py
```diff
@@ -134,7 +135,7 @@ def _collect_native_quantities(name, obj):
                 'shear2',
             ),
             'magnification': (lambda mag: np.where(mag < 0.2, 1.0, mag), 'magnification'),
-            'halo_id':       'hostIndex',
+            'halo_id':       'hostHaloTag',
```
yymao: Is this change backward compatible? If not, need to change the modifier for older catalog files.

dkorytov: It is backwards compatible. But the older versions should remain the same. I've updated the GCR to use the older hostIndex for the older versions.

dkorytov:
Great! Thanks!

evevkovacs:
I have a modified readiness test yaml file to merge,
I'm just trying it now. I checked out a branch
and made a pull request.

Note also that protoDC2.yaml also has to be switched to v3.

I want to run one more test with DESCQA defaults to make sure it all works
once all the changes are in place.

First thing tomorrow morining is fine.
I just launched a test based on my private version of GCRCatalogs


https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-03-19_75

Eve

PR #98, opened 2018-03-28
GCR fix for protoDC2 v3 not having uniform position angle

dkorytov:
The position angle should be random and completely uniform. The angle is assigned in post processing of Galacticus and is uniform. In the pipeline, when we sample from the processed Galacticus, we no longer recover the uniform distribution. This pull requests generates new positions angles from a uniform distribution and recalculates ellipticity 1 and 2 for the total, disk and bulge components. 

Git issue: I did a git pull --rebase on the upstream LSSTDESC/gcr-catalog. It seemed to work fine for alphaq.py but this pull request is showing changes to GCRCatalogs/redmapper.py, README.md, etc that I thought rebase should take care of correctly. Did I mess something up?


GCRCatalogs/alphaq.py
```diff
@@ -60,6 +60,35 @@ def _calc_Av(lum_v, lum_v_dust):
     return Av
 
 
+def _gen_position_angle(size_reference):
```
yymao: should cache `pos_angle` so that it doesn't regenerate every time the function is called. 

```python
def _gen_position_angle(size_reference):
    size = size_reference.size
    if not hasattr(_gen_position_angle, "_pos_angle") or _gen_position_angle._pos_angle.size != size:
        _gen_position_angle._pos_angle = np.random.RandomState(123497).uniform(0, 180, size)
    return _gen_position_angle._pos_angle
```

dkorytov: Sorry about the messed up commit history.. I'm still getting a hang of git--I'll follow the instructions next time. :)

GCRCatalogs/alphaq.py
```diff
@@ -248,6 +273,12 @@ def _collect_native_quantities(name, obj):
         if catalog_version < StrictVersion('2.1.2'):
             self._quantity_modifiers.update({
                 'position_angle_true':     (lambda pos_angle: np.rad2deg(np.rad2deg(pos_angle)), 'morphology/positionAngle'), #I converted the units the wrong way, so a double conversion is required.
+                'ellipticity_1_true':       'morphology/totalEllipticity1'),
```
yymao: These lines should be moved to under `if catalog_version < StrictVersion('3.0'):` around Line 270.

dkorytov: That's right...

GCRCatalogs/alphaq.py
```diff
@@ -243,6 +268,12 @@ def _collect_native_quantities(name, obj):
         if catalog_version < StrictVersion('3.0'):
             self._quantity_modifiers.update({
                 'host_id': 'hostIndex',
+                'ellipticity_1_true':       'morphology/totalEllipticity1'),
```
yymao: Invalid syntax here. Remove `)` in this and the following lines.

Additionally, need to add `'position_angle_true':      'morphology/positionAngle',` here too.

PR #99, opened 2018-03-28
Galaxy Id Fix

dkorytov:
The galaxy IDs were too large to use downstream when they were being multiplied to generate a new for disk/bulge component ids. The old IDs are ignored and a new set of IDs are generated by np.arange(size). 

Git question: This pull request is showing 23 commits starting from May 19. Shouldn't there be only a couple commits after merging from the LSST upstream. Did I mess something up with my commit history with the --rebase or something else? Is there a way to sync it up correctly? I followed https://github.com/LSSTDESC/gcr-catalogs/blob/master/CONTRIBUTING.md to the letter.


yymao:
Looks good. Have you run this in DESCQA?

dkorytov:
No, I haven't. I checked the GCR runs and returns unique values for each
galaxy.

I'm pushing it through DESCQA now.

dkorytov:
The DESCQA run is in link below if you're curious
https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-03-28_3&test=readiness_protoDC2&right=2018-03/2018-03-28_3/readiness_protoDC2/protoDC2_test/SUMMARY.html

PR #101, opened 2018-04-10
Add two required field in redmapper.py

chto:
These two fields are necessary for redmapper CLF calculation.


GCRCatalogs/redmapper.py
```diff
@@ -67,6 +67,8 @@ def _subclass_init(self, catalog_root_dir,
             'richness'          :   'clusters/LAMBDA_CHISQ',
             'halo_id'           :   'clusters/MEM_MATCH_ID',
             'halo_mass'         :   'clusters/M200',
+            'LIM_LIMMAG_DERED'  :   'clusters/LIM_LIMMAG_DERED',
```
yymao: use lower case?

PR #102, opened 2018-04-10
Focal Plane reader

fjaviersanchez:
This is the first version of the focal plane reader to test the e-images (it should work for both phosim and imsim). I developed this to connect with some tests that I will push later (today or tomorrow probably). More to come.


yymao:
@fjaviersanchez do you have further thoughts on this reader before I merge it?

fjaviersanchez:
Sorry @yymao I just saw this. Yeah, I agree we need an example notebook. I am happy to prepare one (but I need some time to put it together). Other than that, I think I am happy with its status.

PR #103, opened 2018-04-11
bug fix in the alphaq catalog reader for backward compatibility

yymao:
This PR fixes a bug when using the new reader to load older catalogs.


dkorytov:
I don't see the to be deleted "ellipticity_1/2" defined anywhere in the AlphaQ reader. Do you mean "ellipticity_1/2_true"?

PR #105, opened 2018-04-18
magnitudes include host dust extinction

### LINKED ISSUE #104: map magnitudes to dusted magnitudes in protoDC2
It is recently discovered that the "dusted" magnitudes in the protoDC2 catalogs are accounting for the dust extinction from the host galaxy, but not from the Milky Way. Hence, these dusted magnitudes are what most general users would want to use. However, currently the magnitudes that are exposed via GCR are the undusted magnitudes, which are mostly only relevant for CatSim. 

Hence, we should map the magnitudes in GCR to dusted magnitudes in protoDC2. 

Thanks to @aphearin @evevkovacs @danielsf @rongpu for raising the clarifying this issue.

dkorytov:
The default magnitudes in the GCR include the dust in the host galaxy. 

I think I fixed the commit history issue. 

fixes #104


rbiswas4:
Sorry, just checking: Is this for the galaxy itself or point sources (like SN) within the galaxy too?

dkorytov:
This just for ugriz bands (SDSS & LSST) of the galaxy. I think SN are added later in CatSim.

rbiswas4:
Thanks!

yymao:
Looks good to be. Once we reach a conclusion in #104, we can include the fix to SEDs in this PR too.

dkorytov:
It all looks good. :)

PR #111, opened 2018-06-01
 flip shear_2 sign and add shear_2_phosim

### LINKED ISSUE #108: add option to retrieve shear values in different sign conventions
The proposal is to add an option (something like `convention='treecorr'` or `convention='phosim'`) to the reader and the reader will return the shear (gamma) values using the specified convention. 

(cc @rmjarvis @cwwalter @danielsf @jchiang87 @katrinheitmann)

### LINKED ISSUE #112: add sign convention to shear parameters in schema
The schema does not list the sign convention for `shear_1` and `shear_2`.  Once we sort out what the convention in protoDC2 is here

https://github.com/LSSTDESC/gcr-catalogs/pull/111

we should document it in the schema.

yymao:
This PR fixes #108 and fixes #112


danielsf:
I remember that @jchiang87 did a study somewhere to help us figure out what the sign convention of the shear parameters in DC2 was vis-a-vis PhoSim, but I cannot find it.  Does anyone remember where that is?  I just want to make sure that the negative sign is getting applied in the right way.

jchiang87:
I think that must have been in the imSim repo as one of the issues.  I can look after the CI meeting.  There is also a more recent (closed) issue in imSim on implementing WL shears that may reference the one you are thinking of.

evevkovacs:
Sign conventions for protoDC2 are [here](https://docs.google.com/document/d/1rUsImkBkjjw82Xa_-3a8VMV6K9aYJ8mXioaRhz0JoqI/edit#)

danielsf:
Thanks, Jim. I checked sims_GCRCatSimInterface and DC2_Repo.  Forgot to check ImSim.  Here is the PR I was thinking of

https://github.com/LSSTDESC/imSim/pull/72

danielsf:
Unfortunately, that test was just dealing with the InstanceCatalog->image part of the ImSim pipeline.  It says nothing about the protoDC2 convention.

danielsf:
Looking at @linan7788626's paragraph at the end of the document Eve linked to, it sounds to me like protoDC2 is consistent with GalSim: off-diagonal elements of the Jacobian matrix are related to (+shear2).  I think that means the negative sign should be applied to `shear_2_phosim`.  Someone should check my reasoning.

yymao:
Well but according to [this thread](https://github.com/LSSTDESC/descqa/issues/8#issuecomment-350122716) DC2 and Buzzard has the phosim convention.... Is that right, @patricialarsen @chihway?

danielsf:
I just want to point out that, earlier in that thread, Eve says that the shear sign convention of DC2 matches GalSim and is off by a negative sign in shear2 from PhoSim.  It is not explicitly clear to me as a lensing non-expert that the question was ever actually resolved.

evevkovacs:
@danielsf @patricialarsen I believe we concluded in the past that protoDC2 conventions matched the phosim convention. (If I said otherwise, it was a mistake). If you look at the google doc I referenced, there is a comment that you have to note that the direction of ra in galsim and phosim are opposite. So, Galsim and Treecorr have the same sign convention, which is opposite to that of phosim/protoDC2.

yymao:
But given [this DESCQA test](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-04_1&test=shear), I think buzzard and protoDC2 actually have the same convention.

And [in the source code](https://github.com/LSSTDESC/descqa/blob/master/descqa/shear_test.py#L273) you can see that `shear_2` in the catalogs is assumed to have an additional negative sign compared to treecorr convention.

evevkovacs:
@yymao @danielsf Yao's comment is consistent with what I just wrote above. The flipped direction of ra is causing the confusion.

danielsf:
Okay.  I am convinced.  Thanks for indulging my paranoia.

yymao:
Can @dkorytov and @j-dr take a look at these changes so that we can merge this? Thanks!

j-dr:
Looks good for Buzzard as well.

yymao:
Thanks, @dkorytov and @j-dr! I'll merge this by the end of Tuesday unless I hear objections. 

We'll then need to update shear test in DESCQA accordingly.

Issue #110, opened 2018-05-30
InstanceCatalog reader needs to be able to handle sprinkled objects


danielsf:
The InstanceCatalog reader currently assumes that the galaxy catalog contains only galaxies and AGN.  The reader ignores AGN.  We need to

a) figure out how we want to handle AGN
b) add code to handle the fact that the sprinkler is going to be adding lensed AGN and SNe to the galaxy catalog

There is also the question of how we are going to deal with objects that are passed to PhoSim as fits image profiles, instead of SEDs.  The number of columns should not change for those rows (@jbkalmbach ?), but the meaning of the spatial parameters will change (see the "Spatial Models" section here)

https://bitbucket.org/phosim/phosim_release/wiki/Instance%20Catalog

jbkalmbach:
Yeah, we are making sure that the number of columns does not change.

I'm not sure if this is helpful, but I do have a way of reading and parsing the sprinkled instance catalogs in this method: https://github.com/LSSTDESC/Twinkles/blob/issue/467/ic_validation/python/desc/twinkles/validate_ic.py#L18 that I am using to develop the instance catalog validation tests.

yymao:
@jbkalmbach thanks for the script. Yes we can implement that in the GCR for the time being. But still I think it'll be great if we can make the underlying format easier to parse :)

jbkalmbach:
@danielsf might want to say more about this, but he is working on making a way to write out the sprinkled catalog pieces separately. However, this will still necessitate an answer to his first point about handling general AGN in the catalog.

yymao:
fixed by #113

PR #113, opened 2018-06-05
Make instance catalog reader parse AGN entries

yymao:
This PR addresses at least in part, if not all, of #110. AGN entries can now be accessed from the instance catalog reader.


GCRCatalogs/instance_catalog.py
```diff
@@ -227,19 +246,21 @@ def _generate_native_quantity_list(self):
...
-            nrows = None
-            if obj_type == 'gal': # the galaxy catalog has agn in it...
+            additional_kwargs = dict()
+            if obj_type in ('gal', 'agn'): # Galaxies and AGNs are in the same file but have different columns
                 this_open = gzip.open if self._object_files[obj_type].endswith('.gz') else open
                 with this_open(self._object_files[obj_type], 'rb') as f:
                     for i, line in enumerate(f):
                         if b'agnSED/' in line:
-                            nrows = i
+                            additional_kwargs['nrows' if obj_type == 'gal' else 'skiprows'] = i
```
fjaviersanchez: Smart!

danielsf:
I would like to be able to test this on some InstanceCatalogs.  How do I run the InstanceCatalog reader?  Does it work like
```
import GCRCatalogs
cat = GCRCatalogs.load_catalog('/path/to/instance/catalog')
```
?

yymao:
@danielsf Nope, the path is part of the [config file](https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/instance_test.yaml). However, you can manully overwrite the path that was specified in the config

```python
gc = GCRCatalogs.load_catalog('instance_test', 
                              config_overwrite=dict(header_file='/path/to/instance/catalog/header/file'))
```

danielsf:
As I understand it, this PR assumes that the galaxy catalog will be strictly segregated such that all AGNs will come after all sersic components.  When we simulate full focal planes with the sprinkler, that will not necessarily be true.  The InstanceCatalog writer will iterate over 100,000 row chunks of the extragalactic catalog, writing all of the sersics for the first chunk, then all of the AGN for the first chunk, then all of the sersics for the second chunk, then all of the AGN for the second chunk, etc.

We will need to be able to handle that case, eventually.  I do not know if that is in scope for this PR or not.

danielsf:
It is also possible that lense SNe will be interspersed with the sprinkled AGN.  I am not totally clear how the sprinkler handles those.

yymao:
Indeed. We can of course address that in this PR but it'll make the load time much longer since we need to check line by line. 

Iam still hoping that we can chat the behavior of sprinkler so that we don't write rows of different columns into the same file. 

If we don't want to create a new file, we can also ask the sprinkler to match the columns in the galaxy catalog.

yymao:
@danielsf so what's the timeline for this functionality? Is it possible to change the sprinkler before we need to merge this? If not, we can make the reader more general and sacrifice some speed.

danielsf:
I might be able to get something working today.  I will let you know at the end of the day.

danielsf:
@yymao This PR should separate PhoSim objects by schema in the InstanceCatalogs

https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/52

yymao:
@danielsf great. Do you know what the file name pattern would look like (I can probably just look into the code for answer, but just in case you know the answer right away I thought I would ask first :)) 

Also, should I merge this first and create a new issue when https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/52 is merged, or should I just wait to modify this PR after https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/52 is merged? I guess I am asking if we need access to AGNs right now?

danielsf:
As the code stands, the files will be named `bulge_gal_cat_[obsHistID].txt.gz`, `disk_gal_cat_[obsHistId].txt.gz`, etc. (so: we add `bulge_`, `disk_`, and `agn_` as prefixes to the old galaxy catalog name).  That may not survive review.

To answer your other question: I don't think we need AGN now.  I think this is mostly a concern for when we want to start validating InstanceCatalog for cosmoDC2.

yymao:
OK. Then I guess I'll wait for https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/52 being merged and then update this PR accrodingly.

danielsf:
I have now merged the relevant PR in `sims_GCRCatSimInterface`.

danielsf:
The names of the `*.txt.gz` did not change relative to my comment above.

yymao:
Thanks for the notification, @danielsf. I'll update this PR accordingly.

yymao:
@danielsf are there any example output files from https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/52 that I can use to test the new code?

danielsf:
Example InstanceCatalogs are in

`/global/cscratch1/sd/danielsf/example_instance_catalogs`

Let me know if you have trouble accessing them.

yymao:
@danielsf can you `chgrp -R lsst /global/cscratch1/sd/danielsf/example_instance_catalogs` ?

danielsf:
Done (sorry about that; forgot the `-R` the first time I did that)

yymao:
@danielsf @fjaviersanchez this is now ready to be reviewed again. The updated reader now can read all types in the instance catalogs.

fjaviersanchez:
@yymao I tested it and works like a charm! Thanks a lot for the work. I just have a question: Is it intended that `GCRCatalogs.available_catalogs` gives a different output than `GCRCatalogs.get_available_catalogs()`?

yymao:
@fjaviersanchez, yes, we tried to select a set of catalogs to be displayed "by default", since there are a lot of catalogs and versions and it'll become difficult to navigate. So `get_available_catalogs` by default only include these "default" catalogs. 

`GCRCatalogs.get_available_catalogs(include_default_only=False)` should have the same output as `GCRCatalogs.available_catalogs`

GCRCatalogs/catalog_configs/instance_2145656.yaml
```diff
@@ -0,0 +1,3 @@
+subclass_name: instance_catalog.InstanceCatalog
+header_file: /global/cscratch1/sd/danielsf/example_instance_catalogs/phosim_cat_2145656.txt
```
danielsf: Do we need to move this InstanceCatalog into a project space, so that we are not generally dependent on access to my scratch drive (which, as a scratch drive, will get purged in 12 weeks)?

yymao:
@danielsf I think we should keep a few instance catalogs for each run, and we put them into yaml configs so that people can easily test/use them. We probably don't need to add *all* instance catalogs to the config files, as people can load any instance catalogs by running:

```python
from GCRCatalogs.instance_catalog import InstanceCatalog
inst_cat = InstanceCatalog(header_file='/path/to/your/favorite/instance/phosim_cat_xxxxxx.txt')
```

Do you want to move some representative instance catalogs to the group space and I can change the yaml configs accordingly?

danielsf:
@yymao I put them in
```
/global/projecta/projectdirs/lsst/groups/SSim/DC2/example_instance_catalogs> 
```
Let me know if you have any problems.  I made their permissions `550` to prevent anyone from overwriting them.

yymao:
@danielsf there are only bulges and stars in the new location. This will cause warnings since the header file specifies other objects (like disk and agn etc)

danielsf:
Sorry. I interpreted your "we shouldn't have *all* InstanceCatalogs" comment to mean that we didn't need all of the sub-catalogs for our example pointing.  Was I mistaken?

yymao:
@danielsf ah, by *all* I meant all the different numbers (which I think they mean different patches of sky or runs?). Each `phosim_cat_xxxxxx.txt` file is consider one single catalog by the reader, just have many different fields spreaded in different files.

danielsf:
Got it.  I have copied the rest of the sub-catalogs over

yymao:
@danielsf I just updated the paths in the config file. Can you approve this PR? I'll then merge it tomorrow, thanks!

PR #117, opened 2018-06-11
V4.6.1

dkorytov:
Updating the GCR for ProtoDC2 v4.6.1

v4.6.0 doesn't have shears, so it's not an official release to reduce possible confusion.


aphearin:
@erykoff - this is the reader you need to run redMaPPer on v4.6.1 - it'll be ready for you to use after @yymao merges this

yymao:
Before this is merged, you can still use this reader by doing
```sh
git clone -b v4.6.1 https://github.com/dkorytov/gcr-catalogs.git
```

erykoff:
I cloned this branch, and get the following error: `ValueError: Native quantity `mag_true_Y_lsst` does not exist (required by `mag_true_Y_lsst`)`.
All the other columns that I need appear to be there.

dkorytov:
@erykoff  - mag_true_Y_lsst is relabeled as mag_true_y_lsst. Did mag_true_Y_lsst work before?

erykoff:
Yes, the capital Y worked before, and now lower-case y works.

yymao:
@erykoff that's expected. Sorry for the backward incompatible change.

dkorytov:
@erykoff. Ah, I see the reader automatically converts the upper case band names into lower case--the hdf5 catalog file has 'y' and not 'Y'. We probably should support both upper and lower case 'y' so we don't break old code and people don't need to memorize to capitalize 'y' in the future.

yymao:
Thanks, @dkorytov. I will review this tomorrow.

yymao:
Thanks, @dkorytov. I've tested this PR as well.

PR #120, opened 2018-06-12
Update protoDC2 default version

yymao:
Update protoDC2 default version to v4.6.1. 

@evevkovacs @dkorytov @aphearin should we merge this now or wait for v4.7?


dkorytov:
There maybe a bug 4.6.1. So I'd hold off till we figure it out.

yymao:
@dkorytov you mean the reader or the catalog?

Also, just to clarify because I wasn't very clear in my original comment --- we don't need to merge this now. This PR serves as a reminder that, once we have a more stable version, we should update the default. If neither 4.6.1 and 4.7 should be promoted to the default version, we can just wait.

dkorytov:
@yymao In the catalog, the reader is fine. 

Okay, it makes sense that this isn't an immediate PR. We should wait until we figure out the fix before promoting the default catalog version.

PR #121, opened 2018-06-13
add shear_2_treecorr

yymao:
Add `shear_2_treecorr` in the readers (protoDC2, Buzzard) and the schema. This is just a duplicate of `shear_2` but we add it so that users can be explicitly using this quantity to avoid confusion.


yymao:
Thanks, @evevkovacs. I'll merge this by the end of today.

PR #123, opened 2018-06-22
UniverseMachine baseline healpix catalog reader

evevkovacs:
Add this reader to the master branch. Reader and catalog example has been tested in descqa
See here:
https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-22_11&catalog=um_hpx_v0.0


yymao:
@evevkovacs thanks for the PR! Should I go ahead to review the code or is this still developing (i.e., may change as the underlying format may change)?

evevkovacs:
@yymao It's pretty close to complete. I was reviewing why the CLF_r test failed and I think it is because I forgot to define is_central (The test doesn't check for this and it probably should). In the base catalog, I have a data column called 'upid' which is -1 for centrals. I wasn't sure how to do the mapping in self._quantity_modifiers.  Is it something like:
 'is_central': (lambda x: x.astype(np.bool), 'upid'==-1)
Thanks!

yymao:
should be
```python
    'is_central': (lambda x: x == -1, 'upid'),
```

evevkovacs:
Ah! I see how it works now. Much simpler than I thought. Thanks And I just retried the CLF test and now it crashes. But the reader is working and delivers a sensible result for is_central

evevkovacs:
@yymao I just discovered a sign error in one of the magnitude calculations. I can fix this in the reader for this version I guess, so we probably don't want to push this code to master yet as it will change for the next iteration.

yymao:
Yep, no worries. I think we should keep this in PR until more people have tested it . I've labeled this PR "developing" to indicate that.

PR #124, opened 2018-06-27
Move DC2 catalog readers from DC2_Repo to here

yymao:
To prepare for DC2 tutorials and to address https://github.com/LSSTDESC/DC2_Repo/issues/170, I move the readers (not including example notebooks) from https://github.com/LSSTDESC/DC2_Repo/pull/157 and https://github.com/LSSTDESC/DC2_Repo/pull/169 to here and also create corresponding yaml config files. 

@slosar @djperrefort can you review this? Also, please check if the catalog names (yaml filenames) make sense. Thanks!

(cc relevant parties: @wmwv @danielsf @fjaviersanchez @katrinheitmann)

### DIFF
--- a/GCRCatalogs/dc2_coadd.py
+++ b/GCRCatalogs/dc2_coadd.py
@@ -1,5 +1,5 @@
 """
-DC2 Static Merged Coadd Catalog reader
+DC2 Coadd Catalog Reader
 """
 
 import os
@@ -10,11 +10,11 @@ import tables
 import pandas as pd
 from GCR import BaseGenericCatalog
 
-__all__ = ['DC2StaticMergedCoaddCatalog']
+__all__ = ['DC2CoaddCatalog']
 
-class DC2StaticMergedCoaddCatalog(BaseGenericCatalog):
+class DC2CoaddCatalog(BaseGenericCatalog):
     r"""
-    DC2 Static Merged Coadd Catalog reader
+    DC2 Coadd Catalog reader
 
     Parameters
     ----------
--- a/GCRCatalogs/dc2_coadd.py
+++ b/GCRCatalogs/dc2_coadd.py
@@ -27,7 +27,7 @@ class DC2CoaddCatalog(BaseGenericCatalog):
     _native_filter_quantities = {'tract', 'patch'}
 
     def _subclass_init(self, **kwargs):
-        self._base_dir = kwargs.get('base_dir', '/global/projecta/projectdirs/lsst/global/in2p3/Run1.1/summary')
+        self._base_dir = kwargs['base_dir']
         self._filename_re = re.compile(kwargs.get('filename_pattern', r'merged_tract_\d+\.hdf5'))
         self._groupname_re = re.compile(kwargs.get('groupname_pattern', r'coadd_\d+_\d\d$'))
         self.use_cache = bool(kwargs.get('use_cache', True))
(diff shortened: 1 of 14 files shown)


yymao:
@slosar thanks for the review. To answer your minor comments:
1. The hdf5 files of merged coadd catalogs are generated with `pandas` so we use `pandas` in the reader to ensure compatibility. Since `pandas` is installed on the shared DESC python environment, I think it's ok to use it.
2. We probably should name the catalogs better, and that's what I want to hear people's suggestions. For example, `dc2_input_v3_fov4` is probably not a great name. (On a related note, is this actually a "reference" catalog or an "input" catalog?)

yymao:
Per the discussion of the Data Access TF, we should rename "input catalog" to "reference catalog" and also rename "DC2 Static Merged Coadd Catalog" to just "DC2 Coadd Catalog". 

I'll do that once i get off my telecons.

GCRCatalogs/dc2_coadd.py
```diff
@@ -0,0 +1,130 @@
...
+    base_dir: str
+    filename_pattern: str, optional (default: r'merged_tract_\d+\.hdf5')
+    groupname_pattern: str, optional (default: r'coadd_\d+_\d\d$')
+    use_cache : bool, optional (default: True)
+    """
+
+    _native_filter_quantities = {'tract', 'patch'}
+
+    def _subclass_init(self, **kwargs):
+        self._base_dir = kwargs.get('base_dir', '/global/projecta/projectdirs/lsst/global/in2p3/Run1.1/summary')
```
djperrefort: This line defines the default base directory, however you also have a base directory specified in the .yaml file. It would be simpler to maintain if the reader only used one or the other.

GCRCatalogs/reference_catalog.py
```diff
@@ -0,0 +1,79 @@
...
+                    break
+                yield data.__getitem__
+                chunk_count += 1
+
+
+    def _generate_native_quantity_list(self):
+        line = None
+        with open(self._filename, 'r') as f:
+            for i, line in enumerate(f):
+                if len(line) > 2 and line[0] == '#'  and line[1] not in (' ', '\n'):
```
danielsf: I'm worried that this is a little fragile against someone making the decision that they want the header line to look like `# col_name1, col_name2...` as opposed to `#col_name1, col_name2, ...` (as it does now).

Would it be safer to just choose the last line that starts with `#`?

Or look for a line that contains `uniqueId`, `raJ2000`, and `decJ2000` (all known column names)?

I'm just being paranoid.  I don't actually have strong feelings about this.

danielsf:
You asked me to look at this PR, Yao.  It looks fine to me.

yymao:
@danielsf I've updated the header search to look for `uniqueId`. But to be honest, ascii files are fragiles anyway, we just need to remember to update the reader when we update the catalog creation pipeline.

yymao:
@wmwv I've updated the path

yymao:
@danielsf do you think there's a better name than `dc2_reference_v3_fov4`?

danielsf:
The only thing I could imagine dropping is the `fov4`, since clearly a 4 degree field of view should be the default.  The `v3` needs to be there to avoid confusion with the reference catalog for run 1.1p, which was based on a different version of protoDC2, with a different distribution of sources.

yymao:
@danielsf can I have your approval again? In particular, the last commit renamed 4 yaml files to be more clear. The new names are:
```
dc2_coadd_run1.1p
dc2_reference_run1.2p
dc2_instance_example1
dc2_instance_example2
```

Thanks!

yymao:
Thanks, @danielsf. I'll merge this tomorrow.

PR #126, opened 2018-06-30
Coadd reader

djperrefort:
This pull request adds additional values to the DC2 static coadd reader. A table of the mapped values is included below. Some things of note:

1. All values listed below for *u* band have analogous columns in each of the *ugrizy* bands.
1. `ICov` provides the 6 unique values of the **cross-band**, 3 x 3 covariance matrix for the second moments (`Ixx`, `Iyy`, `Ixy`).
1. `u_ICov` provides the 6 unique values of the **band specific**, 3 x 3 covariance matrix for the second moments (`u_Ixx`, `u_Iyy`, `u_Ixy`).
1. The second moments mapped by the catalog reader are calculated using the HSM implementation. I didn't see propagated errors for these, so I used errors from the SDSS implementation when calculating the covariance matrices.

|Homonogized Value | Native Value|
|------------------|-------------|
| objectId | id |
| parentObjectId | parent |
| ra | coord_ra (converted to degrees)|
| dec | coord_dec (converted to degrees)|
| centroidX | slot_Centroid_x |
| centroidY | slot_Centroid_y |
| centroidX_err | slot_Centroid_xSigma |
| centroidY_err | slot_Centroid_ySigma |
| centroid_flag | slot_Centroid_flag |
| psNdata | base_PsfFlux_area |
| extendedness | base_ClassificationExtendedness_value |
| I_flag | slot_Shape_flag |
| ICov | **Calculated by reader** |
| Ixx | slot_Shape_xx |
| IxxPSF | slot_PsfShape_xx |
| Iyy | slot_Shape_yy |
| IyyPSF | slot_PsfShape_yy |
| Ixy | slot_Shape_xy |
| IxyPSF | slot_PsfShape_xy |
| u_magLSST | u_mag |
| u_magLSST_err | u_mag_err |
| u_psFlux | u_slot_ModelFlux_flux |
| u_psFlux_flag | u_slot_ModelFlux_flag |
| u_psFlux_err | u_slot_ModelFlux_fluxSigma |
| u_I_flag | u_slot_Shape_flag |
| u_ICov | **Calculated by reader** |
| u_Ixx | u_slot_Shape_xx |
| u_IxxPSF | u_slot_PsfShape_xx |
| u_Iyy | u_slot_Shape_yy |
| u_IyyPSF | u_slot_PsfShape_yy |
| u_Ixy | u_slot_Shape_xy |
| u_IxyPSF | u_slot_PsfShape_xy |


yymao:
@djperrefort thank you for the PR! This looks great. I renamed the reader so now this PR doesn't add a new one but modify the existing one, and some minor coding style changes. I also added a few more quantity modifiers, some are just aliases that we want to use in DESCQA, and others are some quantities I saw @EiffL used before. 

@EiffL and @fjaviersanchez can you test this reader while you prepare your tutorials? It would be great to have your input. Thanks!

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+                warnings.warn(war_msg.format(dataset_info['tract'],
+                                             dataset_info['patch'],
+                                             dataset[0]))
+                continue
+
             def _native_quantity_getter(native_quantity):
                 # pylint: disable=W0640
                 if native_quantity in self._native_filter_quantities:
                     return np.repeat(dataset_info[native_quantity], len(d))
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
                 elif native_quantity not in d:
                     return np.repeat(np.nan, len(d))
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+        return self.load_dataset((file_path, group_id))
+
+
     def _generate_native_quantity_list(self):
+        """Return a set of native quantity names as strings"""
         return self._columns.union(self._native_filter_quantities)
 
+
     def _iter_native_dataset(self, native_filters=None):
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+        Returns:
+            A pandas data frame
+        """
         if not self.use_cache:
             return self._load_dataset(dataset)
+
         if dataset not in self._dataset_cache:
             try:
                 self._dataset_cache[dataset] = self._load_dataset(dataset)
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
-                            break
-                        datasets_this.append((fpath, key))
-                        columns_this.update((c.decode() for c in fh.root[key].axis0))
-                    else:
-                        datasets.extend(datasets_this)
-                        columns.update(columns_this)
+                datasets_this, columns_this = self._read_hdf5_meta(fpath)
+                datasets.extend(datasets_this)
+                columns.update(columns_this)
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+                    warn_msg = '{} has incorrect hdf5 format; skipped'
+                    warnings.warn(warn_msg.format(os.path.basename(fpath)))
+                    return list(), set()
+
+                data_sets.append((fpath, key))
+                columns.update((c.decode() for c in group.axis0))
+
+        return data_sets, columns
+
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+        return modifiers
+
+
+    @property
+    def quantity_modifiers(self):
+        """Return the mapping from native to homogonized value names as a dict"""
+        return self._quantity_modifiers
+
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+            modifiers['{}_psf_size'.format(band)] = (
+                lambda xx, yy, xy: 0.168*2.355*(xx*yy - xy*xy)**0.25,
+                '{}_base_SdssShape_psf_xx'.format(band),
+                '{}_base_SdssShape_psf_yy'.format(band),
+                '{}_base_SdssShape_psf_xy'.format(band),
+            )
+
+        return modifiers
+
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+        Args:
+            dataset (tuple): Of the form (<file path (str)>, <group id (str)>)
+
+        Returns:
+            A dictionary {"tract": <tract (int)>, "patch": <patch (int)>}
+        """
         items = dataset[1].split('_')
         return dict(tract=int(items[1]), patch=','.join(items[2]))
 
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+            A list of available patches
+        """
+        patches = []
+        for dataset in self._datasets:
+            dataset_info = self.get_dataset_info(dataset)
+            if dataset_info['tract'] == int(tract):
+                patches.append(dataset_info['patch'])
+
+        return patches
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
 
     @property
     def base_dir(self):
+        """The directory where data files are stored"""
         return self._base_dir
 
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
     def clear_cache(self):
+        """Empty the catalog reader cache and frees up memory allocation"""
         self._dataset_cache.clear()
 
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+        Args:
+            dataset (tuple): Of the form (<file path (str)>, <group id (str)>)
+
+        Returns:
+            The contents of the specified file's group
+        """
         return pd.read_hdf(*dataset, mode='r')
 
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
             try:
                 self._dataset_cache[dataset] = self._load_dataset(dataset)
+
             except MemoryError:
                 self.clear_cache()
                 self._dataset_cache[dataset] = self._load_dataset(dataset)
+
         return self._dataset_cache[dataset]
 
+
```
wmwv: Remove extra line

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+            err_msg = 'Invalid patch {} for tract {}'
+            raise ValueError(err_msg.format(patch, tract))
+
+        file_name = FILE_PATTERN.replace(r'\d+', str(tract)).replace(r'\.', '.')
+        file_path = os.path.join(self.base_dir, file_name)
+        group_id = GROUP_PATTERN.replace(r'\d+', str(tract)).replace(r'\d\d', patch.replace(',', ''))
+
+        return self.load_dataset((file_path, group_id))
+
+
```
wmwv: Remove extra line

wmwv:
@djperrefort Good work to figure out the `I_cov` stuff.  But, until we have consistency with HSM shape measurements that include uncertainty, we should report NaNs for `I_cov`.  Keep the function in, but we won't use it until we have actual values.

wmwv:
@djperrefort For your convenience, here are the PEP8 recommendations for lines:
https://www.python.org/dev/peps/pep-0008/?#blank-lines

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
         self._dataset_cache = dict()
 
-        self._quantity_modifiers = {
+        self._quantity_modifiers = self._generate_modifiers()
+        self._datasets, self._columns = self._generate_datasets_and_columns()
+        if not self._datasets:
+            err_msg = 'No catalogs were found in `base_dir` {}'
+            raise RuntimeError(err_msg.format(self._base_dir))
+
+
```
wmwv: Remove extra line.

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,97 +85,330 @@ def _subclass_init(self, **kwargs):
...
+                datasets_this, columns_this = self._read_hdf5_meta(fpath)
+                datasets.extend(datasets_this)
+                columns.update(columns_this)
+
             except (IOError, OSError):
                 warnings.warn('Cannot access {}; skipped'.format(fpath))
+
         return datasets, columns
 
+
```
wmwv: Remove extra line.

yymao:
@wmwv sorry, I added that extra blank lines to be consistent with other code in this repo. But it's ok with me to remove them.

yymao:
@wmwv Can you elaborate a bit on your "Future" question. I think the code here can be used directly as a DB ingestion script, so am not sure which part you'd like to expose?

djperrefort:
@mwv I have pep 8 and pep 256 pre-loaded into my Pycharm development environment. I run the inspections in real time during development.

wmwv:
@yymao With the caveat that I don't fully understand all of the capabilities of `GCR`.
1) A direct ingest script might not want to read the per-tract-patch coadd files.  It may choose to use the reading and merging logic in `DC2_Repo/scripts/merge_tract_patch.py` and then apply the logic in `_generate_modifiers` directly to whatever in-memory object it's hard or has created.
2) Someone might have existing analysis code that works well for them, and they want to know how to adapt it to use the conventions in the DPDD.  That information is here in `_generate_modifiers`, but it's not clear to me how to trivial programatically extract it.  Put another way, "we" might someday offer the service of converting other people's scripts to be updated as we update the translation table here.

wmwv:
@djperrefort Ah, I see.  By making these `@staticmethod`s, you've addressed a significant chunk of exposing the interface in at least some form.  That's great for now.

djperrefort:
@wmwv That was the intent, although some of the decorators were already introduced in the original version written by @yymao.

In regards to exposing the information in `_generate_modifiers`, the end user can access that information as a dictionary using the  `quantity_modifiers` attribute. The information is protected against being overwritten, but a user wishing to build their own customized script could use it as a starting point.

wmwv:
@djperrefort   Is the specific `GCRCatalogs/catalog_configs/dc2_coadd_run1.1p_tract4850.yaml` meant to be used in an example or test?

yymao:
@wmwv, again, that comes from me (`git blame` can tell you to blame me). `dc2_coadd_run1.1p_tract4850.yaml` is meant for testing, downstream code developing, and some tutorial notebooks --- basically any cases when you don't want to the code to take too long to run.

wmwv:
@yymao Sounds good and makes sense.  Can you add that to the top-level README.md, or even just a README in this subdirectory.

yymao:
Done!

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,91 +84,303 @@ def _subclass_init(self, **kwargs):
...
+            'centroid_flag': 'slot_Centroid_flag',
+            'psNdata': 'base_PsfFlux_area',
+            'extendedness': 'base_ClassificationExtendedness_value',
+            'HSM_res': 'ext_shapeHSM_HsmShapeRegauss_resolution',
+            'HSM_ell': (np.hypot, 'ext_shapeHSM_HsmShapeRegauss_e1', 'ext_shapeHSM_HsmShapeRegauss_e2'),
         }
 
+        modifiers['pass_basic_flag_cut'] = (
+            create_basic_flag_mash,
+            'deblend_skipped',
```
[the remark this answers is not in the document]
wmwv: The flat file should only include `isPrimary` objects.

There was a bug that meant it did include some non-`isPrimary` objects in the first prototype.  That's been fixed.

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,91 +84,303 @@ def _subclass_init(self, **kwargs):
...
+            'base_PixelFlags_flag_edge',
+            'base_PixelFlags_flag_interpolatedCenter',
+            'base_PixelFlags_flag_saturatedCenter',
+            'base_PixelFlags_flag_crCenter',
+            'base_PixelFlags_flag_bad',
+            'base_PixelFlags_flag_suspectCenter',
+            'base_PixelFlags_flag_clipped',
+            'ext_shapeHSM_HsmShapeRegauss_flag',
```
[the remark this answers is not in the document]
wmwv: I think our focus is on DC2 for now.  I hear your point about maintain something general, and perhaps there's some inheritance one could set up, but there will also be these small differences in different runs that we want to encode in our reader and not make each user have to know.

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,91 +84,303 @@ def _subclass_init(self, **kwargs):
...
+            'centroidX_err': 'slot_Centroid_xSigma',
+            'centroidY_err': 'slot_Centroid_ySigma',
+            'centroid_flag': 'slot_Centroid_flag',
+            'psNdata': 'base_PsfFlux_area',
+            'extendedness': 'base_ClassificationExtendedness_value',
+            'HSM_res': 'ext_shapeHSM_HsmShapeRegauss_resolution',
+            'HSM_ell': (np.hypot, 'ext_shapeHSM_HsmShapeRegauss_e1', 'ext_shapeHSM_HsmShapeRegauss_e2'),
         }
 
+        modifiers['pass_basic_flag_cut'] = (
```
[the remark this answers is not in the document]
wmwv: `good`

GCRCatalogs/dc2_coadd.py
```diff
@@ -12,16 +12,66 @@
...
+    generate a mask for a set of flags
+    for each item, mask will be true if and only if all flags are false
+    """
+    out = np.ones(len(flags[0]), np.bool)
+    for flag in flags:
+        out &= (~flag)
+    return out
+
+
 class DC2CoaddCatalog(BaseGenericCatalog):
```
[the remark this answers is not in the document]
wmwv: For this PR we should keep this, but I agree that if we start using HSC significantly it will make sense to make a more generic class that we inherit from.

GCRCatalogs/dc2_coadd.py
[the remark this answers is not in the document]
wmwv: "reasonable" is more the sense, but that's too long to type.
"pass_basic_flag_cut" seems descriptive, but it's not particularly, because you still have to go look up what those basic cuts were.  So I think "good" might capture that as well.

The current star-galaxy separation isn't very good -- a significant number of barely resolved galaxies are classified as stars -- so we might just stick with `good` for now.

GCRCatalogs/dc2_coadd.py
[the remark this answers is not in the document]
wmwv: But I realize now, that perhaps we should remove that flag because it may not do the right thing for stars.

GCRCatalogs/dc2_coadd.py
```diff
@@ -12,16 +12,66 @@
...
+    Args:
+        ixx_err (float): The error in the second moment Ixx
+        iyy_err (float): The error in the second moment Iyy
+        ixy_err (float): The error in the second moment Ixy
+
+    Returns:
+        The covariance between the given arrays
+    """
+    # If need be, this can be generalized using a mesh grid
+    out_data = np.array([
```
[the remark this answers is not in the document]
djperrefort: The DPDD intends `ICov` as a six-element array containing the six unique elements of the cross-band or reference band covariance matrix. In the coadd data I'm taking this to refer to whatever band was used in determining the shape slot values for a given object.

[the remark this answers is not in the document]
wmwv: I believe @EiffL was asking about the order of the array.  Is it
1. `[xx_xx, xy_xy, yy_yy, xx_xy, xy_yy, xx_yy]`
or
2. `[xx_xx, xx_xy, xx_yy, xy_xy, xy_yy, yy_yy]`

As @EiffL mentions, this order is not defined in the DPDD.  It will be defined in any particular schema, and we should do it in our own.  I would argue for 1.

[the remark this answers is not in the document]
djperrefort: @EiffL and @wmwv I've just changed ICov it to adopt the second order `[xx_xx, xx_xy, xx_yy, xy_xy, xy_yy, yy_yy]`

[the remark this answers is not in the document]
wmwv: Could you add the documentation of the order it returns to the function docstring?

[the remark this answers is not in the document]
EiffL: great! thanks :-) Yes, my question was more about the order, this looks good

GCRCatalogs/dc2_coadd.py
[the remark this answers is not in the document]
EiffL: +1 for `good`

GCRCatalogs/dc2_coadd.py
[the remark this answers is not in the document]
EiffL: it may fail for some stars, not sure... but the point is that this flag is mostly useful if you are going t use the HSM shapes, like for a lensing sample, it may introduce unnecessary/unwanted selections for other use cases

GCRCatalogs/dc2_coadd.py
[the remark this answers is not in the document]
EiffL: Just making sure @fjaviersanchez is aware of this

GCRCatalogs/dc2_coadd.py
```diff
@@ -5,23 +5,77 @@
...
+
+    # If need be, this can be generalized using a mesh grid
+    out_data = np.array([
+        ixx_err * ixx_err,
+        ixx_err * ixy_err,
+        ixx_err * iyy_err,
+        ixy_err * ixy_err,
+        ixy_err * iyy_err,
+        iyy_err * iyy_err
+    ])
```
wmwv: This calculation is not correct.  We need the original covariance values calculated by the pipeline.  These are dependent on the counts in the pixels; we can't reconstruct them later solely from the diagonal terms.

yymao:
I renamed `pass_basic_flag_cut` to `good`. @EiffL, can you help define `good_galaxy` and `good_star`?

yymao:
@djperrefort Can you point out the new quantities that you defined that are not actually in DPDD? (Sorry, I know I can look this up myself; I am just lazy..)

djperrefort:
@yymao Values not defined in the DPDD include the centroid values and values prefixed with a band. The DPDD also defines a single `flags` array, which we have broken up into multiple values.

wmwv:
@djperrefort Can you explicitly list these columns in the documentation?  You don't have to list all several thousand native quantities, just the generalized ones that GCR is presenting.

EiffL:
@yymao I have mixed feellings about defining a good_galaxy and good_star flag, in principle, it's just a matter of using the `extendedness` flag, but that selection is not perfect (as pointed out by Michael), I'm afraid that a simplified `good_star` flag like this would mislead people into thinking that this flag selects a robust sample of stars. I would leave it as is.

yymao:
@EiffL I see what you're saying, but for people like me who has little idea on how to select galaxies, having a more descriptive column helps a lot.  How about `likely_galaxy` and 'likely_star`?

EiffL:
Hummmm sorry, no, I think telling people to use the `extendedness` flag if they want to select stars is a better idea, like  `good & (extendedness == 0)`  is better. The `good` flag is only telling us that the source is not containing any corrupted pixels, it's a set of essential flags people agree on. 
But defining a star/galaxy flag is more tricky and should be left to the analysis working groups, if they come up with a standard set of cuts (like the weak lensing flag in HSC Y1) we can then implement it here for convenience.

wmwv:
I agree with `good & (extendedness == 0)` for now.  `extendedness` is currently just either 0 or 1 (it will eventually be continuous, but that will likely come along with some good improvements in the star/galaxy separation).

yymao:
OK, thanks @EiffL @wmwv. It sounds like we've addressed everything? Do we want to remove `i_magLSST` and stick with `mag_i_lsst` given that neither is in DPDD but the latter is in GCRCatalogs' schema? (cc @djperrefort)

EiffL:
There is still the covariance for the moments issue, as Michael pointed out

wmwv:
Yes, `mag_i_lsst`.

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,91 +88,291 @@ def _subclass_init(self, **kwargs):
...
+            modifiers['{}_psFlux_err'.format(band)] = '{}_slot_ModelFlux_fluxSigma'.format(band)
+
+            # Band specific second moment values
+            modifiers['{}_I_flag'.format(band)] = '{}_slot_Shape_flag'.format(band)
+
+            for ax in ['xx', 'yy', 'xy']:
+                modifiers['{}_I{}'.format(band, ax)] = '{}_slot_Shape_{}'.format(band, ax)
+                modifiers['{}_I{}PSF'.format(band, ax)] = '{}_slot_PsfShape_{}'.format(band, ax)
+
+            modifiers['{}_mag_CModel'.format(band)] = (
```
yymao: Since this is not part of DPDD, we can rename it to `mag_{}_cmodel` so that it follows the same convention as `mag_{}_lsst`, @EiffL?

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,91 +88,291 @@ def _subclass_init(self, **kwargs):
...
+                lambda x: -2.5 * np.log10(x) + 27.0,
+                '{}_modelfit_CModel_flux'.format(band),
+            )
+
+            modifiers['{}_SN_CModel'.format(band)] = (
```
yymao: Same. Rename it to `sn_{}_cmodel`?

GCRCatalogs/dc2_coadd.py
```diff
@@ -34,91 +88,291 @@ def _subclass_init(self, **kwargs):
...
+                np.divide,
+                '{}_modelfit_CModel_flux'.format(band),
+                '{}_modelfit_CModel_fluxSigma'.format(band),
+            )
+
+            modifiers['{}_psf_size'.format(band)] = (
```
yymao: Same. Rename it to `psf_size_{}`?

fjaviersanchez:
@djperrefort @wmwv @yymao @EiffL. Do you think that we make explicit that the `mag_*_lsst` and their errors are PSF mags and not CModel mags? (I guess that this is implicitly specified since you have a CModel mag field) Also, I think it would be good to have CModel mag errors as well (which aren't included now). What's your opinion on this?

yymao:
Meh for first and yes for the second.

wmwv:
@djperrefort Please change to `SNR_CModel_*` rather than `SN_...`

fjaviersanchez:
oh, I didn't realize that `SN_CModel` existed,, thanks! With the SNR it is enough for most users, I think.

yymao:
@djperrefort i just started a new table at the end of [`SCHEMA.md`](https://github.com/djperrefort/gcr-catalogs/blob/coadd_reader/GCRCatalogs/SCHEMA.md#schema-for-dc2-coadd-catalogs), can you complete this table? Thank you!

djperrefort:
Should we omit error values such as `magerr_<band>_lsst` and let their meaning be inferred by `mag_<band>_lsst`?

yymao:
`magerr_<band>_<filter>` is already defined in the extragalactic catalogs' section.

wmwv:
To answer the general question, each newly defined value here be listed, including columns that are the errors of another column.  For example, some columns don't have errors -- rather than negatively indicating each such column, it's easier just to list exactly the newly defined column.

djperrefort:
@EiffL Can you clarify the definition and units of `psf_size_<band>`? I don't recognize any of the numerical values.

    modifiers['psf_size_{}'.format(band)] = (
            lambda xx, yy, xy: 0.168 * 2.355 * (xx * yy - xy * xy) ** 0.25,
            '{}_base_SdssShape_psf_xx'.format(band),
            '{}_base_SdssShape_psf_yy'.format(band),
            '{}_base_SdssShape_psf_xy'.format(band),
       )

EiffL:
sure, but I would maybe suggest dropping that also from the reader, it's only one way of computing the PSF size, people only need the moments, then they can compute whatever size they want.
So, the 0.168 is the pixel scale (at least in HSC.... I'm double checking for LSST, @wmwv might know), 2.355 is a fixed conversion factor to get the PSF FWHM.

yymao:
@EiffL if you will be using this quantity in the tutorial, I'd rather keep it in the reader (but maybe with a better name).

EiffL:
Ok, so apparently, the LSST pixel scale is 0.2 arcsec, not 0.168, at least on online LSST documents.
We should at least call it `psf_fwhm_{band}`, and that pixel scale value should be a parameter in the config file then.

EiffL:
blendedness will be very important, yes. that's a good point, it could be added to the standard schema.

EiffL:
It should point to `base_Blendedness_abs_flux` for now

yymao:
@wmwv just added them as properties of the instance (i.e., `self`).

wmwv:
@yymao last comment -> https://github.com/LSSTDESC/gcr-catalogs/issues/128

fjaviersanchez:
I created [this notebook](https://gist.github.com/fjaviersanchez/8045435cc4e08c9a5ba2cf2e5b55b138) to test the reader. I think the reader works well and loads the data pretty quickly. There aren't any major missing quantities for WL and LSS analyses (beyond metadata: seeing, sky-brightness, etc, which I don't think is going to be very popular among users and I think it is out of scope for us) that I could come up with. Kudos @djperrefort @wmwv and @yymao.

GCRCatalogs/SCHEMA.md
```diff
@@ -86,3 +88,38 @@ Quantity Label | Unit | Definition
...
+`IyyPSF_<band>_lsst` | asec2 | Adaptive second moment for the PSF  in `<band>`. 
+`IxyPSF_<band>_lsst` | asec2 | Adaptive second moment for the PSF  in `<band>`. 
+`I_flag_<band>_lsst` | - | Flag for issues with `Ixx_<band>_lsst`, `Ixx_<band>_lsst`, and `Ixx_<band>_lsst.` 
+`mag_<band>_CModel` | mag | Apparent magnitude in `<band>`, fitted by CModel.
+`magerr_<band>_CModel` | mag | Error value for `mag_<band>_cmodel.` 
+`SNR_<band>_CModel` | - | Signal to noise ratio for magnitude in `<band>`, fitted by CModel.
+` psf_fwhm_<band>` | pixels | The full width at half maximum of the PSF 
+`good` | - | Whether the source contains any corrupted pixels. 
+`I_flag` | - | Flag for issues with `Ixx`, `Iyy`, and `Ixy`. 
+`blendedness` | - |
```
fjaviersanchez: Here we can say that it maps to `base_Blendedness_abs_flux`. There's a description in Section 4.9.11 of https://arxiv.org/pdf/1705.06766.pdf. I don't know if this is the same variable in the paper though. Maybe @TallJimbo or @RobertLuptonTheGood can provide some details?

EiffL: From the documentation of the schema:
`measure of how flux is affected by neighbors: (1 - flux.child/flux.parent)`
I think it's still the same as in HSC

TallJimbo: Yes, this is the exact quantity described in the HSC Pipeline paper.

EiffL:
sounds good to me!

wmwv:
@yymao   Great.  Merge it.

(I don't have permission, which I presume means @djperrefort doesn't either.)

wmwv:
Thanks, @yymao 
Nice work, @djperrefort !
Thanks, @EiffL, for the active and detailed reviewing
Thanks, @fjaviersanchez , for the great testing Notebook.  I love that we can clearly see the discreteness of the cooler stellar types in the simulations.  (the blobs on the vertical part of the g-r, r-i distribution).

fjaviersanchez:
@wmwv thanks! I was actually wondering what those blobs were.

PR #129, opened 2018-07-05
update reference catalog reader

yymao:
This PR updates the reference catalog reader to add some quantities that are defined in the schema.

### DIFF
--- a/GCRCatalogs/reference_catalog.py
+++ b/GCRCatalogs/reference_catalog.py
@@ -34,6 +34,7 @@ class ReferenceCatalogReader(BaseGenericCatalog):
         self._max_chunks = None if self._max_chunks is None else int(self._max_chunks)
 
         self._quantity_modifiers = {
+            'object_id': 'uniqueId',
             'ra' : 'raJ2000',
             'dec' : 'decJ2000',
             'sigma_ra' : 'sigma_raJ2000',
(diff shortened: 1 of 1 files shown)


fjaviersanchez:
@yymao I think that as long as all columns are listed in `list_all_quantities()` it should be fine. One useful quantity for some analysis is going to be `isresolved` (and maybe `isagn`), as long that this one is visible somehow and in the schema, there's no reason to add more modifiers.

yymao:
Question for @fjaviersanchez: should `mag_u_lsst` mapped to `lsst_u` or to `lsst_u_smeared`?

fjaviersanchez:
Honestly, I don't really know. If I understood correctly @danielsf's comment this morning, `lsst_u_smeared` is passed to the DM stack to calibrate the magnitudes/compute the zeropoints, so I guess this should be the one that we have to compare against? But I don't think it is going to make a big difference. @danielsf or @jchiang87 please, correct me if this is wrong.

danielsf:
The various `*_smeared` quantities have Gaussian noise applied to them to simulate that an actual reference catalog (which will be derived from another survey) won't perfectly measure the physical truth.

If we want people to use the reference catalog reader to access the reference catalog in the way it was intended to be used (i.e. as a stand in for data from another survey), then yes, I think the `*_smeared` quantities should be the ones that are aliased to the defaults.

yymao:
@danielsf, thanks! That answers my question. I think we want the ones with `_smeared` to be mapped to the "default" quantities. I've just updated it.

yymao:
Will merge this on Monday unless there's objection.

PR #133, opened 2018-07-06
Fix issues #131 #132 #136 for coadd reader

djperrefort:
Addresses Issue #131 and fixes the mapping of band-specific PSF values to use `_base_` instead of `_slot_` values.

[edited by @yymao: fix #131, fix #132, fix #136]

PR #134, opened 2018-07-06
add mag_<band> as aliases for mag_<band>_lsst

### LINKED ISSUE #132: drop `_lsst` in coadd reader; add `mag_<band>` to extragalactic catalog readers
We have been using `mag_<band>_<filter>` in the extragalactic catalog readers as there are `mag_g_lsst` and `mag_g_sdss` etc. For DC2 catalogs, the filter postfix seems unnecessary, so we will drop it from the coadd reader. And for extragalactic catalog readers, `mag_<band>` will be an alias of `mag_<band>_lsst`.

We will modify the schema and the readers accordingly.

yymao:
This PR, together with #133, address #132.

We've decided that, when the filter postfix is not present, it falls back to `lsst` filter.


yymao:
Thanks, @evevkovacs.

Will merge this on Monday unless there's objection.

PR #137, opened 2018-07-09
update front-page README

yymao:
In light of the recent changes in the readers, catalogs, and NERSC instructions, I've updated the front-page README. [Here's a preview](https://github.com/LSSTDESC/gcr-catalogs/blob/u/yymao/update-readme/README.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,121 +1,180 @@
 # GCR Catalogs
 
-[![Paper DOI](https://img.shields.io/badge/Paper%20DOI-10.3847%2F1538--4365%2Faaa6c3-brightgreen.svg)](https://doi.org/10.3847/1538-4365/aaa6c3)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,121 +1,181 @@
...
+NERSC jupyter-dev environment (*you only need to do this once*).
+To do so, log in to cori.nersc.gov and run:
+```bash
+source /global/common/software/lsst/common/miniconda/kernels/setup.sh
+```
+Detailed instructions can also be found at [this Confluence page](https://confluence.slac.stanford.edu/x/1_ubDQ)
+(*DESC members only*).
+
+Then, you can [start a NERSC notebook server](https://jupyter-dev.nersc.gov)
+and open a notebook with the `desc-python` or `desc-stack` kernel.
```
wmwv: @yymao [Off topic]: Should we train ourselves to use these specific names for the kernels when preparing the Notebook tutorials?  The default kernel gets saved in the Notebook and it was be an easing convenience to the user to not flag up a box making them select a kernel.

yymao: Yes, @heather999 and I are working on a more convenient way of setting up these kernels and now if people follow the one-line instruction, these kernels will be called `desc-python` and `desc-stack`. So I think we should use these kernel names in the tutorial notebooks.

PR #138, opened 2018-07-09
Adds reader for HSC flat catalog

### LINKED ISSUE #135: add config for HSC coadd catalogs
@EiffL is preparing the HSC coadd catalogs that can use the same coadd reader. @EiffL will add a config file once the HSC coadd catalogs are ready. 

In the case that we need to modify the HSC coadd catalog reader slight, we can inherit the DC2 coadd reader.

EiffL:
This PR simply adds a configuration file which reuses the `DC2CoaddCatalog` to load the HSC coadd catalogs. Everything seems to work, except for the fact that we are missing the U-band. 
Do we want a different reader to handle this case, or should we adapt the DC2CoaddCatalog reader to not expose the u-band (or any band) if it's not available in the data ?

[edited by @yymao: this PR fixes #135]

### DIFF
--- a/GCRCatalogs/catalog_configs/hsc-pdr1-xmm.yaml
+++ b/GCRCatalogs/catalog_configs/hsc-pdr1-xmm.yaml
@@ -1,0 +1,5 @@
+subclass_name: dc2_coadd.DC2CoaddCatalog
+base_dir: /global/cscratch1/sd/jchiang8/desc/HSC/
+description: XMM field of the HSC PDR1
+creators: ['Hyper Suprime-Cam (HSC) collaboration']
+included_by_default: true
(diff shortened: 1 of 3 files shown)


yymao:
What's the behavior of @wmwv's script when dealing with missing bands? Does it still populate those fields with some default values or are those fields missing the in the final hdf5 files?

EiffL:
The HDF5 script just ignores these bands, they are not added to the flat file, they are missing from the HDF5 catalog

yymao:
OK, @EiffL, can you test to see if the change works for both HSC and the DC2 Coadd catalogs?

EiffL:
Ok, yep :-) It works ! Thanks Yao !

PR #139, opened 2018-07-09
update to v0.8.0-alpha

yymao:
While there are still a few things to be done before we release v0.8.0, I'd like to tag and release `v0.8.0-alpha` and install it in the DESC shared environment, so that the tutorial writers can use it directly. 

What do you think @wmwv @heather999?


yymao:
I'm going to be a dictator and make a v0.8.0-alpha pre-release.

PR #141, opened 2018-07-11
use string-like native filter as where conditions

yymao:
[GCR v0.6.4](https://github.com/yymao/generic-catalog-reader/releases/tag/v0.6.4) adds an `_allow_string_native_filter` option, which when enabled, the user can supply a list of WHERE statments in as `native_filters` in `get_quantities`. 

This PR utilizes this feature in the DC1 reader.

### DIFF
--- a/GCRCatalogs/dc1.py
+++ b/GCRCatalogs/dc1.py
@@ -19,6 +19,8 @@ class DC1GalaxyCatalog(BaseGenericCatalog):
     """
     DC1 galaxy catalog class.
     """
+    
+    _allow_string_native_filter = True
 
     def _subclass_init(self, **kwargs):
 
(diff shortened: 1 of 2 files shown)


yymao:
Thanks, @danielsf. You can consider using this feature too in the truth catalog reader.

Will merge this by the end of today.

danielsf:
Thanks.  If only I'd remembered that Cori was going to be down today...

GCRCatalogs/dc1.py
```diff
@@ -19,6 +19,8 @@ class DC1GalaxyCatalog(BaseGenericCatalog):
     """
     DC1 galaxy catalog class.
     """
+    
+    _allow_string_native_filter = True
```
danielsf: Sorry... now that I am actually trying to write a GCR reader, I realized I have a question: what is the functional significance of this boolean?  It does not appear to be reference anywhere else that I can see

yymao: In the original design, `native_filters` should be a list of tuples, with each tuple having its first element be a callable and the remaining elements being quantity names, and the base class will check if the user's input is in the right format. 

This design comes from the assumption that native filtering will be done at the Python level, which is not the case when using a database as the underlying data structure. 

Now, because this check is done by the base class, I added this `_allow_string_native_filter` attribute, which, when set to `True`, will allows a list of simple strings also passes the check (see [this line of GCR.py](https://github.com/yymao/generic-catalog-reader/blob/master/GCR.py#L438)).

danielsf: Got it.  Thanks.

PR #144, opened 2018-07-12
U/danielsf/dc2/truth

danielsf:
This PR should add the reader for the DC2 truth catalog


yymao:
Thanks, @danielsf! I am traveling this afternoon; will review later today or tomorrow.

yymao:
@danielsf it seems to me that your implementation load the whole dataset into memory regardless which columns the user requests. Is this the intended behavior? Or should we execute the query within `dc2_truth_native_quantity_getter`?

danielsf:
The only conscious choice was my desire to do the fastest, easiest thing possible.  I can move the query into the getter.

danielsf:
The native_quantity_getter can only take one quantity name at a time, right?

I wonder if that will be less efficient, since we would have to execute one query per requested quantity, as opposed to just selecting all columns in one query.

yymao:
@danielsf In fact, you can just overwrite this method as the following: (I haven't tested it but think it should work)

```python
def _load_quantities(self, quantities, native_quantity_getter):
    native_quantities = tuple(self._translate_quantities(quantities))
    native_data = dict(zip(native_quantities, native_quantity_getter(native_quantities)))
    return {q: self._assemble_quantity(q, native_data) for q in quantities}
```

danielsf:
@yymao 
Your scheme does work.  I have pushed the appropriate changes.  Thanks!

GCRCatalogs/dc2_truth.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _iter_native_dataset(self, native_filters=None):
+        cursor = self._conn.cursor()
+
+        query_where_clause = ' FROM truth'
+
+        if native_filters is not None:
+            query_where_clause += ' WHERE'
+
+            if not isinstance(native_filters, list) and not isinstance(native_filters, tuple):
```
yymao: The user should provide `native_filters` as a list or tuple. So I don't think this is necessary (or should raise error instead?)

GCRCatalogs/dc2_truth.py
```diff
@@ -0,0 +1,79 @@
...
+                native_filters = [native_filters]
+
+            for i_filt, filt in enumerate(native_filters):
```
yymao: No sure if I understand the logic here. In this case, `native_filters` should be a list of strings. So I think we just need
```python
query_where_clause += ' AND '.join(native_filters)
```

GCRCatalogs/dc2_truth.py
```diff
@@ -0,0 +1,79 @@
...
+                if i_filt > 0:
+                    query_where_clause += ' AND'
+                query_where_clause += ' {}'.format(filt)
+
+        # define a method to return a native_quantity_getter
+        # with the API expected by the GCR
+        def dc2_truth_native_quantity_getter(quantities):
+            query = 'SELECT'
+            for i_qty, qty in enumerate(quantities):
```
yymao: Line 60 -- 66, do
```python
query = 'SELECT {} {}'.format(','.join(quantities), query_where_clause)
```

GCRCatalogs/dc2_truth.py
```diff
@@ -0,0 +1,79 @@
...
+from GCR import BaseGenericCatalog
+
+__all__ = ["DC2TruthCatalogReader"]
+
+
+class DC2TruthCatalogReader(BaseGenericCatalog):
+
+    def _subclass_init(self, **kwargs):
+
+        self._allow_string_native_filter = True
```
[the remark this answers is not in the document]
danielsf: Done

GCRCatalogs/dc2_truth.py
```diff
@@ -0,0 +1,79 @@
...
+
+        if not os.path.isfile(kwargs['filename']):
+            raise ValueError("%s is not a valid filename" % kwargs['filename'])
+
+        self._conn = sqlite3.connect(kwargs['filename'])
+
+        # get the descriptions of the columns as provided in the sqlite database
+        cursor = self._conn.cursor()
+        metadata = cursor.execute('SELECT name, description FROM column_descriptions').fetchall()
+        self._column_descriptions = {}
```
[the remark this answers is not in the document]
danielsf: Done

GCRCatalogs/dc2_truth.py
danielsf: The logic here accomplishes the same thing, but I guess it doesn't matter how we do it.

GCRCatalogs/dc2_truth.py
danielsf: Done

yymao:
the catalog name is also a bit confusing. How about `dc2_truth_protoDC2` ?

danielsf:
I thought it would make sense to have the catalog name mirror the protoDC2 extragalactic catalogs.

We should probably include something in the name to indicate this is a truth catalog for run 1.1, not run 1.2.

I guess if we specify the run, we do not need to specify that it is protoDC2, rather than cosmoDC2.

How about `dc2_truth_run1.1`.  I will add a line to the catalog description indicating which version of `protoDC2` the catalog corresponds to.

katrinheitmann:
I think that name captures the important information!

yymao:
I think we should avoid using `proto-dc2` at the beginning because those are all extragalactic catalogs.

danielsf:
Ready for re-review

yymao:
@danielsf all looks good. I'd like to merge this with #148 as both requires a newer version of GCR. So I'll wait until someone reviews #148.

danielsf:
Sounds good.  Push the button whenever you're ready.

PR #147, opened 2018-07-13
make reader works with table format

### LINKED ISSUE #145: Make coadd reader work on table-format hdf5 file
See the discussion in https://github.com/LSSTDESC/DC2_Repo/issues/190

yymao:
OK this PR makes the coadd reader works with `table` hdf5 format to address #145  However, the reader is not yet optimized for `table` format. In particular, it still loads the whole table even when only a few columns are needed. 

To optimize the reader for the `table` format, a more significant change is needed. I'll continue to update this PR in the coming days.  

[Edited: This PR fixes #145]


yymao:
OK, this PR is now ready for review. The major change is that I add another layer of abstraction to deal with the different APIs of the `table` and `fixed` formats.

However, initial tests suggest that `table` format is actually slower :confused:

PR #150, opened 2018-07-15
add sigma8 and n_s to the cosmology attribute

### LINKED ISSUE #44: astropy cosmology does not have sigma8
With gc.cosmology, there is no cosmological parameters associated with the power spectrum amplitude... that seems like an important quantity to have. Maybe we need something beyond astropy?

yymao:
This PR fixes #44 by adding `sigma8` and `n_s` to the astropy cosmology object that can be accessed via
```python
gc = GCRCatalogs.load_catalog('buzzard_test')
sigma8 = gc.cosmology.sigma8
n_s = gc.cosmology.n_s
```
Note that these two attributes have no effect in `astropy.cosmology`. But before we switch to the TJP cosmology class, we will use this to store `sigma8` and `n_s`.

cc @chihway @EiffL @cosmicshear @patricialarsen as they have requested this feature.


yymao:
I have tested this PR, but would appreciate someone can review + test it again.

PR #153, opened 2018-07-17
update truth catalog reader to address #152

### LINKED ISSUE #152: truth catalog reader should not return `object` dtype
The root of the issue @danielsf mentioned in https://github.com/LSSTDESC/DC2_Repo/issues/184#issuecomment-405671855 is that the truth catalog reader return numpy arrays that have the `object` dtype. This happens when any of the magnitudes are asked, because there are objects whose magnitudes are NULL. 

For columns that contain real numbers, the dtype should be np.float. Missing values should be represented as `np.nan`. Either the reader should replace NULL objects with `np.nan`, or there shouldn't be NULL to begin with...

@danielsf is aware of this issue. I open this issue to track the progress.

yymao:
This PR fixes #152 at the reader level. It is still a question why there are nan magnitudes. But the major issue of the original reader is that it uses only one type for all columns. This is now fixed. 

Together with this fix, I added the MD5 check to the truth catalog reader.


GCRCatalogs/dc2_truth.py
```diff
@@ -11,36 +12,46 @@ class DC2TruthCatalogReader(BaseGenericCatalog):
     native_filter_string_only = True
 
     def _subclass_init(self, **kwargs):
+        self._filename = kwargs['filename']
 
-        if not os.path.isfile(kwargs['filename']):
-            raise ValueError("%s is not a valid filename" % kwargs['filename'])
+        if not os.path.isfile(self._filename):
+            raise ValueError('{} is not a valid file'.format(self._filename))
```
danielsf: I'm just curious: is there a reason you chose to change this line?  The new line and the old one look effectively identical to me, unless there is some reason to prefer `format` that I am not aware of...

yymao: Nope, only the consistency with other code in this package. If there's a strong reason to use `%` we can revert it back.

yymao:
@danielsf should we rename `dc2_truth_run1.1.yaml` to `dc2_truth_run1.1p.yaml` (with the 'p')?

yymao:
@danielsf I took the liberty to add 'p' to the catalog name. Let me know if you object.

danielsf:
I guess I don't have a strong opinion about the `p`.

Both PhoSim and ImSim are supposed to be drawing on the same truth information, which is why I left it out. On the other hand: the decision to leave out dust extinction was made because Run 1.1p occurred on a version of PhoSim in which the dust model was broken.  We might as well leave the `p` in then, I guess.

katrinheitmann:
Hmm, I think that might get confusing in the future, no? For Run 1.2 we 
would not have a letter then either since as you say both imsim and 
phosim will have the same truth catalog. I would prefer to drop the p 
and in general have no p or i for the truth catalog. Thanks!

yymao:
Yep, I thought so. That's why I asked, because we do have `p` in the reference catalog names. Do we want to remove `p` from reference catalogs as well? cc @katrinheitmann

PR #156, opened 2018-07-18
cosmoDC2 reader

evevkovacs:
This is a preliminary PR for the cosmoDC2 reader. It will need to be tested again when cosmodc2_v0.1 is uploaded (later tonight). I will update the PR when that happens.


yymao:
Does this supersede #123?

evevkovacs:
No, #123  if for a different catalog (our UniverseMachine baseline catalog). This is for the actual cosmoDC2 that we release to the collaboration.

yymao:
Ah, ok, retitled #123 to make it clear. Thanks.

GCRCatalogs/cosmodc2.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def _generate_native_quantity_list(self):
+        return self._native_quantities
+
+    def _iter_native_dataset(self, native_filters=None):
+        for healpix in self.healpix_pixels:
+
+            if native_filters is not None:
+                fargs = dict(healpix_pixel=healpix)
+                if not all(f[0](*(fargs[k] for k in f[1:])) for f in native_filters if set(f[1:]) == set(fargs)):
```
yymao: I know this PR is not yet ready to be reviewed, but just a heads-up, this line would no longer work with the newer version of GCR that we are using. Change it to:
```python
if not native_filters.check_scalar(fargs):
```

evevkovacs: @yymao Sorry, but do you mean replace line 369 with your substitution?  But fargs isn't defined yet so I would have to move line 370 up, right?

yymao: I meant replacing 371. Or you can combine 369-371 to one line:
```python
if native_filters is not None and not native_filters.check_scalar(dict(healpix_pixel=healpix)):
```

evevkovacs:
cosmo-dc2_v0.1 has been uploaded to nersc. The reader has been tested with DESCQA. See [SkyArea]( https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-07-19_3&catalog=cosmo-dc2_v0.1)
Please update so we will be ready for cosmoDC2-fest. Any minor issues can be fixed tomorrow if need be.

evevkovacs:
@yymao  Just added the yaml file for a test-sized catalog cosmo-dc2_v0.1_test. This covers 1 healpixel. I tried it on the DESCQA sky area test and it worked. You probably want to rename it to cosmoDC2_v0.1_test. That's fine..I have to go catch my flight.

yymao:
@evevkovacs and @dkorytov there are some native quantities that exist in the hdf5 structure but there is no actual data. In particular everything under `emissionLines/` and any `ContinuumLuminosity` that is under `otherLuminosities/`. 

I've removed those native quantities in the reader. Can you fix this in a future version? For now I am going to merge this for cosmoDC2fest.

PR #158, opened 2018-07-20
Issues/142/quantity info dict

### LINKED ISSUE #143: Description of all the fields in DC2 coadd reader also accesible through GCR
@wmwv @yymao @djperrefort  I realised that some of the fields computed at the `merge_tract_cat.py` level are not documented in `SCHEMA.md`, for  instance `mag_i` is a psf magnitude but is documented in `SCHEMA.md` because `mag_i` was already defined in the  extra catalog section. 
Link to this issue is  #142 , so I'm thinking maybe we shouldn't store the schema in a markdown file, but in something more "parsable" so that we can easily publish/access/query it from various places

djperrefort:
Implements the `_get_quantity_info_dict` for the `DC2CoaddCatalog` class.

[edited by @yymao: this PR fixes #143]


djperrefort:
@yymao Moving it to a seperate file would be best. I’ll add that now.

yymao:
thanks @djperrefort. This looks good. I'll test it and approve.

PR #166, opened 2018-07-31
New experimental feature: composite catalog reader

### LINKED ISSUE #65: Update protoDC2 tidal addon catalog to common addon framework
I created a  more generic addon framework to add a random walk component to protoDC2 which loads all the quantities of the  main catalog and can add additional quantities from a secondary hdf5 file. This feature (currently in https://github.com/LSSTDESC/gcr-catalogs/pull/64) is slightly different from the default behavior of  the tidal addon and it  should be updated for homogeneity

yymao:
I am happy to introduce a new experimental feature: composite catalog reader. To test this PR, you will need use the [the latest version of `generic-catalog-reader`](https://github.com/yymao/generic-catalog-reader). See also https://github.com/yymao/generic-catalog-reader/pull/17 for details about the underlying update to GCR.

Since the catalog configs are stored in GCRCalalogs, not in GCR, the addition here adds a simple helper class `CompositeReader` to parse a composite catalog config (see [`um_v0.1_composite_test_shear.yaml`](https://github.com/LSSTDESC/gcr-catalogs/blob/u/yymao/composite-reader/GCRCatalogs/catalog_configs/um_v0.1_composite_test_shear.yaml) as an example) and pass the loaded instance back to `GCR.CompositeCatalog`. 

An illustration of the concept of how composite catalog works can be found [here](https://github.com/yymao/generic-catalog-reader/blob/master/GCR/composite.py#L80). 

I'd like to thank @EiffL @duncandc @dkorytov @patricialarsen for the very useful discussion that took place at the cosmoDC2fest at CMU. 

Comments and suggestions are very welcome. 

Fix #65


yymao:
Since (1) the composite catalog feature is pretty separate from other main feature and, (2) it's more likely to gather feedback when there's something to play with, and (3) GCRCatalogs is still at a rapid development stage, I plan to merge this and install it in the shared environment. 

But if there's objection, I'm happy to hear and then reconsider.

yymao:
I have updated and tested this PR to make sure it works with previous addon catalogs (tidal, knots) from @EiffL. 

[An example notebook can be found here](https://github.com/LSSTDESC/gcr-catalogs/blob/u/yymao/composite-reader/examples/CompositeReader%20example.ipynb) 

To run this notebook, you need this PR and GCR 0.8.2.

PR #170, opened 2018-08-02
cache constant arrays for object catalog reader

### LINKED ISSUE #168: Could the object catalog reader generate and cache just one immutable NaN array?
@yymao Could the object catalog reader be configured to generate just one immutable NaN array and cache that array to be returned anytime some one asked for a column that was not present but should be?

yymao:
This PR enables caching of constant arrays for the object catalog reader. When a user tries to access a column that is supposed to be there (i.e., in the schema) but not actually in that patch, a constant array of NaN is returned. Similarly when the user tries to obtain a column of tract number or patch string. 

Since these constant arrays are, as the name suggests, constant, there's no need to recreate them. This PR make it possible to cache those arrays. 

This change does slightly improve the data access performance.  

Thanks for @wmwv's suggestion in #168. This PR fixes #168.

### DIFF
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -83,6 +83,7 @@ class TableWrapper(object):
         self._columns = None
         self._len = None
         self._cache = None
+        self._constant_arrays = dict()
 
     @property
     def columns(self):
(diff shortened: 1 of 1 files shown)


yymao:
Thanks @wmwv

PR #171, opened 2018-08-03
Issues/169 - Read DC2 object catalog schema if available

wmwv:
Reading the pre-mapped schema saves about 4 seconds out of 11 when reading in the 'dc2_object_run1.1p' catalog.

### DIFF
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -16,6 +16,7 @@ __all__ = ['DC2ObjectCatalog']
 FILE_DIR = os.path.dirname(os.path.abspath(__file__))
 FILE_PATTERN = r'(?:merged|object)_tract_\d+\.hdf5$'
 GROUP_PATTERN = r'(?:coadd|object)_\d+_\d\d$'
+SCHEMA_PATH = 'schema.yaml'
 META_PATH = os.path.join(FILE_DIR, 'catalog_configs/_dc2_object_meta.yaml')
 
 
@@ -157,6 +158,7 @@ class DC2ObjectCatalog(BaseGenericCatalog):
     base_dir          (str): Directory of data files being served, required
     filename_pattern  (str): The optional regex pattern of served data files
     groupname_pattern (str): The optional regex pattern of groups in data files
+    schema_path       (str): The optional location of the schema file
     pixel_scale     (float): scale to convert pixel to arcsec (default: 0.2)
     use_cache        (bool): Whether or not to cache read data in memory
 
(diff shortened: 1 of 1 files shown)


wmwv:
I created a `schema.yaml` by hand and stuck it in the standard DC2 Run 1.1p base_dir on NERSC.  I'll next work on automatically generating them in DC2-production.

GCRCatalogs/dc2_object.py
```diff
@@ -366,6 +377,31 @@ def _generate_datasets(self):
...
+
+        with open(schema_path, 'r') as schema_stream:
+            schema = yaml.load(schema_stream)
+
+        if schema is None:
+            warn_msg = 'Schema file "{}" empty.'
+            warnings.warn(warn_msg.format(schema_path))
+            return set()
+
+        columns = set(schema.keys())
```
yymao: No need to use `.keys()`
```python
columns = set(schema)
```

GCRCatalogs/dc2_object.py
```diff
@@ -366,6 +377,31 @@ def _generate_datasets(self):
...
+            If one or more column names are repeated.
+        """
+
+        with open(schema_path, 'r') as schema_stream:
+            schema = yaml.load(schema_stream)
+
+        if schema is None:
+            warn_msg = 'Schema file "{}" empty.'
+            warnings.warn(warn_msg.format(schema_path))
+            return set()
```
yymao: I think this should either raise an error, or we modify the logic in Lines 191-197 so that we first run `_generate_columns_from_yaml` and check if the result is empty, and if so, fall back to  `_generate_columns`.

GCRCatalogs/dc2_object.py
```diff
@@ -185,11 +188,18 @@ def _subclass_init(self, **kwargs):
             err_msg = 'No catalogs were found in `base_dir` {}'
             raise RuntimeError(err_msg.format(self.base_dir))
 
-        self._columns = self._generate_columns(self._datasets)
+        if self._schema_path is None or not os.path.exists(self._schema_path):
```
yymao: See my comment below on Line 400, and consider changing the logic here to something like:
```python
        self._columns = None
        if self._schema_path and os.path.exists(self._schema_path):
            self._columns = self._generate_columns_from_yaml(self._schema_path)
        if not self._columns:
            warn_msg = 'No schema file found: "{}".\nFalling back to reading all datafiles for column names'
            warnings.warn(warn_msg.format(self._schema_path))
            self._columns = self._generate_columns(self._datasets)
```

PR #173, opened 2018-08-22
bug fix in the UMShearCatalog reader

yymao:
With the new composite reader, `UMShearCatalog` can just use the same `_subclass_init` as in `UMGalaxyCatalog`.

This should fix a bug that @evevkovacs encountered.

### DIFF
--- a/GCRCatalogs/cosmodc2.py
+++ b/GCRCatalogs/cosmodc2.py
@@ -466,8 +466,3 @@ class UMShearCatalog(UMGalaxyCatalog):
             'shear_2_phosim':   (np.negative, 'shear_2'),
         }
         return quantity_modifiers
-
-    def _subclass_init(self, catalog_root_dir, catalog_path_template, cosmology, healpix_pixels=None, zlo=None, zhi=None, check_file_metadata=False, **kwargs):
-        super(UMShearCatalog, self)._subclass_init(catalog_root_dir, catalog_path_template, cosmology, healpix_pixels=None, zlo=None, zhi=None, check_file_metadata=False, **kwargs)
-        self.composite_compatible = kwargs.get('composite_compatible', [])
-        self.composite_matched_format = kwargs.get('composite_matched_format', [])


evevkovacs:
Yes, this is now working:
If I load the composite catalog, only healpix 564 is read, as expected...

 In [3]: import GCRCatalogs

In [4]: gc = GCRCatalogs.load_catalog('baseDC2_v0.4.5_shear_addon_test')
Reading file /global/projecta/projectdirs/lsst/groups/CS/cosmoDC2/baseDC2_min_9.8_centrals_v0.4.5/baseDC2_z_0_1_cutout_564.hdf5
Reading file /global/projecta/projectdirs/lsst/groups/CS/cosmoDC2/baseDC2_min_9.8_centrals_v0.4.5/baseDC2_z_1_2_cutout_564.hdf5
Reading file /global/projecta/projectdirs/lsst/groups/CS/cosmoDC2/baseDC2_min_9.8_centrals_v0.4.5/baseDC2_z_2_3_cutout_564.hdf5
Reading file /global/projecta/projectdirs/lsst/groups/CS/cosmoDC2/baseDC2_shear_addon_v0.4.5/baseDC2_z_0_1_cutout_564.hdf5
Reading file /global/projecta/projectdirs/lsst/groups/CS/cosmoDC2/baseDC2_shear_addon_v0.4.5/baseDC2_z_1_2_cutout_564.hdf5
Reading file /global/projecta/projectdirs/lsst/groups/CS/cosmoDC2/baseDC2_shear_addon_v0.4.5/baseDC2_z_2_3_cutout_564.hdf5
/global/common/software/lsst/common/miniconda/py3-4.3.21-env/lib/python3.6/site-packages/GCR/composite.py:105: UserWarning: CompositeCatalog is still an experimental feature. Use with care!
  warnings.warn('CompositeCatalog is still an experimental feature. Use with care!')

evevkovacs:
OK, here is the next bug (I tried running a readiness test on the composite catalog and got a crash. I reproduced the problem in an interactive python session: 
When I try the following
all_quantities = sorted(gc.list_all_quantities(True))
I get
TypeError: '<' not supported between instances of 'tuple' and 'str'
So, gc.list_all_quantities() works and gives a list, but gc.list_all_quantities(True) gives a list with some elements as a tuple. I think I understand. 'GCR' (schema) quantities are strings and native quantities are a tuple with the catalog name. So if I want the readiness test to work on the shear add-on, I need to handle these tuples. I changed line 141; will test and submit a PR if all is well.

yymao:
@evevkovacs ah, so that's a bug in our test(s) --- GCR doesn't guarantee that the quantity names must be string. So we should have done:
```python
sorted(map(str, gc.list_all_quantities(True)))
```

evevkovacs:
@yymao @patricialarsen Patricia is just checking: The galaxy-id problem may be a bug in one of her files.

yymao:
@evevkovacs thanks for the update. If there's further update on this, please open a new issue because this issue has been closed, and the new problem doesn't seem to be related to this issue.

Issue #175, opened 2018-08-23
path in the focal plane config files no longer exists


yymao:
The config files for the focal planes ([`focal_plane_0_test`](https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/focal_plane_0_test.yaml) and [`focal_plane_16_test`](https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/focal_plane_16_test.yaml)) point to the following path
```
/global/cscratch1/sd/descpho/Pipeline-tasks/DC2-R1-2p-WFD-r/000000/work
```
which no longer exist on NERSC. @fjaviersanchez can you fix this?

PR #176, opened 2018-08-24
Updated path to e-images

fjaviersanchez:
This PR tries to solve #175. I also added compatibility for files with no instance catalogs available in the parent directory.

### DIFF
--- a/GCRCatalogs/focal_plane.py
+++ b/GCRCatalogs/focal_plane.py
@@ -84,9 +84,13 @@ class FocalPlaneCatalog(BaseGenericCatalog):
         self._filelist = glob.glob(os.path.join(catalog_root_dir, 'lsst_e*.fits*'))
         self.rebinning = rebinning
         parent_path = os.path.dirname(catalog_root_dir)
-        instcat_path = glob.glob(os.path.join(parent_path, 'instCat/phosim*.txt'))[0]
-        self.phosim_pars = pd.read_table(instcat_path, index_col=0, header=None, sep=' ').T
-        self.visit = self.phosim_pars['obshistid'].values[0]
+        try:
+            instcat_path = glob.glob(os.path.join(parent_path, 'instCat/phosim*.txt'))[0]
+            self.phosim_pars = pd.read_table(instcat_path, index_col=0, header=None, sep=' ').T
+            self.visit = self.phosim_pars['obshistid'].values[0]
+        except:
+            self.phosim_pars = None
+            self.visit = os.path.split(self._filelist[0])[1].split('_')[2]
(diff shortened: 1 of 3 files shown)


GCRCatalogs/focal_plane.py
```diff
@@ -84,9 +84,13 @@ def _subclass_init(self, catalog_root_dir, rebinning=16, **kwargs):
...
         self.rebinning = rebinning
         parent_path = os.path.dirname(catalog_root_dir)
-        instcat_path = glob.glob(os.path.join(parent_path, 'instCat/phosim*.txt'))[0]
-        self.phosim_pars = pd.read_table(instcat_path, index_col=0, header=None, sep=' ').T
-        self.visit = self.phosim_pars['obshistid'].values[0]
+        try:
+            instcat_path = glob.glob(os.path.join(parent_path, 'instCat/phosim*.txt'))[0]
+            self.phosim_pars = pd.read_table(instcat_path, index_col=0, header=None, sep=' ').T
+            self.visit = self.phosim_pars['obshistid'].values[0]
+        except:
```
yymao: Please add the specific error this try-except is trying to catch.

GCRCatalogs/focal_plane.py
```diff
@@ -84,9 +84,14 @@ def _subclass_init(self, catalog_root_dir, rebinning=16, **kwargs):
...
-        self.phosim_pars = pd.read_table(instcat_path, index_col=0, header=None, sep=' ').T
-        self.visit = self.phosim_pars['obshistid'].values[0]
+        try:
+            instcat_path = glob.glob(os.path.join(parent_path, 'instCat/phosim*.txt'))[0]
+            self.phosim_pars = pd.read_table(instcat_path, index_col=0, header=None, sep=' ').T
+            self.visit = self.phosim_pars['obshistid'].values[0]
+        except IndexError:
+            print('No instance catalog found in the expected path')
+            self.phosim_pars = None
+            self.visit = os.path.split(self._filelist[0])[1].split('_')[2]
```
yymao: So here you are catching the `IndexError` coming only from Line 88 (not Line 90), right? If so, move Lines 89-90 to the `else` clause:
```python
        try:
            instcat_path = glob.glob(os.path.join(parent_path, 'instCat/phosim*.txt'))[0]
        except IndexError:
            print('No instance catalog found in the expected path')
            self.phosim_pars = None
            self.visit = os.path.split(self._filelist[0])[1].split('_')[2]
        else:
            self.phosim_pars = pd.read_table(instcat_path, index_col=0, header=None, sep=' ').T
            self.visit = self.phosim_pars['obshistid'].values[0]
```

PR #183, opened 2018-09-02
add support to legacy instance catalogs

yymao:
To allow the validation test proposed in  https://github.com/LSSTDESC/descqa/issues/152 work with legacy instance catalogs (AGNs and galaxies are printed in the same file), we need to update the instance catalog reader so that it works with both cases. This PR adds the said support.

### DIFF
--- a/GCRCatalogs/instance_catalog.py
+++ b/GCRCatalogs/instance_catalog.py
@@ -217,6 +217,13 @@ class InstanceCatalog(BaseGenericCatalog):
                 raise ValueError('cannot determine whether this is a legacy instance catalog!')
             for t in self._legacy_gal_types:
                 self._object_files[t] = self._object_files['gal']
+            del self._object_files['gal']
+
+        try:
+            self.visit = int(self.header.get('obshistid'))
+        except (TypeError, ValueError):
+            warnings.warn('Cannot parse visit id {}'.format(self.header.get('obshistid')))
+            self.visit = None
 
         shape_quantities = ('gal/a_bulge',
                             'gal/b_bulge',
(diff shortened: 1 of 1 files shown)


GCRCatalogs/instance_catalog.py
```diff
@@ -246,47 +265,82 @@ def _generate_native_quantity_list(self):
...
+        )
+
+    def _load_legacy_gal_catalog(self, obj_type):
+        if '_legacy_gal_line_index' not in self._data:
+            path = self._object_files[obj_type]
+            this_open = gzip.open if path.endswith('.gz') else open
+            with this_open(path, 'rb') as f:
+                for index, line in enumerate(f):
+                    if b' agnSED/' in line:
+                        self._data['_legacy_gal_line_index'] = index
```
danielsf: Just checking: are we absolutely certain that the AGN are all piled up at the end of the catalog?  I am specifically worried about the sprinkler breaking this assumption.

yymao: This is at least true for all the Run 1.2 instance catalogs, but I am not sure if this assumption is *always* true. However, given that (1) this assumption does help a lot in term of load performance though, (2) if there are galaxy entries appearing after an AGN line, the reader will throw an error rather than proceed silently, and (3) the instance catalog will be in the new format for Run 2.0 (is that right?), I think it is ok to leave the legacy support as it. What do you think, @danielsf?

danielsf: If it's true for Run 1.2, then I agree it is safe to use.  We should not be going back to the legacy format any time soon (or, really, at all).

danielsf:
Merge away

PR #188, opened 2018-09-11
New addon catalog for Cosmo dc2 v1.0 (image subset)

EiffL:
This PR adds a new addon catalog for the cosmoDC2 v1.0 release. Due to time constraints, I'm submitting only the subset of healpixels that correspond to the image simulation footprint.


yymao:
@EiffL is there any specific reason for `CosmoDC2AddonCatalog` to not directly inherit from `CosmoDC2ParentClass`? It seems to me that this addon class is very similar to `BaseDC2ShearCatalog` and can just be a subclass of `CosmoDC2ParentClass` and put under `cosmodc2.py`.

EiffL:
Yes, I didn't want to bother with storing a cosmology and other
metadata.... Which is why I simplified the class...

yymao:
@EiffL we can simply skip setting cosmology if `cosmology` is not supplied in the config file. I don't think it's a good idea to create new classes from scratch when there is existing class that does exactly the same thing.

yymao:
@EiffL I've updated the add-on class to be a subclass of `CosmoDC2ParentClass` ([see here](https://github.com/LSSTDESC/gcr-catalogs/blob/d375c1c3b1a7eb17145fd4d9bfc94cb204916e3e/GCRCatalogs/cosmodc2.py#L500)), the new subclass requires only 3 lines of code. 

Please let me know if this looks good. I've done some basic tests, but it would be great of you have some code to test it.

EiffL:
Thanks Yao, I'll be able to test it within 30 min

EiffL:
Yes this works :-D thanks so much Yao !

PR #189, opened 2018-09-11
implement file size check

### LINKED ISSUE #184: enforce md5sum check for cosmoDC2
To ensure reproducibility, we will enforce md5sum check for the version of cosmoDC2 that will go into image simulation, similar to what we did for protoDC2.

This task includes adding md5 sum check in the cosmoDC2 reader and add the md5 sum values in the corresponding config file.

yymao:
This PR implements the file size check. The file sizes can be specified in `GCRCatalogs/catalog_configs/_cosmoDC2_check.yaml`. 

edited: If we have time we would add md5 sums to this PR, but apparently we don't have much time...


yymao:
We should add md5 sums to `GCRCatalogs/catalog_configs/_cosmoDC2_check.yaml` before we merge this PR. @evevkovacs are you currently generating the md5 sums?

evevkovacs:
Ah, I see. I am using subprocess to compute the md5sums for the yaml file. Answers will be the same I hope!

yymao:
It should be!

To update your review, go to https://github.com/LSSTDESC/gcr-catalogs/pull/189/files and you'll see a button on the upper right corner.

PR #191, opened 2018-09-11
Use nside info to set default sky area when not available in file metadata

yymao:
This PR fixes the implementation of default sky area (when not available in file metadata) to use nside info instead of assuming a constant.


yymao:
With this fix, the `sky_area` attribute will have the correct value for the add-on catalogs that do not store sky area in the file metadata (e.g. shear, knots).

PR #193, opened 2018-09-11
Using the convergence value in the catalog files instead of calculating it

### LINKED ISSUE #159: Use native convergence quantity in CosmoDC2 Reader
CosmoDC2 will have have correct convergence as a native quantity, so there's no need for a derived quantity.

dkorytov:
I double checked that the values in the files agree with the calculated method. 

fixes #159


yymao:
@dkorytov, thanks!  Will this change be backward-compatible with `cosmoDC2_v0.1` and `cosmoDC2_v0.4`?

yymao:
The code works for `cosmoDC2_v0.1` and `cosmoDC2_v0.4` but note that `convergence` is always zero for these two catalogs.

yymao:
thanks for confirming!

PR #195, opened 2018-09-12
add back the missing healpixel 10444

yymao:
@evevkovacs @patricialarsen @EiffL mentioned that there's a missing healpixel 10444. This PR adds it back. 

@jbkalmbach @rbiswas4 @danielsf does this change matter for the AGN database? And if so, should we not merge this until later?


yymao:
Agree, I'll hold off on merging this until I hear go from @jbkalmbach @rbiswas4 @danielsf.

danielsf:
AGN and supernovae depend on sqlite databases of parameters that were generated from the old `*_images` catalog.  If we were to add the errant healpixel in now, those galaxies would have no AGN or supernovae assigned to them, which is probably fine.

More worrisome is that Antonio is generating InstanceCatalogs *right now*, and those InstanceCatalogs are being drawn from an `*_image` catalog that does not include this healpixel, which means the image simulators won't put any galaxies in that healpixel.

I wonder if we should just leave the healpixel out (i.e. never merge this).  We are only losing 0.8 square degrees.  I don't want to enter a state where some images include the objects in this healpixel and some images do not.  I'm not sure how coadd generation will handle that.

Thoughts?

yymao:
(cc @wmwv @humnaawan as they may have thoughts about this)

evevkovacs:
Just so everyone knows, the missing healpixel (10444) covers a small area right at the SW corner of the field (most of the healpixel's area is outside of the field). Regarding co-adds, won't there be cases in the real data where visits are missing? There must be a provision in DM to handle that.

evevkovacs:
@villarrealas @yymao @danielsf Not clear to me that Antonio is actually running yet. He just sent me a Slack message saying he was dealing with other issues, so there may be time to include this change.
To quote "Well, we're dealing with another issue right now, so there's some time for last minute resolutions now."

yymao:
@jchiang87 just to clarify --- do you mean that it's fine to never merge this?

jchiang87:
> just to clarify --- do you mean that it's fine to never merge this?

From the perspective of doing DM processing and connecting the associated outputs back to the input truth, I think that's correct.   There may be other use cases where merging would be desirable.

danielsf:
We actually still aren't running, and I need to re-generate the AGN database.

If we can merge this in the next ~ 10 minutes, I think the missing sources will get incorporated (with the caveats that there will be no supernovae or galaxies with knots in the errant healpixel).

yymao:
@danielsf This is ready to be merged. I can press the button now but as long as no one objects. I know @EiffL has generated knots for this missing healpixel.

danielsf:
Okay.  Push the button now

EiffL:
Yep, the files are there but haven't tested  it, I can do that now (testing) it will take 2 mins

EiffL:
ok everything looks good to me!

PR #197, opened 2018-09-12
List of md5sums for cosmoDC2_v1.0.0

evevkovacs:
File contains list of md5sums for 672 files. I changed the default name to include _md5 to differentiate it from the check file containing sizes. I suggest we keep the files separate so that the check can be done either with sizes or md5 sums, since the latter will take a very long time. We need to make some minor code changes to allow this option. I haven't attempted to make those yet, but am happy to do so.


evevkovacs:
Right, so if we wanted to keep both checking options, we would have to change the code. It would be easy enough to define CHECK_FILE_PATH_SIZE and CHECK_FILE_PATH_MD5. 
The image sims are using the file check with the sizes, so we should keep that around and rename it with _size or similar. I think it is worth keeping both checking options for now because it may turn out that the md5 sum check takes a long time to run.

yymao:
@evevkovacs, there is no need to change the code at all. The file `_cosmoDC2_check.yaml` should contain both size and md5 info. The content would just look like :

```yaml
'1.0.0' :
  size :
    z_0_1.step_all.healpix_10064.hdf5 : 4983541383
    z_0_1.step_all.healpix_10065.hdf5 : 4943967537
    ...
  md5 :
    z_0_1.step_all.healpix_10064.hdf5 : 9428ed4b3cb8d30747496ec764f49174
    z_0_1.step_all.healpix_10065.hdf5 : b01b5b45c284bdc74afc5f711b958bbe
    ...
```

On the other hand, whether or not the reader actually executes these checks is controlled by the `check_md5` and `check_size` options in the catalog config files, not by this `_cosmoDC2_check.yaml` file.

evevkovacs:
Ah, ok, for some reason, I thought we wanted them in separate files. I'll merge them and resubmit.
Thanks.

evevkovacs:
OK, files are merged

evevkovacs:
Oh sorry, I am asleep at the switch today. Done.

PR #207, opened 2018-10-05
Truth catalogs for Run 1.2

### LINKED ISSUE #203: Update reader to serve Run 1.2 truth catalogs
We need to update the truth catalog reader for the Run 1.2 truth catalogs.

For static objects this is straightforward. We merely need to change the database path, and no additional work is needed. 

For transients and variables, we want to provide both (1) a summary table that has object IDs, positions, and some summary statistics, and (2) a collection of light curves. The summary table can be served in a similar fashion as how the static objects are served (with different columns, of course, but the data interface can be similar); however, we may need to compute some summary statistics in advance. We need a good way to serve the collection of light curves. 

This is being developed in the [`u/yymao/truth-run1.2`](https://github.com/LSSTDESC/gcr-catalogs/tree/u/yymao/truth-run1.2) branch. See also [this file](https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/u/danielsf/truth/catalogs/workspace/truth/workflow.txt) for the underlying database structure. 

(cc @wmwv @danielsf)

yymao:
This PR adds the truth catalogs for Run 1.2 (cf. https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/54). This PR fixes #203.

The truth catalog for static objects works just like Run 1.1. It is temporarily called `dc2_truth_run1.2_static_test`. 

We also added another two catalogs for the variable objects: `dc2_truth_run1.2_variable_summary_test` and `dc2_truth_run1.2_variable_lightcurve_test`. 

The "summary" catalog is just a simple table, so the reader works just like the static object catalog. The `filters` and `native_filters` options in `get_quantities` work as you would normal expected. The available columns are:

```python
>>> cat = GCRCatalogs.load_catalog('dc2_truth_run1.2_variable_summary_test')
>>> cat.list_all_native_quantities()
['dec', 'galaxy_id', 'agn', 'uniqueId', 'sn', 'ra', 'sprinkled']
```

The "lightcurve" catalog is a bit different. The data returned by the reader are the light curves themselves and the `filters` would work on the columns of the light curves:

```python
>>> lc = GCRCatalogs.load_catalog('dc2_truth_run1.2_variable_lightcurve_test')
>>> lc.list_all_native_quantities()
['filter', 'mag', 'obshistid', 'uniqueId', 'mjd']
```

However, the different thing about the "lightcurve" reader is that the `native_filters` would work on the columns of the "summary" catalog. And if `return_iterator` is turned on, it will iterate over objects. 

It's probably easier to illustrate this by an example: 
```python
n_plotted = 0
for q in lc.get_quantities(['mjd', 'mag'], filters=['filter == 3'], native_filters=['sn == 1', 'ra > 40'], return_iterator=True):
    if len(q['mjd']) >= 20:
        plt.plot(q['mjd'], q['mag'], '.-')
        n_plotted += 1
    if n_plotted >= 3:
        break
plt.ylim(29, 24)
plt.xlabel('MJD')
plt.ylabel('$i$ [mag]')
```
This will generate a plot like:
image

Comments about this API are very welcome (cc @wmwv). All examples above would work when using this PR, so people can play with it.


yymao:
Thanks, @danielsf! I think I'll keep this PR open for a few more days to collect feedback about the data access interface. 

In the meantime, should we move the database to a more permanent location? If you do, let me know and I'll update the catalog config files and remove `_test` from the config filenames.

yymao:
@wmwv I think we should merge this for the upcoming sprint week, any objection? 

Before we merge, however, should we move the truth catalog to a more permanent location, @danielsf? Also are the truth catalog we point to here the final version (i.e., complete) or just a smaller testing version? If the former I'll also change the yaml filename. 

(cc @reneehlozek as this is one of the hackuDC2 preparation task)

danielsf:
@yymao I have moved all of the databases underlying the Run 1.2 truth catalog to
```
/global/projecta/projectdirs/lsst/groups/SSim/DC2/Run1.2
```
You can redirect the config file.

These are the final copies (they are not a smaller test area or anything like that)

yymao:
Update paths and remove `_test` from config names. @danielsf can you re-approve? Thanks!

yymao:
@danielsf sorry to ping you again on this. I'd like to get this merged and prepare a simple tutorial based on my example above.

yymao:
Thanks @danielsf

PR #209, opened 2018-10-09
Issues/208 - Add dtype, default to schema.yaml

wmwv:
DC2 Reader
Use dtype and default values from schema.yaml.  If not present default to float, nan.

tract, patch handled on a per-ObjectTableWrapper basis by creating schema columns with default entries equal to the tract, patch for that particular ObjectTableWrapper.


wmwv:
@yymao Can you re-approve.  I had to submit one more change to satisfying the linter so that I could merge.

PR #211, opened 2018-10-11
update/review of issues/208 (#209)

yymao:
@wmwv I felt it's a bit difficult to well explain the change I would request in my review of #209. Since I'm out of town and won't be able to chat with you in person, I use this PR to show you my recommended changes. I hope you don't mind. 

The main change is the behavior of `TableWrapper._generate_constant_array` and `TableWrapper._generate_constant_array`. 

I think we should use the former to find the correspondence between column name (`key`) and the dtype and default value for that column, and we use the latter to construct and cache a constant array according to the dtype and default value. The latter does not care about the column name. 

Note that when using the dictionary `_constant_arrays` to cache the constant arrays, I also need a "key" --- this key has nothing to do with the column name. 

In other words, [this line of your PR](https://github.com/LSSTDESC/gcr-catalogs/compare/issues/208...update-issues/208?expand=1#diff-adb727eb66fa5b3cfd2ec51ff506169fL154) should be outside of the `except` block. I further edit the code so that `_generate_constant_array` takes dtype and default value directly to avoid confusion. 

Also, we need to ensure the constant array constructed actually conform to the dtype specified ([which is now implemented here](https://github.com/LSSTDESC/gcr-catalogs/compare/issues/208...update-issues/208?expand=1#diff-adb727eb66fa5b3cfd2ec51ff506169fR150)).

I also make `schema` an optional argument, and this allows us simply the logic in the main reader.

Unfortunately I haven't got a chance to actually test my update. I'll have more time on Friday if you don't get to it first.

### DIFF
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -75,16 +75,9 @@ class TableWrapper(object):
     And a schema to specify dtypes and default values for missing columns.
     """
 
-    def __init__(self, file_handle, key, schema):
+    def __init__(self, file_handle, key, schema=None):
         if not file_handle.is_open:
             raise ValueError('file handle has been closed!')
-
-        # The schema will be slightly different for each patch
-        # because we're implementing the tract, patch
-        # as default values in the schema
-        # This is perhaps wrong, but in any event we don't
-        # want to modify the originally passed-in schema.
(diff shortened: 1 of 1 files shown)


yymao:
Thanks, @wmwv

PR #213, opened 2018-10-16
change modifiers for cModel parameters

### LINKED ISSUE #201: Where does +27.0 come in _modelfit_Cmodel_flux?
@yymao @djperrefort Where does the `27.0` come from in the definition of the magnitude for `{}_modelfit_Cmodel_flux`?

https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/dc2_object.py#L270

```
  modifiers['mag_{}_cModel'.format(band)] = (
                lambda x: -2.5 * np.log10(x) + 27.0,
                '{}_modelfit_CModel_flux'.format(band),
            )
```

plaszczy:
update modfiers compatible with https://github.com/LSSTDESC/DC2-production/issues/272

[edited by @yymao: fix #201]

### DIFF
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -296,22 +296,9 @@ class DC2ObjectCatalog(BaseGenericCatalog):
                 modifiers['I{}_{}'.format(ax, band)] = '{}_base_SdssShape_{}'.format(band, ax)
                 modifiers['I{}PSF_{}'.format(ax, band)] = '{}_base_SdssShape_psf_{}'.format(band, ax)
 
-            modifiers['mag_{}_cModel'.format(band)] = (
-                lambda x: -2.5 * np.log10(x) + 27.0,
-                '{}_modelfit_CModel_flux'.format(band),
-            )
-
-            modifiers['magerr_{}_cModel'.format(band)] = (
-                lambda flux, err: (2.5 * err) / (flux * np.log(10)),
-                '{}_modelfit_CModel_flux'.format(band),
-                '{}_modelfit_CModel_fluxSigma'.format(band),
-            )
-
-            modifiers['snr_{}_cModel'.format(band)] = (
-                np.divide,
-                '{}_modelfit_CModel_flux'.format(band),
-                '{}_modelfit_CModel_fluxSigma'.format(band),
-            )
+            modifiers['mag_{}_cModel'.format(band)] = '{}_modelfit_mag'.format(band)
+            modifiers['magerr_{}_cModel'.format(band)] = '{}_modelfit_mag_err'.format(band)
+            modifiers['snr_{}_cModel'.format(band)] = '{}_modelfit_SNR'.format(band)
 
             modifiers['psf_fwhm_{}'.format(band)] = (
                 lambda xx, yy, xy: pixel_scale * 2.355 * (xx * yy - xy * xy) ** 0.25,
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -240,7 +240,7 @@ class DC2ObjectCatalog(BaseGenericCatalog):
         # A slightly crude way of checking for version of schema to have modelfit mag
         # A future improvement will be to explicitly store version information in the datasets
         # and just rely on that versioning.
-        modelfit_mag_re = re.compile('mag_.+_cModel')
+        modelfit_mag_re = re.compile('.+_modelfit_mag')
         has_modelfit_mag = [col for col in self._columns if modelfit_mag_re.match(col)]
         if has_modelfit_mag:
             self._schema_version = '1.2'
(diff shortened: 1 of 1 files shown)


plaszczy:
sorry I forgot to update master. I merged it now. could you check?

wmwv:
I think we should either add a check for the existence of the columns or make a new reader for Run 1.2p.

yymao:
I would vote for adding a check.  But if more people want to make a reader, we should make sure just inherit from current reader instead of making a new one.

wmwv:
Or we could add version field to the `DC2ObjectCatalog` to trigger some small if/else for this.

If there begins to be too much divergence, then we can write separate objects with different `_generate_modifiers` functions.

What's your opinion, @yymao ?

yymao:
Adding version to metadata or catalog config and let the reader changes it's behavior accordingly is certainly the best option. That's what we do for extragalactic catalogs. Checking column is just a quick fix. In either case I suggest adding conditions in current reader instead of making a separate one.

yymao:
@plaszczy let me know if you need any help on this!

wmwv:
The initialization of the DC2Object now checks for the column existence and sets a version accordingly.  This version setting should be improve and standardized as we work that in to providing the version in the catalog files in the generating scripts in `DC2-production`.

I have tested this for both Run 1.1 and Run 1.2 and it works.

GCRCatalogs/dc2_object.py
```diff
@@ -229,14 +236,25 @@ def _subclass_init(self, **kwargs):
             self._columns = self._generate_columns(self._datasets)
 
         bands = [col[0] for col in self._columns if len(col) == 5 and col.endswith('_mag')]
-        self._quantity_modifiers = self._generate_modifiers(self.pixel_scale, bands)
+
+        # A slightly crude way of checking for version of schema to have modelfit mag
+        # A future improvement will be to explicitly store version information in the datasets
+        # and just rely on that versioning.
+        modelfit_mag_re = re.compile('.+_modelfit_mag')
+        has_modelfit_mag = [col for col in self._columns if modelfit_mag_re.match(col)]
```
yymao: Line 243-244: regular expression seems overkill here; `endswith` is sufficient and more readable. Also, we can use `any` to allow short circuit (and also make code more direct in terms of readability)

```suggestion
        has_modelfit_mag = any(col.endswith('_modelfit_mag') for col in self._columns)
```

GCRCatalogs/dc2_object.py
```diff
@@ -229,14 +236,25 @@ def _subclass_init(self, **kwargs):
...
+
+        self._quantity_modifiers = self._generate_modifiers(self.pixel_scale, bands, version=self._schema_version)
         self._quantity_info_dict = self._generate_info_dict(META_PATH, bands)
 
     def __del__(self):
         self.close_all_file_handles()
 
     @staticmethod
-    def _generate_modifiers(pixel_scale=0.2, bands='ugrizy'):
+    def _generate_modifiers(pixel_scale=0.2, bands='ugrizy', version=''):
```
yymao: set default to `None`
```suggestion
    def _generate_modifiers(pixel_scale=0.2, bands='ugrizy', version=None):
```

GCRCatalogs/dc2_object.py
```diff
@@ -296,22 +314,25 @@ def _generate_modifiers(pixel_scale=0.2, bands='ugrizy'):
...
+                )
+                modifiers['snr_{}_cModel'.format(band)] = (
+                    np.divide,
+                    '{}_modelfit_CModel_flux'.format(band),
+                    '{}_modelfit_CModel_fluxSigma'.format(band),
+                )
+            else:
+                modifiers['mag_{}_cModel'.format(band)] = '{}_modelfit_mag'.format(band)
+                modifiers['magerr_{}_cModel'.format(band)] = '{}_modelfit_mag_err'.format(band)
+                modifiers['snr_{}_cModel'.format(band)] = '{}_modelfit_SNR'.format(band)
```
yymao: Move _current_ case outside conditional statement, so that in the future when removing deprecated code, it'll be a cleaner removal.

```python
            modifiers['mag_{}_cModel'.format(band)] = '{}_modelfit_mag'.format(band)
            modifiers['magerr_{}_cModel'.format(band)] = '{}_modelfit_mag_err'.format(band)
            modifiers['snr_{}_cModel'.format(band)] = '{}_modelfit_SNR'.format(band)

            if version == '1.1':
                modifiers['mag_{}_cModel'.format(band)] = (
                    lambda x: -2.5 * np.log10(x) + 27.0,
                    '{}_modelfit_CModel_flux'.format(band),
                )
                modifiers['magerr_{}_cModel'.format(band)] = (
                    lambda flux, err: (2.5 * err) / (flux * np.log(10)),
                    '{}_modelfit_CModel_flux'.format(band),
                    '{}_modelfit_CModel_fluxSigma'.format(band),
                )
                modifiers['snr_{}_cModel'.format(band)] = (
                    np.divide,
                    '{}_modelfit_CModel_flux'.format(band),
                    '{}_modelfit_CModel_fluxSigma'.format(band),
                )
```

wmwv:
Thanks, @yymao .   Branch update with your requested changes.

I have a set of Run 1.2i files that this really works on NERSC.  I need to do some further checking to make sure there's nothing crazy, then I'm going to ping #desc-dc2-data-access to ask for people to start checking out the Run 1.2i data.

PR #214, opened 2018-10-16
update magnitude modifiers in cosmoDC2 reader

### LINKED ISSUE #212: Add `mag_x` aliases in cosmoDC2 reader
We should add `mag_x` as aliases for `mag_x_lsst`, as we describe in the schema. This addition is also required for make the cosmoDC2 reader compatible with the existing tutorial notebook.

Basically, we need to copy [these three lines (274-276) from the protoDC2 reader](https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/alphaq.py#L274) to the cosmoDC2 reader (around Line 424).

(cc @reneehlozek as this is one of the hackuDC2 preparation task)

yymao:
Update the magnitude modifiers in cosmoDC2 reader to be consistent with protoDC2. 
Fixes #212.

### DIFF
--- a/GCRCatalogs/cosmodc2.py
+++ b/GCRCatalogs/cosmodc2.py
@@ -405,22 +405,23 @@ class CosmoDC2GalaxyCatalog(CosmoDC2ParentClass):
         # add magnitudes
         for band in 'ugrizyY':
             if band != 'y' and band != 'Y':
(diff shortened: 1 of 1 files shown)


yymao:
Actually there was a bug. Direct copy and paste from protoDC2 reader would not work since in cosmoDC2 we do not edit `self._quantity_modifiers` directly. 

I've tested this version to make sure it works. @evevkovacs can you re-approve? Thanks.

PR #224, opened 2018-11-16
Specify the schema_path for dc2_object_run1.2i{,_all_columns}

wmwv:
The full schema for the full file is `schema.yaml`
The trimed schema for the trim file is `trim_schema.yaml`.

The `trim_object*` files in the `Run1.2i/object_catalog` directory have also been updated.  There was an error in their original generation.

### DIFF
--- a/GCRCatalogs/catalog_configs/dc2_object_run1.2i.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_object_run1.2i.yaml
@@ -1,5 +1,6 @@
 subclass_name: dc2_object.DC2ObjectCatalog
 base_dir: /global/projecta/projectdirs/lsst/global/in2p3/Run1.2i/object_catalog
+schema_path: trim_schema.yaml
 filename_pattern: 'trim_object_tract_\d+\.hdf5$'
 description: DC2 Run 1.2i Object Catalog
 creators: ['Michael Wood-Vasey']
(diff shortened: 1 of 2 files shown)


GCRCatalogs/dc2_object.py
```diff
@@ -212,7 +213,13 @@ def _subclass_init(self, **kwargs):
         self.base_dir = kwargs['base_dir']
         self._filename_re = re.compile(kwargs.get('filename_pattern', FILE_PATTERN))
         self._groupname_re = re.compile(kwargs.get('groupname_pattern', GROUP_PATTERN))
-        self._schema_path = kwargs.get('schema_path', os.path.join(self.base_dir, SCHEMA_PATH))
+
+        _schema_filename = kwargs.get('schema_filename', SCHEMA_FILENAME)
+        if os.path.isabs(_schema_filename):
+            self._schema_path = _schema_filename
+        else:
+            self._schema_path = os.path.join(self.base_dir, _schema_filename)
```
yymao: There is in fact no need to check if `_schema_filename` is absolute path, because the behavior of `os.path.join` is that everything before an absolute path will be ignored ([ref](https://docs.python.org/3.6/library/os.path.html#os.path.join)). 

Hence, you can safely replace Lines 218-221 with just
```python
self._schema_path = os.path.join(self.base_dir, _schema_filename)
```

wmwv: Woohoo!

PR #227, opened 2018-11-19
minor bug fix for empty timestep +config files for baseDC2

evevkovacs:
This PR fixes a bug in the base class for the reader. The Nside=32 healpixels are small enough that at low z, sometimes the timestep is empty. The code needs to check if the group is empty before proceeding. Config files for the basedDC2 v1.1 catalogs have also been added.


GCRCatalogs/cosmodc2.py
```diff
@@ -286,7 +286,8 @@ def _iter_native_dataset(self, native_filters=None):
             with h5py.File(file_path, 'r') as fh:
                 for group in self._get_group_names(fh):
                     # pylint: disable=E1101,W0640
-                    yield lambda native_quantity: fh['{}/{}'.format(group, native_quantity)].value
+                    if len(fh[group]) > 0:
```
yymao: minor code style change:
```suggestion
                    if len(fh[group]):
```

PR #228, opened 2018-11-19
object catalog reader now provides fallback schema

### LINKED ISSUE #225: outdated schema and mismatched native quantities in object catalog 1.2i
The schema file for object catalog 1.2i is still outdated (i.e., native quantities listed in schema.yaml do not match the actual available native quantities). 

(To test this bug, one can obtain the actual available native quantities by setting `{'schema_filename': 'X'}` when loading the catalog, and compare the list with the quantities listed in schema.yaml.)

Some of the mismatched native quantities are used in quantity modifiers, and these modifiers need to be versioned to work on both 1.1p and 1.2i:

- `base_Blendedness_abs_flux` -> `base_Blendedness_abs_instFlux`
- `base_SdssCentroid_xSigma` -> `base_SdssCentroid_xErr`
- `base_SdssCentroid_ySigma` -> `base_SdssCentroid_yErr`
- `*_base_PsfFlux_flux` -> `*_base_PsfFlux_instFlux`
- `*_base_PsfFlux_fluxSigma` -> `*_base_PsfFlux_instFluxErr`

One can obtain this list by running the following code. 

```
import GCRCatalogs

cat = GCRCatalogs.load_catalog('dc2_object_run1.2i_all_columns')
native_quantities_needed = set(cat._translate_quantities(cat.list_all_quantities()))

cat_true = GCRCatalogs.load_catalog('dc2_object_run1.2i_all_columns', {'schema_filename': 'X'})
native_quantities_true = set(cat_true.list_all_native_quantities())

print(native_quantities_needed - native_quantities_true)
```

Hence, to fix this bug, we need to 
- [ ] update `schema.yaml`
- [x] update affected quantity modifiers in the reader.

yymao:
This PR updates the object catalog reader: If the schema yaml file is not available, the reader can now automatically generate the schema from the hdf5 data files as a fallback mechanism.

A bonus of this fallback feature is that we can actually use it to generate the schema yaml file, despite the fact that the schema yaml should really be generated by `merge_tract_cat.py`; see https://github.com/LSSTDESC/DC2-production/issues/292. To do so, one can simply run:

```python
import GCRCatalogs
for cat in ('dc2_object_run1.1p', 'dc2_object_run1.2i', 'dc2_object_run1.2i_all_columns'):
    GCRCatalogs.load_catalog(cat).generate_schema_yaml(overwrite=True)
```

@wmwv, once you approve this PR, I'll re-generate the schema yaml files by running the code snippet above.

With #226 and this PR, we fix #225.

### DIFF
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -76,6 +76,8 @@ class TableWrapper():
     And a schema to specify dtypes and default values for missing columns.
     """
 
+    _default_values = {'i': -1, 'b': False}
+
     def __init__(self, file_handle, key, schema=None):
         if not file_handle.is_open:
             raise ValueError('file handle has been closed!')
@@ -135,10 +137,12 @@ class TableWrapper():
         If not found, default to np.float64 and np.nan.
         """
         schema_this = self._schema.get(key, {})
-        return self._generate_constant_array(
-            dtype=schema_this.get('dtype', np.float64),
-            value=schema_this.get('default', np.nan),
+        dtype = schema_this.get('dtype', np.float64)
+        default = schema_this.get(
+            'default',
+            self._default_values.get(np.dtype(dtype).kind, np.nan)
         )
+        return self._generate_constant_array(dtype=dtype, value=default)
 
     def _generate_constant_array(self, dtype, value):
         """
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -6,6 +6,7 @@ import os
 import re
 import warnings
 import itertools
+import shutil
 
 from distutils.version import StrictVersion # pylint: disable=no-name-in-module,import-error
 import numpy as np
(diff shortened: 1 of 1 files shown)


GCRCatalogs/dc2_object.py
```diff
@@ -171,8 +196,15 @@ def __init__(self, file_handle, key, schema=None):
         super(ObjectTableWrapper, self).__init__(file_handle, key, schema)
         # Add the schema info for tract, path
         # These values will be read by `get_constant_array`
-        self._schema['tract'] = {'dtype': int, 'default': self.tract}
-        self._schema['patch'] = {'dtype': str, 'default': self.patch}
+        self._schema['tract'] = {'dtype': 'int64', 'default': self.tract}
+        self._schema['patch'] = {'dtype': '<U3', 'default': self.patch}
```
wmwv: In practice, this will be fine, but formally patches could be indexed by a more than one-digit number.  E.g., patch `31,5`

If this happened, it would break some other stuff where we do int<->string conversions for patch names.  But perhaps throw in a comment to the effect that we are assuming single-digit patch indexes.

yymao: Good point. There's in fact no need to specify the number of character for this particular case. 

Code updated. Great if you can re-approve. Thanks!

PR #231, opened 2018-11-22
Adds knots for version 1.1

EiffL:
This PR just adds a catalog config for the new knots catalog sampled for cosmoDC2 v1.1


yymao:
Thanks for the PR @EiffL. I think this should be merged *after* `cosmoDC2_v1.1` is available in `GCRCatalogs`.

yymao:
@EiffL should we name it `cosmoDC2_v1.1_addon_knots` (instead of `cosmoDC2_v1.1.0_addon_knots`)?

yymao:
@EiffL can you update the config (and the underlying knot data) to match 1.1.4 instead?

yymao:
superseded by #239

PR #235, opened 2018-11-30
add size check to cosmoDC2 1.1.4

yymao:
Since cosmoDC2 1.1.4 is going to be used for instance catalog generation, it's better that we ensure the underlying files are not corrupted, and hence we add this size check.


yymao:
This PR has been tested with the tutorial notebooks.

yymao:
Hmm, I didn't realized `cosmoDC2_v1.1.4` and `cosmoDC2_v1.1.4_image` are the same. Should we remove `cosmoDC2_v1.1.4` then? Becuase `cosmoDC2_v1.0` is actually larger and it might give people the impression that `cosmoDC2_v1.1.4` is larger.

evevkovacs:
Yes, I think that might be a good idea. Then there is no confusion and only cosmoDC2_v1.1.4_image is available. (Due to the time constraint for starting Run 2.1i, we only generated as many healpixels as required for the image sims.)

yymao:
OK, working on it. How about 1.1 and 1.1.3? Are they of the size of image sim as well?

yymao:
@evevkovacs OK, take a look at the recent changes: https://github.com/LSSTDESC/gcr-catalogs/pull/235/files/da4655a..25dfe0a

yymao:
@evevkovacs we should avoid remove previously released catalogs as it breaks backward compatibility. I'll remove the reference in README though. 

I'll also remove 1.1.0. Should we keep 1.1.3 then? It seems to be that would be useful for some comparison that people may want to do?

yymao:
OK, @evevkovacs I updated README so only v1.0 and v1.1.4 are references. 

I keep v0.1 and v0.4 (for backward compatibility) and v1.1.3 (for potentially useful comparison with v1.1.4), but these catalog versions are not displayed by default. 

I hope this arrangement is OK?

evevkovacs:
Yes, agreed, we shouldn't remove anything that has been released,but I wasn't suggesting that.  (ie keep all v1.0*)  **v1.1.0** was never released. It has bugs, and should not be used. Scott did some preliminary stuff with it, but it is super-ceded by v1.1.3 and v1.1.4. As regards v1.1.3 and 1.1.4, if we have space we can keep both. The differences between them are only the red-sequence galaxies in massive clusters. So I think it is fine to have both for the time being.

evevkovacs:
Yes, v0.1 was officially released so definitely keep it. v0.2 and v0.4 were not released. Are you aware of anybody outside of CS who used them? The only config file I see for v0.4 is cosmoDC2_v0.4_test.yaml and it points to CS/cosmoDC2/all_v0.4.2_centrals, which has bugs. I am fine with either deleting v0.4 or keeping v0.4.12_full in the CS directory, and update the config file for v0.4_test to point to that.  What do you think?

yymao:
I'm ok with removing 0.4 in that case. For 1.1.3 should we just keep the "small" one so we can remove some files?

yymao:
@evevkovacs I went ahead to remove cosmoDC2 v0.4 and v1.1.3_image (but kept v1.1.3_small).  Can you review again (re-approve or let me know what further changes we should make)? Thanks!

evevkovacs:
OK I approved. Keeping v1.1.3_small is a good compromise. Thanks!

PR #237, opened 2018-11-30
added cosmodc2 mass definition

### LINKED ISSUE #236: missing halo mass definition in cosmoDC2 catalog/reader
@marina-ricci pointed out that the `halo_mass_def` attribute in cosmoDC2 reader is missing. 

Indeed, the `halo_mass_def` is not defined for `CosmoDC2ParentClass` in `cosmodc2.py`. Also, the value is not set in the catalog files or the config files. 

@dkorytov what's the halo mass definition? I imagine it's a lot of trouble to put this into the data files, but we can put it in the config and modify the reader accordingly?

dkorytov:
fixes #236


GCRCatalogs/cosmodc2.py
```diff
@@ -139,6 +139,7 @@ def _subclass_init(self, catalog_root_dir, catalog_filename_template, **kwargs):
         self.sky_area, self._native_quantities, self._quantity_info = self._process_metadata(**kwargs)
         self._quantity_modifiers = self._generate_quantity_modifiers()
         self._native_filter_quantities = {'healpix_pixel', 'redshift_block_lower'}
+        self.halo_mass_def = "FoF, b=0.168"
```
yymao: Thanks @dkorytov. Let's allow overwriting this attribute when specified config file, for potential future use. 

```suggestion
        self.halo_mass_def = kwargs.get('halo_mass_def', 'FoF, b=0.168')
```

PR #238, opened 2018-11-30
Applying emissionLine 

### LINKED ISSUE #190: fix cosmoDC2 native quantities: emissionLines and ContinuumLuminosity
In the current cosmoDC2 reader, we manually remove native quantities `emissionLines/*` and `*ContinuumLuminosity` ([see here](https://github.com/LSSTDESC/gcr-catalogs/blob/eab3974/GCRCatalogs/cosmodc2.py#L424)). 

@dkorytov @evevkovacs, has this been fixed in the catalog? If so, we should update the reader. If not, should we fix it before we release v1.0?

dkorytov:
Moved the emissionLine/ContinuumLuminosity fix to only apply to earlier versions of the catalog. 

I checked that the fix was required at the earlier versions and not needed at later versions. 

fixes #190


dkorytov:
Sure thing. :)

PR #244, opened 2018-12-13
Update SCHEMA.md with DC2 truth catalogs

JulienPeloton:
This PR updates the SCHEMA.md to include native quantities from DC2 reference catalogs (mostly `dc2_truth_run1.1_static` and `dc2_truth_run1.2_static`).

There are currently missing definitions (`star`, `agn`, `sprinkled`), and doubts concerning other definitions (`redshift`, `<band>`).

@yymao @wmwv @fjaviersanchez @plaszczy your inputs are welcome!

### DIFF
--- a/GCRCatalogs/SCHEMA.md
+++ b/GCRCatalogs/SCHEMA.md
@@ -140,10 +140,9 @@ Quantity Label | Unit | Definition | GCRbase | DPDD
 `blendedness` | - | measure of how flux is affected by neighbors: (1 - flux.child/flux.parent) (see 4.9.11 of [1705.06766](https://arxiv.org/abs/1705.06766)) |   |   |
 `extendedness` | - | 0:star, 1:extended.  DM Stack `base_ClassificationExtendedness_value` |   |   |
 
-## Schema for DC2 reference catalogs
+## Schema for DC2 truth catalogs
 
-DC2 reference catalogs correspond to DC2 Truth Catalog such that `dc2_truth_run1.1_static` or `dc2_truth_run1.2_static`
-Definition of quantities can be accessed using `get_quantity_info`:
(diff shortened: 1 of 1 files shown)


yymao:
@JulienPeloton thanks for the PR!

First I want to mention that this `SCHEMA.md` file was intended for documenting "derived" (i.e. GCR-normalized) quantities, not the native ones. The idea is that the availability and definition of native quantities are more likely to change, and we don't want to update this document as often. 

Of course, the definition of native quantities should still be available. This is made possible by the `get_quantity_info` in GCR. In other words, we ask the catalog creators to take care of documenting native quantities. For example:

```python
cat = GCRCatalogs.load_catalog('dc2_truth_run1.2_static')
cat.get_quantity_info('healpix_2048')
```
which will give you
```python
{'description': 'healpixel containing the object (nside=2048; nested)'}
```

Hence, I think the newly added table may not be necessary (in this particular file), but maybe we should add a note about `get_quantity_info` here?

JulienPeloton:
Thanks @yymao  for the detailed explanations. I did not notice the descriptions were stored inside the database

```
+------------+--------------------+
|        name|         description|
+------------+--------------------+
|healpix_2048|healpixel contain...|
|   object_id|an int uniquely i...|
|        star|an int; ==1 if a ...|
|         agn|an int; ==1 if ga...|
|   sprinkled|an int; ==1 if ob...|
|          ra|          in degrees|
|         dec|          in degrees|
|    redshift|   cosmological only|
|           u|observed lsst u m...|
|           g|observed lsst g m...|
|           r|observed lsst r m...|
|           i|observed lsst i m...|
|           z|observed lsst z m...|
|           y|observed lsst y m...|
+------------+--------------------+
```

I agree then the table does not need to be reported here, just a note on `get_quantity_info` should be sufficient.

JulienPeloton:
OK it should be reflected in the last commit.

PR #253, opened 2019-01-24
add config for new extragalactic truth catalog

danielsf:
This PR adds the new truth catalog for static extra galactic sources.  It includes columns

u - observed u-band magnitude
du_mw - extinction in u-band due to Milky Way dust
du_internal - extinction in u-band due to internal dust
etc.

It does not include truth information for stars.  Let me know if that is a problem.


yymao:
@danielsf does it only has u band, or that was just an example? 

Not having stars may be an issue with matching, but should be OK for folks that are looking at galaxies only (e.g. WL, CL). I wonder what @rmandelb thinks.

danielsf:
It has all the bands.  I did not want to type them out

danielsf:
Should we get Eli's input.  The impression I got from yesterday's phonecon was that his needs were driving the push for this catalog.

yymao:
Sure. My guess is that for @erykoff's specific need, not having stars is not an issue?

erykoff:
I don't need stars for these comparisons.  What's driving the question is that I want to be able to distinguish differences in red cluster galaxies between the extragalactic catalog and the object catalog that are due to the way the galaxies were simulated (in the truth catalog) and issues with the processing.  I'm sure this won't be the only use case.

yymao:
Minor question: for the truth catalog, galaxies are always static objects, right? If so we can just name this `dc2_truth_run1.2_galaxies.yaml` (which is what we used for run1.1)?

danielsf:
Some galaxies contain AGN, which vary.  These galaxies are going to appear as variable sources in the catalogs of detected objects.

katrinheitmann:
I am not an expert in any of this but we probably should keep in mind that for Run 2.1i a lot will be different regarding colors and in particular regarding colors in the truth catalog? I am worried that we are trying to fix something that in the future is not a problem anymore.

I might be totally wrong ...

On 1/24/19 10:55 AM, erykoff wrote:

I don't need stars for these comparisons. What's driving the question is that I want to be able to distinguish differences in red cluster galaxies between the extragalactic catalog and the object catalog that are due to the way the galaxies were simulated (in the truth catalog) and issues with the processing. I'm sure this won't be the only use case.

yymao:
@danielsf ah, right. So are galaxies that contain AGNs also excluded from this truth catalog?

danielsf:
@yymao This catalog includes the static magnitudes of the galaxies (i.e. the magnitudes of the bulge and disk).  All galaxies have an `agn` column that is `1` if the galaxy contains an AGN and `0` otherwise.

yymao:
@katrinheitmann I than for Run 2.x the truth catalog solution would be completely different, as we discussed in the CS telecon yesterday, but in the meantime we still need a truth catalog for people who want to analyze Run 1.2i, so I think an updated 1.2 truth catalog is still needed. This PR already provides a solution for the galaxies; the remaining question is whether we need the same for stars.

yymao:
I think this PR is ready to be merged. I'll wait till the end of the day in case there's further comments on this. 

There's still a separate question on whether we need stars with the extinction to be included in the truth catalog.

PR #260, opened 2019-02-08
Buzzard 1.9

j-dr:
Updated buzzard reader to support catalog version 1.9. Of most relevance to most people is that catalog now contains WFIRST and VISTA magnitudes in addition to LSST and DES bands that were in old versions. It is also the version documented in the buzzard paper, so should be the preferred buzzard catalog to use for the time being.


GCRCatalogs/buzzard.py
```diff
@@ -175,11 +175,23 @@ def _subclass_init(self,
                 'velocity_z': 'truth/VZ',
             }
 
+            if self.version > 1.6:
```
yymao: Does this work? I believe `self.version` is a string, right?

GCRCatalogs/buzzard.py
```diff
@@ -175,11 +175,23 @@ def _subclass_init(self,
...
+
+                    self._quantity_modifiers['Mag_true_{}_z0'.format(b)] = (_abs_mask_func, 'auxmag/AMAG/{}'.format(i))
+                    self._quantity_modifiers['mag_true_{}'.format(b)] = (_mask_func, 'auxmag/TMAG/{}'.format(i))
+                    self._quantity_modifiers['mag_{}'.format(b)] = (_mask_func, 'auxmag/LMAG/{}'.format(i))
+
+            else:
+                self._quantity_modifiers['Mag_true_{}_lsst_z0'.format(b)] = (_abs_mask_func, 'lsst/AMAG/{}'.format(i))
+                self._quantity_modifiers['mag_true_{}_lsst'.format(b)] = (_mask_func, 'lsst/TMAG/{}'.format(i))
+
             for i, b in enumerate('ugrizyY'):
```
yymao: Just want to make sure it is intended that this for loop should be outside of the else clause?

GCRCatalogs/buzzard.py
```diff
@@ -188,6 +200,10 @@ def _subclass_init(self,
                     self._quantity_modifiers['mag_true_{}_des'.format(b)] = (_mask_func, 'truth/TMAG/{}'.format(i))
                     self._quantity_modifiers['mag_{}_des'.format(b)] = (_mask_func, 'truth/OMAG/{}'.format(i))
                     self._quantity_modifiers['magerr_{}_des'.format(b)] = (_mask_func, 'truth/OMAGERR/{}'.format(i))
+                
+                
+
+
```
yymao: remove the above extra empty lines

yymao:
@j-dr just want to check in with you to see if you can address the review comments. Thanks!

GCRCatalogs/buzzard.py
yymao: This indeed does not work. It raises `TypeError`.

yymao:
Also, during initiation, `self.check_healpix_pixels()` fails for v1.9.2_3. I need to look into it to know what exactly is wrong, but I suspect some naming convention may have changed. This comment is to make sure that I remember to investigate this. 

(Thanks to @cosmicshear for reporting this.)

j-dr:
I think the problem with self.check_healpix_pixels  might be a permissions issue. I’ll address these issues update this PR.

j-dr:
Okay, I just updated the version checking and changed a few things about how the new magnitudes are read in. I think the permissions should work, but I'm going to need to move this catalog very soon since it's in my scratch space right now. It's 5Tb though with the new sets of magnitudes and increased depth, so I'm not sure where the best place to put it is.

GCRCatalogs/buzzard.py
```diff
@@ -198,12 +203,6 @@ def _subclass_init(self,
                     i -= 1
                     self._quantity_modifiers['Mag_true_{}_des_z01'.format(b)] = (_abs_mask_func, 'truth/AMAG/{}'.format(i))
                     self._quantity_modifiers['mag_true_{}_des'.format(b)] = (_mask_func, 'truth/TMAG/{}'.format(i))
-                    self._quantity_modifiers['mag_{}_des'.format(b)] = (_mask_func, 'truth/OMAG/{}'.format(i))
-                    self._quantity_modifiers['magerr_{}_des'.format(b)] = (_mask_func, 'truth/OMAGERR/{}'.format(i))
```
yymao: Is the removal of these two lines (201-202) intended?

j-dr: Yes, they were intended. I don't have noisy DES magnitudes in the new version, so these are only read in the old version. I did break backwards compatibility by accident though. The update I just pushed should fix this.

README.md
```diff
@@ -55,11 +55,12 @@ Confluence page (*DESC member only*).
    - `buzzard`: full catalog, DES Y3 area
    - `buzzard_test`: same as `buzzard` but a small subset for testing purpose / faster access
    - `buzzard_high-res`: higher resolution, smaller sky area
-   - `buzzard_v1.6_x`: different realizations of `buzzard`. You can run
+   - `buzzard_v1.6_x`: different realizations of an older version of `buzzard`. You can run
      ```python
      sorted((name for name in GCRCatalogs.get_available_catalogs(False) if name.startswith('buzzard_v1.6')))
      ```
      to see all available versions.
+   - `buzzard_v1.9.2_x`: different realizations of the version of the `buzzard` catalog documented in arxiv:1901.02401.
```
yymao: Move this line to above the python code snippet (i.e., to Line 59).

yymao:
I went head and fixed the README. Also, do you want to change the `buzzard` alias to point to `buzzard_1.9.2_3` instead of `buzzard_1.6` (which is currently used).

j-dr:
Thanks Yao. Yeah, would be good to change the default to version 1.9.3.

PR #263, opened 2019-02-18
Issues/254

wmwv:
Adding cModelFlux, cModelFluxErr to gcr-catalogs is a trivial two-line change.

I snuck in a fix to the Run 1.2i tract_4850 YAML config as well.  Not formally part of this tick.  But I ran across it while implementing and testing and figured I'd just include it here as well.

I have also updated at NERSC the trim catalogs and DPDD Parquet files.

### DIFF
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -364,6 +364,9 @@ class DC2ObjectCatalog(BaseGenericCatalog):
             modifiers['psFluxErr_{}'.format(band)] = '{}_base_PsfFlux_{}{}'.format(band, FLUX, ERR)
             modifiers['I_flag_{}'.format(band)] = '{}_base_SdssShape_flag'.format(band)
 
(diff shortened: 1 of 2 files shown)


GCRCatalogs/dc2_object.py
```diff
@@ -364,6 +364,9 @@ def _generate_modifiers(pixel_scale=0.2, bands='ugrizy',
             modifiers['psFluxErr_{}'.format(band)] = '{}_base_PsfFlux_{}{}'.format(band, FLUX, ERR)
             modifiers['I_flag_{}'.format(band)] = '{}_base_SdssShape_flag'.format(band)
 
+            modifiers['cModelFlux_{}'.format(band)] = '{}_modelfit_CModel_{}'.format(band, FLUX)
+            modifiers['cModelFluxErr_{}'.format(band)] = '{}_modelfit_CModel_{}{}'.format(band, FLUX, ERR)
+
```
yymao: should we add `cModelFlux_flag_<band>`?

wmwv: Thank you for catching this omission!

PR #270, opened 2019-02-27
U/wmwv/update schema with cmodel njy

wmwv:
* Update the SCHEMA to document modelfit flux, flux err, and flux flag
* Note that x, y values are in local tract+patch coordinates.
* Note flux is now in nJy.


yymao:
thanks @wmwv

PR #273, opened 2019-03-05
Buzzard 1.9 magnitude index error fix

j-dr:
This PR fixes an index error that existed when reading in non-DES magnitudes.


yymao:
@j-dr, thanks. Merge at will.

PR #286, opened 2019-04-25
Proposed changes to fix halo_mass bugs

evevkovacs:
@yymao  Here are the changes to fix the halo_mass units issue. I sent you an email about my attempt to fix the units reported by get_quantity_info. There may be a better way. Please let me know.

### DIFF
--- a/GCRCatalogs/cosmodc2.py
+++ b/GCRCatalogs/cosmodc2.py
@@ -319,7 +319,7 @@ class CosmoDC2GalaxyCatalog(CosmoDC2ParentClass):
             'convergence': 'convergence',
             'magnification': (lambda mag: np.where(mag < 0.2, 1.0, mag), 'magnification'),
             'halo_id':       'uniqueHaloID',
-            'halo_mass':     'hostHaloMass',
+            'halo_mass':     (lambda x: x/self.cosmology.h, 'hostHaloMass'),
             'is_central':    (lambda x: x.astype(np.bool), 'isCentral'),
             'stellar_mass':  'totalMassStellar',
             'stellar_mass_disk':        'diskMassStellar',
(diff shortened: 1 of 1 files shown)


GCRCatalogs/cosmodc2.py
```diff
@@ -451,6 +451,13 @@ def _generate_quantity_modifiers(self):
 
         return quantity_modifiers
 
+    def get_quantity_info(self,q):
+        info_dict = self._get_quantity_info_dict(q)
+        if q == 'hostHaloMass':
+            if 'units' in info_dict:
+                info_dict['units'] = 'Msun/h'
+        
+        return info_dict
```
yymao: We should update the upstream method instead. 

```python
    def _collect_native_quantities(self, fh, collect_info_dict=False):
        orig_output = super()._collect_native_quantities(fh, collect_info_dict)
        # a hot fix of the unit of native halo mass (hostHaloMass), which should be Msun/h
        if collect_info_dict:
            orig_output[1]['hostHaloMass']['unit'] = 'Msun/h'
        return output
```

evevkovacs: OK, this is now implemented and tested. This PR is ready to merge. baseDC2 does not have units so no modifications are necessary. protoDC2 remains to be checked.

evevkovacs:
@yymao  Oh, yes, I suppose we do since these are schema quantities. There are no unit designations to fix however. I'll make that change and re-commit. I'll need to fix this for the snapshot baseDC2 as well. Thanks for catching this.

evevkovacs:
@yymao  I fixed the quantity modifier for alphaq.py (protoDC2 reader) and added the code you suggested for modifying the units for cosmoDC2. That is, I defined _collect_native_quantities(self, fh, collect_info_dict=False) etc. But when I tested on protoDC2, it didn't work. I still got units of Msun for hostHaloMass. Does this function exist for the AlphaQGalaxyCatalog class? Should I use something else? Thanks

yymao:
@evevkovacs protoDC2 (`alphaq.py`) uses a different method to collect the quantity info. We will need to modify the `_get_native_quantity_info_dict` method. I'll be happy to make another PR for it, but can you confirm that protoDC2 also needs the fix 
I have thought that protoDC2 was schema conforming, but I could be wrong. 

In any case, this PR is indeed ready to be merged.

evevkovacs:
OK will merge now and address other issues next.

Issue #290, opened 2019-05-03
Create DIASource table reader


wmwv:
1. [x] Restructure `DC2SourceCatalog` and `DC2ForcedSourceCatalog` to at the minimum inherit from some slightly more generic class that will be Parquet-based.
2. [x] `DC2DiaSourceCatalog` to inherit from this new class
3. [x] Define quantity modifiers for `DC2DiaSourceCatalog` using list of translations from
https://github.com/lsst/ap_association/blob/master/python/lsst/ap/association/mapApData.py#L109
4. [x] Define reader config for the Run1.2p test run.  Perhaps `dc2_dia_source_run1.2p_test`
5. [x] Test and verify reader.

PR #292, opened 2019-05-08
Adds support for metacal catalog

EiffL:
This PR adds a catalog reader for a preliminary version of the metacal catalog for tract 5063 in run 1.2i. Still WIP

@wmwv as mentioned in the DC2-production repo, this particular version of the metacal catalog contains the reference catalog, which we would eventually remove and use a composite catalog instead. But does this broadly corresponds to what you had in mind?

@joezuntz we need to formally declare what metacal columns we want to have access to in this `dc2_metacal.py`. In the data access framework, the reader also defines the official schema, acting as a standardization interface between the raw DM stack names and the user. Shall we just declare here the columns you were emulating here: https://github.com/LSSTDESC/TXPipe/blob/9512dda4b299a5514869d3defdaf517e55fb7393/txpipe/input_cats.py#L333 ?


wmwv:
@EiffL Yes, this is broadly exactly the right direction.

EiffL:
I have added the definition of most metacal fields, adopting the current TxPipe naming conventions for the columns. The ones not explicitly declared in this schema are actually not used anywhere in TxPipe, so in principle, it should be straightforward to load this catalog and run txpipe as is.

Let me know if I need to do anything more before we can merge this branch. This tiny metacal catalog is intended for testing, I don't expect this schema to be final at this stage.

joezuntz:
I was looking at this catalog yesterday and couldn't see any measured magnitude values - can we get those from somewhere?

Also there's a challenge which is that for redshift estimation we need metacalibrated values for all the magnitudes, whereas for .  In the meantime we can just ignore this factor (which lead to a selection bias when constructing tomographic bins), but in future we do need it.  In DES two metacal runs were done, one on just the bands used for the shapes and one on all the bands, so I guess we need something like that.

Eventually metadetect may make us do this all differently of course.

joezuntz:
Also, I'd like to switch the TXPipe code to use an HDF5 shear catalog internally.  So can we drop the conversion of the g1 and g2 columns into combined g columns, since that was just to match the fits catalogs?

joezuntz:
And finally, could we add the Tpsf and flag columns?

EiffL:
Sure, sounds good!

wmwv:
@EiffL #290 has been merged.

Can you either rebase to master or make this PR with respect to the commit that merged #290?  I'm getting confused as I start to go through the code by seeing both the changes introduced by #290 and the changes here in this present ticket.

EiffL:
@wmwv ok, this should be better now.

So, one thing that just came up, in the DC2DMCatalog, there is no support to provide the `_generate_modifiers` function with extra flags and values defined in the config file. For instance, I have to apply a fix to the flux for this run of metacal, I would like to just have to enable a flag in the configuration file that would apply this fix only for this specific version of the catalog.

yymao:
On mobile so I haven't checked the code, but you can make `_generate_modifiers` a regular method to include `self` where you can store the flag, I think.

EiffL:
@yymao well... sure, but that forces me to redefine the `_subclass_init` function, which is not something I want to do. But, the `_subclass_init` function of `DC2DMCatalog` could forward its `**kwargs` to `_generate_modifiers`, this way I would have direct access to the config file option from within that function.

wmwv:
@EiffL Specifically on this issue, I would suggest that `fix_flux_scaling` be re-designed so that the config file just specifies a `flux_scaling`, that defaults to one.  On the initialization of the class, this kwargs is read.

Then the `_generate_modifiers` can just use this value by reading it from `self`.

EiffL:
@wmwv  Right... I'm just annoyed at having to duplicate the `_subclass_init` code in order to do that :-/

yymao:
In fact, all `kwargs` passed to `_subclass_init` are always (secretly) stored in the `self._init_kwargs` attribute ([ref](https://github.com/yymao/generic-catalog-reader/blob/master/GCR/base.py#L26)). 

But the better way may be to just use `super` in `_subclass_init`:

```python
    self._YOUR_FLAG = kwargs['YOUR_FLAG']
```

wmwv:
@EiffL I guess what I was thinking is that in `DC2MetacalCatalog` you can define
`_subclass_init` to just pick out the kwargs you want to add.
And then call `super()._subclass_init` to do the rest of the standard stuff.

This way you don't have to duplicate (and in the future maintain) the code from the base class `_subclass_init`.

Perhaps you're like, "yeah, I know that, but I wouldn't have to do any of that if we just update the base class `_generate_modifiers` to accept `**kwargs` and the base class `_subclass_init` to pass `**kwargs` through."

EiffL:
Thanks both, so I've implemented your solution, I'm fine with it, only had 2 concerns:
 - 1: making `_generate_modifiers` not static in my derived class, presumably there was a good reason for having this method static.
 - 2: It felt like a useful mechanism that we might have liked to automatically have at the level of the parent class

But I don't have very strong feelings there :-)

GCRCatalogs/dc2_metacal.py
```diff
@@ -33,12 +33,32 @@ class DC2MetacalCatalog(DC2DMCatalog):
...
+        self._bands = 'riz'
+
+        if kwargs.get('bands'):
+            self._bands = kwargs['bands']
+
+        if kwargs.get('fix_metacal_test3'):
+            # In Run 1.2i metacal_test3, the fluxes returned by metacal are not
+            # properly scaled by the pixel size, it's necessary to apply a
+            # 0.2**2 correction factor
+            self._flux_scaling = 1. / 0.2**2
```
yymao: Lines 41-52. This is mostly just preference... Sorry for being annoying.  
```python
        self._bands = kwargs.get('bands', 'riz')
 
		self._flux_scaling = 1.0
		if kwargs.get('fix_metacal_test3'):
            # In Run 1.2i metacal_test3, the fluxes returned by metacal are not
            # properly scaled by the pixel size, it's necessary to apply a
            # 0.2**2 correction factor
            self._flux_scaling /= (0.2**2)
```

EiffL: lol, sure :-D

wmwv:
The idea of making `_generate_modifiers` static is that this function often encodes the key translation information between native columns and DPDD (or other external standard) representations.  This might be information that someone would want to use without having to actually instantiate (and nominally load) a catalog.

We use this, e.g., in `merge_source_cat.py` to only extract the columns we actually need to present the columns promised in the `dc2_source.py` reader:

EiffL:
@wmwv Yeah, I saw that when I was making an updated version of merge_tract_cat... that's were my concern came from.

GCRCatalogs/dc2_metacal.py
```diff
@@ -0,0 +1,102 @@
...
+                modifiers['mcal_flux_{}{}'.format(band, variant)] = (
+                    lambda x: x * self._flux_scaling,
+                    'mcal_gauss_flux_{}{}'.format(band, variant)
+                )
+                modifiers['mcal_flux_err_{}{}'.format(band, variant)] = (
+                    lambda x: x * self._flux_scaling,
+                    'mcal_gauss_flux_err_{}{}'.format(band, variant)
+                )
+                modifiers['mcal_mag_{}{}'.format(band, variant)] = (
+                    lambda x: -2.5 * np.log10(x * self._flux_scaling) + 27.0,
```
wmwv: Define `27.0` in a named variable somewhere to document where it comes from (and if we ever might have to change it).

For the standard DM catalogs I'm directly reading the calibration information.  Does metacal explicitly use `27.0` as a standard reference zeropoint?

GCRCatalogs/dc2_metacal.py
```diff
@@ -0,0 +1,102 @@
...
+    Notes
+    -----
+    """
+    # pylint: disable=too-many-instance-attributes
+
+    FILE_DIR = os.path.dirname(os.path.abspath(__file__))
+    FILE_PATTERN = r'metacal_tract_\d+\.parquet$'
+    SCHEMA_FILENAME = 'schema.yaml'
+    META_PATH = os.path.join(FILE_DIR, 'catalog_configs/_dc2_metacal_meta.yaml')
+
```
wmwv: Remove line.  Just one blank line between methods with a class definition.

Here and below.

GCRCatalogs/dc2_metacal.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def _subclass_init(self, **kwargs):
+        """
+        Overrides default init method to apply various corrections to the catalog
```
wmwv: I would say "Wraps" instead of "Overrides".

GCRCatalogs/dc2_metacal.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        # Default values of reader parameters
+        self._bands = kwargs.get('bands', 'riz')
```
wmwv: I prefer being explicit about `default` with getting from a dict.

`self._bands = kwargs.get('bands', default='riz')`

I meet a lot of grad students who are unfamiliar with this behavior.

GCRCatalogs/dc2_metacal.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self._flux_scaling = 1.0
+        if kwargs.get('fix_metacal_test3'):
```
wmwv: I would suggest that the `catalog_configs/dc2_metacal_run1.2i_tract5063.yaml` file define a 'flux_scaling' and then this hardcoded bit of code here can just be replaced here in the code with:

`self._flux_scaling = kwargs.get('flux_scaling', default=1)`

And any future such fixes can live with their config files, without having to add a series of `fix_metacal_test4`... checks.

GCRCatalogs/dc2_metacal.py
EiffL: Yes, according to @esheldon , I'm  going to define a METACAL_ZEROPOINT variable and set it to 27.0

GCRCatalogs/dc2_metacal.py
EiffL: :-/ I like having in one place the set of corrections that need to be applied for each metacal versions.  I also don't like exposing the option to apply whatever correction at the level of the config file. when someone sees in the config file `apply_metacal_test3_fix` they understand exactly what that means. If instead we have 'flux_scaling=0.004' the next person that generate the metacal catalog may think "oh well, I guess that number is there for some good reason"

GCRCatalogs/dc2_metacal.py
EiffL: argh, almost, I'm getting:
```
~/repo/gcr-catalogs/GCRCatalogs/dc2_metacal.py in _subclass_init(self, **kwargs)
     40         """
     41         # Default values of reader parameters
---> 42         self._bands = kwargs.get('bands', default='riz')
     43 
     44         self._flux_scaling = 1.0

TypeError: get() takes no keyword arguments
```

GCRCatalogs/dc2_metacal.py
```diff
@@ -0,0 +1,102 @@
...
+    SCHEMA_FILENAME = 'schema.yaml'
+    META_PATH = os.path.join(FILE_DIR, 'catalog_configs/_dc2_metacal_meta.yaml')
+    METACAL_ZEROPOINT = 27.0
+
+    def _subclass_init(self, **kwargs):
+        """
+        Wraps default init method to apply various corrections to the catalog.
+        """
+        # Default values of reader parameters
+        self._bands = kwargs.get('bands', default='riz')
```
yymao:
```suggestion
        self._bands = kwargs.get('bands', 'riz') # default to 'riz' if 'bands' not exist
```

GCRCatalogs/dc2_metacal.py
```diff
@@ -0,0 +1,102 @@
...
+                modifiers['mcal_flux_{}{}'.format(band, variant)] = (
+                    lambda x: x * self._flux_scaling,
+                    'mcal_gauss_flux_{}{}'.format(band, variant)
+                )
+                modifiers['mcal_flux_err_{}{}'.format(band, variant)] = (
+                    lambda x: x * self._flux_scaling,
+                    'mcal_gauss_flux_err_{}{}'.format(band, variant)
+                )
+                modifiers['mcal_mag_{}{}'.format(band, variant)] = (
+                    lambda x: -2.5 * np.log10(x * self._flux_scaling) + METACAL_ZEROPOINT,
```
yymao:
```suggestion
                    lambda x: -2.5 * np.log10(x * self._flux_scaling) + self.METACAL_ZEROPOINT,
```

GCRCatalogs/dc2_metacal.py
wmwv: My fault.  There must be something confused in my brain about this.

EiffL:
@yymao Awesome thanks, I just tested the composite catalog `dc2_object_run1.2i_tract5063_with_metacal.yaml`, everything seems to work \o/

yymao:
Thanks @EiffL. Everything looks good now. Merge at will.

EiffL:
Great :-) thanks both, this data access framework is definitely very neat !

PR #299, opened 2019-05-10
switch from yaml.load to yaml.safe_load

### LINKED ISSUE #298: yaml.load deprecation warning
With the most recent versions of pyyaml, the plain `yaml.load()` is deprecated without additional options concerning safety, leading to this warning when importing the library:

```
> python -c 'import GCRCatalogs'
/global/cscratch1/sd/zuntz/env1/lib/python3.7/site-packages/GCRCatalogs/register.py:23: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  config = yaml.load(f)
```

The seven `yaml.load`s in gcr-catalogs should be replaced with either `yaml.safe_load`, or `yaml.full_load` if we expect configuration files to contain anything beyond the most basic structures (i.e. anything that needs pickling).  I would guess that the former would make more sense here.

There is more info here: https://msg.pyyaml.org/load

yymao:
This PR fixes #298

### DIFF
--- a/GCRCatalogs/cosmodc2.py
+++ b/GCRCatalogs/cosmodc2.py
@@ -132,7 +132,7 @@ class CosmoDC2ParentClass(BaseGenericCatalog):
         if kwargs.get('check_md5', True) or kwargs.get('check_size', True):
             try:
                 with open(CHECK_FILE_PATH, 'r') as f:
-                    self.file_check_info = yaml.load(f)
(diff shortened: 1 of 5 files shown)


yymao:
Thank you @wmwv, but I still need your approval :)

PR #308, opened 2019-05-30
serialize the metadata dump for photoz catalog

yymao:
This PR serializes the metadata dump for photoz catalog so that the metadata yaml file can be loaded by load_safe


yymao:
Thanks, @sschmidt23.

PR #311, opened 2019-05-30
remove baseDC2 v0.4.5

yymao:
I found that baseDC2 v0.4.5 is no longer available on disk. Is the intent to remove these catalogs? If so, the corresponding configs should be removed too. 

This PR removes baseDC2 v0.4.5 configs.


evevkovacs:
Yes, thanks, that is correct. Those catalogs were obsolete.

yymao:
Thanks, @evevkovacs.

PR #321, opened 2019-06-03
Add photo-z mag err catalog reader

yymao:
This PR adds a photo-z magnitude error estimate catalog reader (the catalog matches to cosmoDC2) and the corresponding catalog config.

The catalog data are provided by @sschmidt23. @sschmidt23 can you double check if the reader is implemented correctly and also suggest descriptions when appropriate? 

@MarkusMichaelRau, can you check out this branch and test the reader to see if it works as needed? 

Thank you both.

--
Also, we now have 3 different readers for 3 different photo-z products, and have 4 different readers for 4 different cosmoDC2 add-ons. We need to an unified format to prevent reader proliferation, which is already taking place (cc @wmwv).


MarkusMichaelRau:
Works perfectly. Thanks!

yymao:
Thanks, @sschmidt23. The reader proliferation issue is indeed something that the DATF needs to think about and provide directions for catalog creators. For now I'll still merge this PR.

PR #326, opened 2019-06-11
Refactor register module and add `based_on` option for catalog configs

yymao:
This PR refactors `register` module by introducing the `ConfigRegister` class, to make things a bit more readable and easier to maintain, and also to reduce the load time of `GCRCatalogs`.

Also, a new `based_on` option is introduced with this refactoring, to allow users creating catalog configs that are based on other catalog configs, but with slight updated options. There has been the `alias` option that allows users creating catalog aliases, but in many cases a user may want to create a config that is almost identical to another one, but with some small adjustments to some options. 

With this feature, the yaml file of `dc2_object_run1.1p_tract4850`, for example, can now be written as:
```yaml
based_on: dc2_object_run1.1p
filename_pattern: '(?:merged|object)_tract_4850\.hdf5$'
```
which would mean: "use `dc2_object_run1.1p` but also using other options here to update the original yaml". This can reduce duplications among catalog config files.

This PR by itself does not include changes to existing catalog configs.

### DIFF
--- a/GCRCatalogs/composite.py
+++ b/GCRCatalogs/composite.py
@@ -3,7 +3,7 @@ composite reader
 """
 import warnings
 from GCR import CompositeCatalog, MATCHING_FORMAT, MATCHING_ORDER
-from .register import load_catalog, load_catalog_from_config_dict, available_catalogs
+from .register import load_catalog, load_catalog_from_config_dict, has_catalog
 
 class CompositeReader(CompositeCatalog):
     def __init__(self, **kwargs):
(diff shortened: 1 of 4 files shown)


yymao:
Side note: This PR does not necessarily need to be shipped with v0.11.0. I am current running a full test on v0.11.0 to make sure everything work. Once the test is done I'll tag the master and make a release. If this PR makes it that's great, it not it's OK too.

yymao:
Thanks @wmwv. A full test run on NERSC with this PR also passes. Merging and tagging now.

PR #329, opened 2019-06-20
Issues/293 - Add DIA Object reader

wmwv:
Adds dc2_dia_object reader class and catalog config for Run 1.2p test.
Data have been installed at NERSC for this test Run 1.2p DIA run.

See verification Notebook in DC2-Production to demonstrate basic sanity checking.


GCRCatalogs/dc2_dia_object.py
```diff
@@ -0,0 +1,107 @@
...
+# 'pmParallaxNdata': 
+#            'totFluxMean': 'ip_diffim_forced_PsfFlux_instFlux',
+#            'totFluxMeanErr': 'ip_diffim_forced_PsfFlux_instFluxErr',
+#            'totFluxSigma': 'ip_diffim_forced_PsfFlux_instFluxErr',
+#  'lcPeriodic'
+#  'lcNonPeriodic'
+#  These are possible to calculate if you require the Object Table
+#            'nearbyObj':
+#            'nearbyObjDist':
+#            'nearbyObjLnP':
```
yymao: Lines 66-81 --- it's ok to keep these comment-out lines, but can you format them a bit better? (e.g., indentation, additional #'s for comment in comment)

GCRCatalogs/catalog_configs/dc2_dia_object_run1.2p_test.yaml
```diff
@@ -0,0 +1,7 @@
+subclass_name: dc2_dia_object.DC2DiaObjectCatalog
+base_dir: /global/projecta/projectdirs/lsst/production/DC2_PhoSim/Run1.2p/dpdd/dia_object_catalog_test
+schema_filename: schema.yaml
+filename_pattern: 'dia_object_tract_\d+\.parquet$'
+description: DC2 Run 1.2p DIA Object Catalog test patch
+creators: ['Michael Wood-Vasey']
+included_by_default: true
```
yymao: remove this from the default listing
```suggestion
```

GCRCatalogs/dc2_dia_object.py
```diff
@@ -0,0 +1,107 @@
+"""
+DC2 DIA Object Catalog Reader
+"""
+
+import os
+import re
```
yymao: unused
```suggestion
```

GCRCatalogs/dc2_dia_object.py
```diff
@@ -0,0 +1,107 @@
...
+import warnings
```
yymao: unused
```suggestion
```

GCRCatalogs/dc2_dia_object.py
```diff
@@ -0,0 +1,107 @@
...
+        """Creates a dictionary relating native and homogenized column names
+
+        Args:
+            dm_schema_version (int): DM schema version (1, 2, or 3)
+
+        Returns:
+            A dictionary of the form {<homogenized name>: <native name>, ...}
+        """
+        flux_name = 'flux' if dm_schema_version <= 2 else 'instFlux'
+        flux_err_name = 'Sigma' if dm_schema_version <= 1 else 'Err'
```
yymao: `flux_name` and `flux_err_name` are not used

GCRCatalogs/dc2_dia_object.py
```diff
@@ -0,0 +1,107 @@
...
+        modifiers['clean'] = modifiers['good']
+
+        multiband_columns_to_copy = [
+            'psFluxMean', 'psFluxMeanErr',
+            'psFluxSigma', 'psFluxChi2', 'psFluxNdata']
+
+        for band in bands:
+            for base_col in multiband_columns_to_copy:
+                    col_name = f'{base_col}_{band}'
+                    modifiers[col_name] = col_name
```
yymao: Lines 93, 94: bad indentation

GCRCatalogs/dc2_dia_object.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+
+import os
+import re
+import warnings
+
+import numpy as np
+
+from .dc2_dm_catalog import (DC2DMCatalog,
+                             convert_flux_to_nanoJansky,
```
yymao: `convert_flux_to_nanoJansky` unused 
```suggestion
```

GCRCatalogs/dc2_dia_object.py
```diff
@@ -0,0 +1,107 @@
...
+    -----
+    """
+    # pylint: disable=too-many-instance-attributes
+    FILE_DIR = os.path.dirname(os.path.abspath(__file__))
+    FILE_PATTERN = r'dia_object_\d+\.parquet$'
+    SCHEMA_FILENAME = 'schema.yaml'
+    META_PATH = os.path.join(FILE_DIR, 'catalog_configs/_dc2_dia_object_meta.yaml')
+
+    @staticmethod
+    def _generate_modifiers(dm_schema_version=3, bands='ugrizy'):
```
yymao: the `bands` argument does not appear in the base class. Is the intention to overwrite the call signature? If so, add `# pylint: disable=arguments-differ` and also update the docstring.

wmwv: Yes, we will need `bands` for [DIA]Object tables.

wmwv: However, we don't need `dm_schema_version` for DIAObject.  It's not used.  Should I:
A) Leave that in?
B) Take it out, leaving just `bands='ugrizy'`, which leaves the risk of a positional argument call getting confused and overriding the default `bands` with, e.g., `3` (for the dm_schema_version)?
C) Use `def _generate_modifiers(*, bands='ugrizy')` to ignore any positional arguments?

wmwv:
@yymao Better now.  My apologies for not linting before submitting (or even reading the Travis output).  My workflow when working remotely on Cori needs some improvement.

GCRCatalogs/dc2_dia_object.py
yymao: Is the reason that we don't need `dm_schema_version` because now we hard code the new flux name in this reader? That is ok, and I think Approach A (which you implemented) is fine (and maybe in the future there is other name changes...). But maybe it should raise a warning or error when people try to specify `dm_schema_version < 3`?

wmwv: `instFlux` will very like remain the internal pipelines name, and there is no DIA DPDD product that we have or will generate for DM Schema Version <= 3 (I would argue that they're really actually all >=4, so I'll change that default to 4).  It is possible that it could change so keep it makes sense.

So (A) and a warning makes sense.  Thanks for the suggestion for a warning.

PR #338, opened 2019-07-06
Add hack to replace 'obshistid' with 'obs_meta.obshistid' in SQL query.

### LINKED ISSUE #337: Correct ambiguous "obshistid" error in truth variable catalog.
When querying the truth catalogs for `obshistid`, one gets an error for an ambiguous column reference because `obshistid` is shared between the "obs_meta" and "lightcurves" tables.

Fix this in `dc2_truth.py`

wmwv:
The "quantity" is "obshistid", but when we run the SQL query join
we will want to refer to it as "obs_meta.obshistid" to disambiguate.

This PR introduces a specific hack for "obshistid".  It's not pretty, but it's not terrible, and I think it's worth adding to make progress.  A fuller fix should come when we fundamentally re-do the sqlite3 access for disk access reasons.

fix #337

### DIFF
--- a/GCRCatalogs/dc2_truth.py
+++ b/GCRCatalogs/dc2_truth.py
@@ -274,9 +274,15 @@ class DC2TruthCatalogLightCurveReader(BaseGenericCatalog):
 
         for id_this in ids_needed:
             def dc2_truth_light_curve_native_quantity_getter(quantities):
+                # When 'obshistid' is needed, change it to 'obs_meta.obshistid' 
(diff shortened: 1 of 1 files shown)


wmwv:
@yymao Thanks for focusing on the real goal.  Suggestion implemented!

(and even tested)

wmwv:
@yymao Oh, sorry.  Forgot that a cosmetic rebase would still require re-approval.  Could you kindly re-approve when you have a moment.

PR #341, opened 2019-07-10
Issues/339

### LINKED ISSUE #339: Test using Parquet file Run 2.1i DR1b Object Table files with is_dpdd=true
Present a view to the Run 2.1i DR1b Object Table Parquet files
with a light wrapping, perhaps in a DMDC2 object, and `is_dpdd=true`.

The goal is to achieve significant performance benefits over the current HDF5 files.

wmwv:
This completely works and was pretty easy.  However, performance is not spectacularly different than with the trim HDF5 files.

> If this pull request is to address specific issues, please add "fix #ISSUE_NUMBER" below so that when this PR is merged, the associated issues will be automatically closed.

fix #339

### DIFF
--- a/GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
@@ -1,0 +1,9 @@
+subclass_name: dc2_dm_catalog.DC2DMCatalog
+base_dir: /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1i/dpdd/calexp-v1:coadd-dr1b-v1/object_table_summary
+schema_filename: trim_schema.yaml
+filename_pattern: 'dpdd_object_tract_\d+\.parquet$'
+description: DC2 Run 2.1i DPDD Object Table
+creators: ['DESC DC2 Team']
+pixel_scale: 0.2
+include_in_default_catalog_list: true
+is_dpdd: true
--- a/GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
@@ -2,7 +2,7 @@ subclass_name: dc2_dm_catalog.DC2DMCatalog
 base_dir: /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1i/dpdd/calexp-v1:coadd-dr1b-v1/object_table_summary
 schema_filename: dpdd_schema.yaml
 filename_pattern: 'dpdd_object_tract_\d+\.parquet$'
-description: DC2 Run 2.1i DPDD Object Table
+description: DC2 Run 2.1i DPDD Object Table - Identical to dc2_object_run2.1i_dr1b, but with underlying Parquet files instead of HDF5.
 creators: ['DESC DC2 Team']
 pixel_scale: 0.2
 include_in_default_catalog_list: true
--- a/GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
@@ -5,5 +5,4 @@ filename_pattern: 'dpdd_object_tract_\d+\.parquet$'
 description: DC2 Run 2.1i DPDD Object Table - Identical to dc2_object_run2.1i_dr1b, but with underlying Parquet files instead of HDF5.
 creators: ['DESC DC2 Team']
 pixel_scale: 0.2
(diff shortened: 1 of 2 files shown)


GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
```diff
@@ -0,0 +1,9 @@
+subclass_name: dc2_dm_catalog.DC2DMCatalog
+base_dir: /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1i/dpdd/calexp-v1:coadd-dr1b-v1/object_table_summary
+schema_filename: dpdd_schema.yaml
+filename_pattern: 'dpdd_object_tract_\d+\.parquet$'
+description: DC2 Run 2.1i DPDD Object Table
+creators: ['DESC DC2 Team']
+pixel_scale: 0.2
+include_in_default_catalog_list: true
```
yymao: Should we include this in the default catalog list? Maybe too confusing?

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_tract4850_parquet.yaml
```diff
@@ -0,0 +1,9 @@
+subclass_name: dc2_dm_catalog.DC2DMCatalog
+base_dir: /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1i/dpdd/calexp-v1:coadd-dr1b-v1/object_table_summary
+schema_filename: dpdd_schema.yaml
+filename_pattern: 'dpdd_object_tract_4850\.parquet$'
+description: DC2 Run 2.1i DPDD Object Table
+creators: ['DESC DC2 Team']
+pixel_scale: 0.2
+include_in_default_catalog_list: true
```
yymao: Same question as above.

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
```diff
@@ -0,0 +1,9 @@
...
+is_dpdd: true
```
yymao: This is not a argument for `dc2_dm_catalog.DC2DMCatalog` reader I think. It'll be ignored so it's ok, but can be removed to avoid future confusion.

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_tract4850_parquet.yaml
```diff
@@ -0,0 +1,9 @@
...
+is_dpdd: true
```
yymao: Same comment as above. 

In fact, a new and simpler way to write this whole config would be:

```yaml
based_on: dc2_object_run2.1i_dr1b_parquet
filename_pattern: 'dpdd_object_tract_4850\.parquet$'
```

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
```diff
@@ -0,0 +1,9 @@
+subclass_name: dc2_dm_catalog.DC2DMCatalog
+base_dir: /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1i/dpdd/calexp-v1:coadd-dr1b-v1/object_table_summary
+schema_filename: dpdd_schema.yaml
+filename_pattern: 'dpdd_object_tract_\d+\.parquet$'
+description: DC2 Run 2.1i DPDD Object Table
```
yymao: Maybe we should say something like this is identical to `dc2_object_run2.1i_dr1b` but just with a different underlying data format to reduce confusion?

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
wmwv: Oh, yes.  Good call.  Someday I mean to seamlessly transition.  But no need to confuse for now.  People who really want the Parquet file access often just want the Parquet files directly (Dask, Spark, e.g.,)

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_parquet.yaml
wmwv: `is_dpdd` is a kwarg for `dc2_dm_catalog.DC2DMCatalog` and does exactly what we want here.

https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/dc2_dm_catalog.py#L130

yymao: ah you're right... I misremembered!

wmwv:
Updates made.  Used and tested.

PR #344, opened 2019-07-11
h5py `Dataset.value` property deprecated

yymao:
`h5py` in DESC python environments has been updated and the use of `Dataset.value` property is now deprecated ([ref](http://docs.h5py.org/en/stable/whatsnew/2.1.html#dataset-value-property-is-now-deprecated)).

This PR replaces all the occurrence of `Dataset.value` with `Dataset[()]` as recommended. Several pylint "no member" exceptions can be removed as well.


yymao:
Thanks, @wmwv!

PR #349, opened 2019-07-16
Implement faster `__len__` method for DC2 object, DM, and composite catalogs

### LINKED ISSUE #348: Speed of __len__ for DC2DMCatalog and composite metacal
As noted here https://github.com/yymao/generic-catalog-reader/issues/26 reading the length of these catalogs is very slow.  Caching their lengths would make sense.

yymao:
This PR implements faster `__len__` method for DC2 object and DM catalogs by reading the metadata. In addition, for composite catalogs, the `__len__` method is now set to the `__len__` method of the main catalog if the format is matched. This prevents the fallback to the slow `__len__` method that was implemented in the base class.

Fixes #348


yymao:
I've tested this PR and the preference is indeed much better, not sure if it satisfies Joe's requirement though. If not, we might need to hardcoded the number in the config file.

PR #357, opened 2019-07-17
update dr1a metacal configs to improve performance

### LINKED ISSUE #355: Memory leak in dc2_object_run2.1i_dr1a_with_metacal_griz
I see a memory leak when iterating through the above catalog, which shows up for me in practice when using lots of columns.  The attached script (I had to change the ending to .txt to upload) demonstrates this using the psutil library, which reports memory usage.  The memory usage keeps increasing throughout the iteration.

The same thing also happens with dr1b in Francois' branch.

[gcr_mem_leak.txt](https://github.com/LSSTDESC/gcr-catalogs/files/3400755/gcr_mem_leak.txt)

yymao:
This PR updates dr1a metacal configs to improve performance, in particular:
1. turns off `use_cache` for the full object catalog (fix #355)
2. use `MATCHING_ORDER` instead of `ID` for matching. I believe that, even though the iteration scheme is different in the add-on, the order of records is still the same. In this case, `MATCHING_ORDER` will give better performance. @EiffL can you test and confirm this works?


EiffL:
Will check, thanks  for this !

yymao:
hmmm, this does seem to work properly, need some investigation.

yymao:
two issues identified: 
1. a bug in the base class in generic-catalog-reader. now fixed in v0.8.8
2. the record order wasn't preserved.

PR #360, opened 2019-07-17
allows customized dataset "info" (tract, visit, etc; parsed from filename) for `DC2DMCatalog`

yymao:
This PR fixes a bug that was introduced in v0.13.0 when the `DC2DMCatalog` is updated. The filenames of `DC2DMCatalog` files contain some header information (tracts, visits, etc) but it was assumed to be just tract in the previous edit. This PR fixes this issue. In particular, now customized dataset "info" are allowed. 

For convenience, two derived classes `DC2DMTractCatalog` and `DC2DMVisitCatalog` for catalogs that are split in tract/visit, respectively. 

I have not yet tested this PR at all, so I'm opening as a draft PR for now. I'm still  assigning a reviewer for early iteration.

### DIFF
--- a/GCRCatalogs/dc2_dia_object.py
+++ b/GCRCatalogs/dc2_dia_object.py
@@ -6,7 +6,7 @@ import os
 
 import numpy as np
 
-from .dc2_dm_catalog import (DC2DMCatalog,
+from .dc2_dm_catalog import (DC2DMTractCatalog,
                              convert_nanoJansky_to_mag,
                              convert_flux_err_to_mag_err,
                              create_basic_flag_mask)
@@ -14,7 +14,7 @@ from .dc2_dm_catalog import (DC2DMCatalog,
 __all__ = ['DC2DiaObjectCatalog']
 
 
-class DC2DiaObjectCatalog(DC2DMCatalog):
+class DC2DiaObjectCatalog(DC2DMTractCatalog):
     r"""DC2 DIA Object Catalog reader
 
     Parameters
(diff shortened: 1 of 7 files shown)


yymao:
This is now ready for review. Tested on `dc2_dia_object_run1.2p_test`, `dc2_dia_source_run1.2p_test`, `dc2_forced_source_run1.2p`, `dc2_object_run2.1i_dr1b_parquet`, `dc2_source_run1.2i`, and `dc2_metacal_griz_run2.1i_dr1b`.

There is a separate issue that many derived quantities defined in `dc2_dia_object_run1.2p_test` are in fact not available. This should be fixed but probably in another PR.

yymao:
What's the error message you got?

wmwv:
Ah, checking more carefully, I just can't load `dc2_object_run2.1i_dr1b_tract4850_parquet`:

```
based_on: dc2_object_run2.1i_dr1b_parquet
filename_pattern: 'dpdd_object_tract_4850\.parquet$'
description: DC2 Run 2.1i DPDD Object Table - Just Tract 4850
```

because
a) it's based on `dc2_object_run2.1i_drb1_parquet` which specifies a different base_dir than the location of the `dpdd_*parquet` files.
b) it should be marked `is_dpdd=true`

yymao:
@wmwv thanks for the clarification. I'm going to merge this and open another PR to clean up and fix the configs of DC2 Run2.1i catalogs, especially now I have DR1a and DR4 both in parquet format ready.

PR #363, opened 2019-07-18
Clean up Run 2.1i object configs

### LINKED ISSUE #347: Select different one-tract for DC2 as Tract 4850 is close enough to unsimulated DDF region
I'd like to consider selecting a different tract than 4850 for our one-tract catalogs.

The DDF area of the WFD field is _not_ included in the current WFD Run 2.1i dr1b.

The 4850 tract is close enough to the DDF that the depth in the northwest corner is notably lower and thus it is not representative of the overall WFD region..

I suggest perhaps tract 3830.  It's relatively in the center of the DC2 WFD region.

You can orient yourself to the tract arrangement on the sky with this tract list provided by @johannct 

image

And here's the object density in Run 2.1i dr1b DC2 (with a 10x subsampling)

image

### LINKED ISSUE #351: Put Run2.1i object catalogs in parquet format with tracts being the chunks
We need to put Run2.1i object catalogs (trimmed and "all columns") in parquet format with tracts being the chunks (but not patches):

There two main motivations:

1. With the size of Run2.1i, loading the Run2.1i object catalogs takes several minutes because the reader has to collect all the patches and tracts by opening all files.

2. Many add-on catalogs (photo-z, metacal) are generated in chunks of tracts. Keeping the chunk format the same is needed to enable fast composite catalog access. 

Note that because there are users who need columns that are not in DPDD, we will need to support not just the trimmed version but also the "all columns" version in parquet format.

yymao:
This PR is to clean up all the Run 2.1i object configs. 

### For the object catalogs

Keeps **only**: (* included in default list)

- `dc2_object_run2.1i_dr1a`
- `dc2_object_run2.1i_dr1b`*
- `dc2_object_run2.1i_dr4`*

These all use the parquet format files. Because of the much improved performance, now there is no distinction between trimmed and all columns, and no one-tract versions either. 

### For metacal add-on

Keeps **only**: (* included in default list)

- `dc2_object_run2.1i_dr1a_with_metacal_griz`*
- `dc2_object_run2.1i_dr1b_with_metacal_griz`*

These all use the parquet format files. Because of the much improved performance, now there is no distinction between trimmed and all columns, and no one-tract versions either.

### For photo-z add-on

Keeps **only**: (* included in default list)

- `dc2_object_run2.1i_dr1b_with_photoz`
- `dc2_object_run2.1i_dr1b_with_photoz_tract3830`*

These all use the outdated HDF5 format files. For the base object catalog, the "all columns" is used. Because the performance, we include an one-tract version, and include just the one-tract version in the default list, but people can still access the full version (just very slow...)

I've already run some tests to this PR, and I'd like to get an approve from @wmwv (for the object configs and code update), from @EiffL (for metacal configs), and from @sschmidt23 (for photo-z configs). Thanks.

This PR fixes #347 and fixes #351.


yymao:
tested on the following:
```
dc2_object_run2.1i_dr1a
dc2_object_run2.1i_dr1b
dc2_object_run2.1i_dr4
dc2_object_run2.1i_dr1a_with_metacal_griz
dc2_object_run2.1i_dr1b_with_metacal_griz
dc2_object_run2.1i_dr1b_with_photoz_tract3830
```

wmwv:
I'd like to still keep one-tract versions.  Those load in ~10 seconds.

Not having figured anything out wrong about it, tract 3830 still seems like a good choice.

yymao:
@wmwv by "load" do you mean catalog instance initiation or loading column? For the former, it takes way less than 10 seconds for parquet format even with all tracts. For the latter, users can use `native_filters`. 

But I don't have super strong opinion here. If one-tract version is still useful, I can add them back (for just dr1b and dr4?). @EiffL is one-tract version still useful for metacal?

wmwv:
Oh, sorry, I mean `get_quantities` rather than load.

You mean something like `native_filters = ['tract' == 3830]`?

wmwv:
I think it will be easier user experience to provide the one-tract version.

yymao:
OK, `dc2_object_run2.1i_dr1b_tract3830` and `dc2_object_run2.1i_dr4_tract3830` added back and tested. Confirmed with @EiffL on Slack that for metacal is not needed.

wmwv:
I think we should make the visible catalog config names just `dr1`. (which points to `dr1b`).

We can keep `dr1a` and `dr1b` as not included by default catalog configs.

I think there a 5 people who want to know about `dr1a` and they already do.  Everyone else should never have to think about this and we should just advertise `dr1`.

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_with_photoz_tract3830.yaml
```diff
@@ -0,0 +1,15 @@
+subclass_name: composite.CompositeReader
```
EiffL: Why not using the `based_on` technique here to restrict the previous catalog to one tract ?

yymao:
OK, then I propose we do:

```
dc2_object_run2.1i_dr1a
dc2_object_run2.1i_dr1b
dc2_object_run2.1i_dr1* (-> dc2_object_run2.1i_dr1b)
dc2_object_run2.1i_dr1_tract3830*
dc2_object_run2.1i_dr4*
dc2_object_run2.1i_dr4_tract3830*


dc2_object_run2.1i_dr1a_with_metacal_griz
dc2_object_run2.1i_dr1b_with_metacal_griz
dc2_object_run2.1i_dr1_with_metacal_griz* (-> dc2_object_run2.1i_dr1b_with_metacal_griz)

dc2_object_run2.1i_dr1_with_photoz
dc2_object_run2.1i_dr1_with_photoz_tract3830*
```

* included in default list
-> alias

wmwv:
> dc2_object_run2.1i_dr1_with_photoz

Should also be in the default list.

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_with_photoz_tract3830.yaml
yymao: because the tract specification needs to be pass onto each of the sub-catalogs. If we use `based_on`, we still need to write out the whole `catalogs` list, so not much save here.

yymao:
Updated, tested, ready to be re-reviewed. current default list:

```
['buzzard',
 'buzzard_high-res',
 'cosmoDC2',
 'dc2_object_run1.1p',
 'dc2_object_run1.2i',
 'dc2_object_run1.2p',
 'dc2_object_run2.1i_dr1',
 'dc2_object_run2.1i_dr1_tract3830',
 'dc2_object_run2.1i_dr1_with_metacal_griz',
 'dc2_object_run2.1i_dr1_with_photoz',
 'dc2_object_run2.1i_dr1_with_photoz_tract3830',
 'dc2_object_run2.1i_dr4',
 'dc2_object_run2.1i_dr4_tract3830',
 'dc2_truth_run1.1_static',
 'dc2_truth_run1.2_static',
 'dc2_truth_run1.2_variable_lightcurve',
 'dc2_truth_run1.2_variable_summary',
 'hsc-pdr1-xmm',
 'protoDC2',
 'protoDC2_run1.2']
```

yymao:
Thanks. Unless I hear objecction, I'll merge this Friday early morning (Paris time) in preparation for Hack Day.

sschmidt23:
Took a look as well, things work well, seems ready to merge.

PR #364, opened 2019-07-18
preparing for v0.14.0 release

yymao:
The bug fix in #360 and the performance improvement due to #363 are important enough to make another release in a short time scale. It would be useful to get this in in time for the Hack Day.


wmwv:
I agree.

We should also update the Notebook tutorials so it refers to `tract3830`
https://github.com/LSSTDESC/DC2-analysis/issues/108
This isn't a blocker, just something I should do early Friday morning as soon as v0.14 is deployed.

PR #367, opened 2019-07-19
Put super() init to end of DC2DMVisitCatalog _subclass_init.

### LINKED ISSUE #366: `DC2DiaSourceCatalog` doesn't inherit `self._visits` from `DC2DMVisitCatalog`
`DC2DiaSourceCatalog` does not get `self._visits` defined.

`DC2DMVisitCatalog` defines `self._visits` in:
```
class DC2DMVisitCatalog(DC2DMCatalog):
    _native_filter_quantities = {'visit'}
    FILE_PATTERN = r'.+_visit_\d+\.parquet$'

        self._visits = None
        if 'visit' in kwargs and 'visits' in kwargs:
            raise ValueError('Conflict options (visit and visits) defined')
        if 'visit' in kwargs:
            self._visits = [int(kwargs['visit'])]
        if 'visits' in kwargs:
            self._visits = [int(t) for t in kwargs['visits']]
```

But this appears to not get executed in `DC2DMVisitCatalog`.

wmwv:
Fixes error reading DIA Source Object catalog, e.g., `dc2_dia_source_run1.2p_test`

Moves `super` init to end of `_subclass_init` so that `self._visits` is defined when super init is run.

fix #366

### DIFF
--- a/GCRCatalogs/dc2_dm_catalog.py
+++ b/GCRCatalogs/dc2_dm_catalog.py
@@ -335,7 +335,6 @@ class DC2DMVisitCatalog(DC2DMCatalog):
     FILE_PATTERN = r'.+_visit_\d+\.parquet$'
 
     def _subclass_init(self, **kwargs):
-        super()._subclass_init(**kwargs)
         self._visits = None
         if 'visit' in kwargs and 'visits' in kwargs:
             raise ValueError('Conflict options (visit and visits) defined')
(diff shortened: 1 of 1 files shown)


wmwv:
Tested and passes on `dc2_dia_source_run1.2p_test`

wmwv:
Also confirmed that 'dc2_source_run1.2i' failed previously and works now 'dc2_source_run1.2i'.

PR #370, opened 2019-07-25
Rewrite _generate_modifiers for Parquet Object Catalog to calculate magnitudes.

### LINKED ISSUE #369: Add magnitude calculation to quantity modifier of `DC2ObjectParquetCatalog`
With the new script introduced in https://github.com/LSSTDESC/DC2-production/pull/358, we need to add magnitude calculation to the quantity modifiers in `DC2ObjectParquetCatalog` as the magnitudes would no long exist as native quantities.

wmwv:
Generate all magnitudes from flux quantitites.
Does not retain the complicated dm_schema_version, has_modelfit_mag
logic for the new Parquet Object files.  Those are presently not needed.

fix #369

### DIFF
--- a/GCRCatalogs/dc2_dm_catalog.py
+++ b/GCRCatalogs/dc2_dm_catalog.py
@@ -22,6 +22,15 @@ __all__ = ['DC2DMCatalog', 'DC2DMTractCatalog', 'DC2DMVisitCatalog']
 
 
 #pylint: disable=C0103
+def convert_flux_to_mag(flux, fluxmag0):
+    """Convert calibrated flux to AB mag.
+    """
+    flux_nJ = convert_flux_to_nanoJansky(flux, fluxmag0)
+    mag_AB = convert_nanoJansky_to_mag(flux_nJ)
+    return mag_AB
+
+
+#pylint: disable=C0103
 def convert_nanoJansky_to_mag(flux):
     """Convert calibrated nanoJansky flux to AB mag.
     """
--- a/GCRCatalogs/dc2_object.py
+++ b/GCRCatalogs/dc2_object.py
@@ -14,6 +14,7 @@ import yaml
 from GCR import BaseGenericCatalog
 
 from .dc2_dm_catalog import DC2DMTractCatalog
+from .dc2_dm_catalog import convert_flux_to_nanoJansky, convert_nanoJansky_to_mag, convert_flux_err_to_mag_err
 
 __all__ = ['DC2ObjectCatalog', 'DC2ObjectParquetCatalog']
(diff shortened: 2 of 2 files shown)


wmwv:
@yymao I have tested anything yet.  But can you take a quick look to provide suggestions about overall direction.

I decide that writing a new function was simpler than keeping track of overrides against the existing `DC2ObjectCatalog._generate_modifiers`

yymao:
The overall direction looks alright. Wouldn't `FLUXMAG0` be needed in the quantity modifiers? I don't see it though.

wmwv:
OK.  Work done and tested.

New catalogs with `{band}_FLUXMAG0` values created for previous Parquet versions that did not have that column.  Catalogs tested and worked.  New catalogs were put in same place; previous version of catalogs moved to directories to be deleted on 2019 Nov 1 to provide cushion against discovering an error.  Note that the new catalogs are in-fact smaller because of better compression in the storage of quantities in the re-written Parquet files.

On a side issue I ended up re-writing the HDF5-based catalog modifiers to used a named function instead of hard-coded zp=27 mag.  This is formally out of scope for this issue, but a reasonable time to do it.

wmwv:
@yymao Can you take a look at this PR again.  I think it's ready for merging now.

yymao:
@wmwv I am a bit nervous about the fact the new quantity modifiers would not work with old parquet catalogs at all. I understand that you have update the parquet catalog files, but this update means we won't be able to use this new reader on the old parquet catalog files. 

I'm ok with merging this but want to make this incompatibility explicit before doing so.

wmwv:
@yymao Are you also going through and doing a more detailed line-by-line review?

yymao:
I've reviewed the code. I haven't approved because I want to see if you would acknowledge that "I intentionally make this reader to be not compatibility with older parquet files". If that is not the case, then the reader should be modified to be backward compatible.

wmwv:
I certainly individually acknowledge that.
But the full answer is that "but you will never know those incompatible catalogs ever existed because we've moved them and will delete them soon", so I'm not sure how one would document that.

wmwv:
Are you worried that there are existing Parquet files outside of NERSC that we need to track down?

yymao:
That's one example (in2p3). But more generally, I consider the reader classes in GCRCatalogs also user-facing features (though here "users" are catalog creators) and hence making an incompatible API change requires a clear statement.

wmwv:
OK.  Where should I put that statement?

yymao:
This discussion suffices. I'll also mention this in the upcoming release note.

PR #376, opened 2019-08-30
Update to AGN reader to include new quantity

### LINKED ISSUE #373: Reader for AGN catalog
This issue tracks progress on developing a reader for the agn catalog prepared by Scott Daniel. The agn catalog contains a compilation of the galaxies that contain an agn component, along with the original magnitudes from cosmoDC2 and the magnitudes of their agn components. The final goal is to have a reader for a composite catalog comprising cosmoDC2 and the agn catalog as an add-on so that the composite catalog can be passed through DESCQA. After discussion with @yymao, we split the problem into 2 steps:
1) Reader for the agn catalog
2) Reader for the composite catalog

evevkovacs:
This PR updates the quantities list for the AGN reader to include a new property required for the validation tests.   fix #373

### DIFF
--- a/GCRCatalogs/agn.py
+++ b/GCRCatalogs/agn.py
@@ -47,7 +47,6 @@ class AGNCatalog(BaseGenericCatalog):
             'is_central':               'is_central',
             'ra':                       'ra',
             'redshift':                 'redshift',
-            'redshift_true':            'redshift',
         }
 
         # magnitudes
--- a/GCRCatalogs/agn.py
+++ b/GCRCatalogs/agn.py
@@ -55,7 +55,7 @@ class AGNCatalog(BaseGenericCatalog):
 
         # magnitudes
         for band in ['u', 'g', 'r', 'i', 'z', 'y']:
-            quantity_modifiers['mag_{}_noagn_lsst'.format(band)] = 'mag_{}_lsst(galaxy)'.format(band)
+            quantity_modifiers['mag_{}_noagn_lsst'.format(band)] = 'mag_{}_lsst(galaxy)'.format(band) # galaxy with no AGN
             quantity_modifiers['mag_{}_agnonly_no_agn_extinction_lsst'.format(band)] = 'mag_{}_lsst(agn)'.format(band)
             quantity_modifiers['mag_{}_agnonly_lsst'.format(band)] = 'mag_{}_lsst(agn_extincted)'.format(band)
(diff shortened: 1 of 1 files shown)


evevkovacs:
Added a few more quantities needed by latest update to AGN catalog. Reader has been tested in DESCQA See [example](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-09-03_14&catalog=AGN_mean_mags_v1.1.4)

GCRCatalogs/agn.py
```diff
@@ -38,28 +38,51 @@ def __del__(self):
...
+            quantity_modifiers['mag_{}_no_agn_extinction_lsst'.format(band)] = (_calc_mag_sum,
                                                               'mag_{}_lsst(galaxy)'.format(band),
                                                               'mag_{}_lsst(agn)'.format(band),
                                                               )
+            quantity_modifiers['mag_{}_lsst'.format(band)] = (_calc_mag_sum,
+                                                              'mag_{}_lsst(galaxy)'.format(band),
+                                                              'mag_{}_lsst(agn_extincted)'.format(band),
+                                                              )
+        # Magnitudes
+        for band in ['g', 'r', 'i']:
```
yymao: Why this derived quantity has only g,r,i bands? Can you justify (and potentially put a comment in the code for future reference)?

GCRCatalogs/agn.py
```diff
@@ -38,28 +38,51 @@ def __del__(self):
...
+            quantity_modifiers['Mag_true_{}_noagn_lsst_z0'.format(band)] = 'Mag_true_{}_lsst_z0'.format(band)
 
+        for band in ['i']:
```
yymao: Why this derived quantities has only i band? Can you justify (and potentially put a comment in the code for future reference)? Also, if in the future there will be no more band for these quantities, we should not use the for statement here.

evevkovacs: The catalog currently supplies only i-band rest-frame AGN magnitude. That may change.
Comment added.

GCRCatalogs/agn.py
evevkovacs: Only gri are available in the catalog for now. That may change.

evevkovacs: Comment added

evevkovacs:
@yymao I think this is ready to merge

PR #377, opened 2019-08-30
Config files for agn composite catalogs; reader and config files for …

evevkovacs:
…agn combined catalogs

Here are the config files and the reader for the composite agn catalogs and the combined catalogs that merge arrays and will run in descqa. See for example [here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-08-30_13&test=ApparentMagFuncTest_HSCg)


yymao:
Thanks @evevkovacs. Do we need all four configs? In particular, do we need `cosmoDC2_v1.1.4_*_composite_agn.yaml` since `cosmoDC2_v1.1.4_*_combined_agn.yaml` has the same information?

yymao:
@evevkovacs can you take a look at this (see if it's ok to remove unused configs) so that we can get this in for 0.14.2 release? Thanks!

evevkovacs:
It is ok to remove the unused config files. The reader however needs a couple of changes. There are some quantity names that need to be fixed to use the cosmoDC2 values. (eg. ra, dec, redshift etc) otherwise they will be returned as numpy masked arrays from the AGN catalog.

evevkovacs:
@yymao I would like to make one further change. I wish to suppress the overwrite of specific quantities in the second catalog (AGN_mean_mags) in favor of the quantities in the first catalog (cosmoDC2). I committed a sample config file in which I added a variable called "suppress_overwrite" which contains a list of the variables that I don't want to be overwritten. I am not sure how this needs to be implemented. Is this an option that you could provide in the base class or is it something I need to add in the subclass? I suspect this issue will arise again. Thanks

yymao:
@evevkovacs while I understand what you need, the implementation you proposed seem too complicated. It seems to me that a better way is to add option to the AGN reader so that it can expose only a subset of quantities (in particular, not exposing quantities that you do not want to overwrite the base catalog). This way there is no need to change the behavior of the composite catalog reader. What do you think?

evevkovacs:
@yymao I wanted to have the quantities available in the AGN_mean_mags catalog so it would run as a standalone catalog in DESCQA tests. But you were correct: there was a much simpler way to implement what I wanted. The new version of the reader is now committed and tested in DESCQA.

GCRCatalogs/agncombined.py
```diff
@@ -0,0 +1,47 @@
...
+                 (self._catalog_names[1], 'mag_{}_lsst(agn)'.format(band)),
+                 )
+             self._quantity_modifiers['mag_{}_noagn_lsst'.format(band)] = (
+                 self._catalog_names[0], 'mag_{}_lsst'.format(band)
+             )
+        
+        suppress_overwrite = kwargs.get('suppress_overwrite', None) #schema variables from main catalog
+        if suppress_overwrite:
+            for q in suppress_overwrite:
+                self._quantity_modifiers[q] = (self._catalog_names[0], q)
```
yymao: to avoid unintended overwrite if `suppress_overwrite` is not correctly specified 

```suggestion
                if self._quantity_modifiers.get(q) == (self._catalog_names[1], q):
                    self._quantity_modifiers[q] = (self._catalog_names[0], q)
```

GCRCatalogs/catalog_configs/cosmoDC2_v1.1.4_small_combined_agn.yaml
```diff
@@ -0,0 +1,8 @@
+subclass_name: agncombined.AGNCombinedCatalog
```
yymao: This whole config file can be written as:

```yaml
based_on: cosmoDC2_v1.1.4_small_composite_agn
suppress_overwrite: ['ra', 'redshift', 'is_central', 'halo_mass', 'galaxy_id', 'dec', 'stellar_mass']
```

to avoid repetition.

GCRCatalogs/catalog_configs/cosmoDC2_v1.1.4_image_combined_agn.yaml
```diff
@@ -0,0 +1,8 @@
+subclass_name: agncombined.AGNCombinedCatalog
```
yymao: This whole config file can be written as:

```yaml
based_on: cosmoDC2_v1.1.4_image_composite_agn
suppress_overwrite: ['ra', 'redshift', 'is_central', 'halo_mass', 'galaxy_id', 'dec', 'stellar_mass']
```

to avoid repetition.

GCRCatalogs/agncombined.py
evevkovacs: Hmm, there is no guarantee that the variable name delivered by the reader for the second catalog will match the native name of that quantity. Take for example 'redshift'. The reader for the second catalog has 'redshift' as a non-native and a native quantity.  But the native quantity could have been called 'z' instead, and I would still want to do the override, because  'redshift' is a quantity name in the second catalog that matches a quantity name in the first catalog.  Wouldn't the above test fail for this example, leading to no suppression of the overwrite?

yymao: Not sure if I follow. In the second catalog, regardless `redshift` is native or not, it will overwrite the first catalog's `redshift` (also regardless native or not). In this case, in the composite catalog, `redshift` will point to the second catalog (i.e., `(self._catalog_names[1], 'redshift')`).

evevkovacs: Sorry, not sure I understand. I thought that native quantities in the composite catalog are always tuples and hence guaranteed to be unique. The only possibility of a name collision is if the reader for catalog1 defines a quantity with the same name as the reader for catalog 2.  If there is a name collision, default behavior is to point to the catalog 2? Do I have that correct?

yymao: Yes, that's correct. Hence the code I suggested first check if the quantity is already pointing to the quantity in the second catalog (i.e., `if self._quantity_modifiers.get(q) == (self._catalog_names[1], q)`), and if so, then set it to the first catalog instead (`self._quantity_modifiers[q] = (self._catalog_names[0], q)`).

GCRCatalogs/agncombined.py
```diff
@@ -0,0 +1,47 @@
+"""
+AGN composite catalog (matched to cosmoDC2) reader
+
+This reader was designed by Yao-Yuan Mao and Eve Kovacs
+based a catalog provided by Scott Daniel.
+"""
+
+import re
```
yymao: Not used
```suggestion
```

GCRCatalogs/agncombined.py
```diff
@@ -0,0 +1,47 @@
...
+import os
```
yymao: Not used
```suggestion
```

GCRCatalogs/agncombined.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _subclass_init(self, **kwargs):
+
+        """
+        AGNComposite catalog reader, inherited from CompositeCatalog class 
+        """
+        self._catalog_names = [cat.identifier for cat in self._catalogs]
+
+        for band in 'ugriz':
+             self._quantity_modifiers['mag_{}_lsst'.format(band)] = (
```
yymao: 35-42 -- wrong indentation. remove one leading space from 35-42

GCRCatalogs/catalog_configs/cosmoDC2_v1.1.4_small_combined_agn.yaml
evevkovacs: Oops, I agreed to this but based on an earlier conversation I had already removed cosmoDC2_v1.1.4_small_composite_agn in favor of cosmoDC2_v1.1.4_small_combined_agn.  I don't have a strong preference, but I think it would be less confusing to stick with that earlier change and just use the reader for the combined catalog rather than having a composite catalog available as well.

GCRCatalogs/agncombined.py
evevkovacs: OK, so suppose the native quantity in the second catalog is called 'z',  but the reader for the second catalog renames it to 'redshift'. This collides with 'redshift' for the first catalog and we would like to prevent the overwrite. But if we use the test, doesn't it fail?  (because self._quantity_modifiers.get('redshift') == (self._catalog_names[1], 'z') and not (self._catalog_names[1], 'redshift') ).  Sorry if I have misunderstood, but I don't think we can assume that the native name and the reader name are always the same.

GCRCatalogs/catalog_configs/cosmoDC2_v1.1.4_small_combined_agn.yaml
yymao: OK, let's remove these configs that are not needed then.

GCRCatalogs/agncombined.py
yymao: In the scenario that you describe, both `redshift` and `z` will appear in the composite catalog, and `redshift` will point to `(self._catalog_names[1], 'redshift')` while `z` will point to `(self._catalog_names[1], 'z')`

evevkovacs: Ah, thank you I hadn't realized that both of the names would appear. I thought that 'z' would only appear in the tuple. In that case, I am fine with adding the check. I will modify the code with all the suggested changes and do a final test when nersc comes back up.

evevkovacs:
OK will do.

evevkovacs:
@yymao Just pushed the changes to address the comments. Did a quick reader test which worked as expected.

PR #381, opened 2019-09-18
allows multi-character filters in object catalog reader

yymao:
This PR allows multi-character filters in object catalog reader (such as `i2`).


wmwv:
Great.  Thanks for fixing this.  Good simple fix.
I've started a more general issue to clean up the bands for DC2ObjectCatalogs, #383, but we don't have to do anything about it until this comes up for any non-Parquet Object Table.

PR #384, opened 2019-09-19
Add test that fils with Boutigny's reported pyarrow 0.12.1 error.

wmwv:
The error is using a dict in index_columns and trying to create a frozenset
including a dict, which is not a hashable type.

This adds a test that fails with pyarrow version 0.12.1.
It passes with pyarrow 0.13.0 and 0.14.1.

The current CVMFS installation (or at least `w_2019_32`) of the DM Stack has pyarrow 0.12.1, if you want an easy way to get an environment to test in.

### DIFF
--- a/tests/pyarrow_dict_index.py
+++ b/tests/pyarrow_dict_index.py
@@ -1,0 +1,68 @@
+
+# 2019-09-19
+# Dominique Boutigny reported that a version of pyarrow 0.12.1
+# failed doing the following
+# while pyarrow 0.13.0 worked.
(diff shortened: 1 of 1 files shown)


yymao:
This is already tested in https://github.com/LSSTDESC/gcr-catalogs/blob/master/tests/test_catalogs.py (but not ran with travis because catalogs are not available)

wmwv:
Ah, right.  So we do test for this.  But we don't run the tests at CC.
Closing.

PR #385, opened 2019-09-19
Modifications to include duty-cycle parameter

evevkovacs:
Change reader to incorporate effect of new duty-cycle flag. Default is to combine only agn whose duty-cycle flag is on. Included configuration file for catalog with all agn included (ie duty cycle flag is ignored). Both branches of code tested with desqa: See tests with  [duty-cycle on](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-09-19_1&catalog=cosmoDC2_v1.1.4_small_combined_agn) and [duty cycle off](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-09-19_3&catalog=cosmoDC2_v1.1.4_small_combined_agn_allon)


evevkovacs:
Yes it is intended. The only analysis code that used this reader is the AGN Number density test and that has been changed accordingly. This code was really development code, so some of the quantities that included agn extinction are not going to be used in the future.

PR #388, opened 2019-09-24
Update dc2_metacal_run2.1.1i.yaml

EiffL:
I forgot that the optional arguments in the reader are actually important. As far as I know, we still need the flux correction, and we are now using the 4 bands g r i z.
@maho3 I'm sorry, I should have thought of it when reviewing your PR, can you just confirm that there was no particular reason for not including these in the first place?

### DIFF
--- a/GCRCatalogs/catalog_configs/dc2_metacal_run2.1.1i.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_metacal_run2.1.1i.yaml
@@ -1,4 +1,8 @@
 subclass_name: dc2_metacal.DC2MetacalCatalog
 base_dir: /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1.1i/dpdd/calexp-v1:coadd-v1/metacal_table_summary/final
+schema_filename: schema.yaml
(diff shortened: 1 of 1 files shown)


GCRCatalogs/catalog_configs/dc2_metacal_run2.1.1i.yaml
```diff
@@ -1,4 +1,8 @@
 subclass_name: dc2_metacal.DC2MetacalCatalog
 base_dir: /global/projecta/projectdirs/lsst/production/DC2_ImSim/Run2.1.1i/dpdd/calexp-v1:coadd-v1/metacal_table_summary/final
+schema_filename: schema.yaml
```
yymao: Do we still need this? I believe this is for hdf5 files only?

EiffL: maybe not... yes, I think you are right

EiffL: better :-) ?

Issue #389, opened 2019-09-24
Parquet DM Object catalog reader is missing some fields


EiffL:
@joezuntz pointed out that the dm object reader for the parquet files is missing the `snr_` fields, and maybe other ones as well. This is easily fixed in a couple of lines, but I wanted to raise that issue to highlight that we don't have checks that the catalogs do conform to the official schema.

yymao:
@EiffL I think this has been fixed by #390? Can you confirm and close this?

yymao:
@EiffL Can you confirm that this issue has been fixed by #390, and if so, close this issue? Thanks!

yymao:
I think one of the points that you mentioned that is certainly not in #390 is a new test... but we should open a new issue for that.

PR #390, opened 2019-09-24
Adds snr field in Parquet DM object catalog

EiffL:
This PR aims to address #389 
@joezuntz Can you check that all the columns you need are now defined here?


yymao:
@EiffL while you are at this, can you compare the modifiers in the parquet reader with the original reader (https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/dc2_object.py#L335)?

yymao:
In addition, @EiffL, can you also rebase to or merge with master? That should fix the linter error.

yymao:
I am merging this because the edit was fine, but I'm not closing the issue #389 to make sure all needed fields have been implemented in the new object reader.

PR #391, opened 2019-09-25
Fix the definition of mag_{}_cModel for HDF5 DC2ObjectCatalog

wmwv:
This PR fixes a bug that was introduced (by me) in cc8bbc14 as part of #369.

There was a bug where `mag_{}_cmodel` was defined based on `mag_{}_cModel`

This PR fixes it to be `{}_modelfit_CModel_{}.format(band, FLUX)` as it should have been (and should be).

This should have only affected Run 1.1p and HSC catalogs.
because it was triggered if the modelfit mags were not already calculated
and only for the default (HDF5) catalogs.

The Parquet class was fine.


wmwv:
@yymao OK.  Linter now happy.  Ready for re-approval.

PR #393, opened 2019-10-08
Reader for matched-table catalog and config files for composite add-o…

evevkovacs:
@yymao  @fjaviersanchez  Here is a reader for the new table, created by Javi to find objects in the dc2 object catalog that have matches in cosmoDC2. This reader just reads in the fits file and generates some useful quantities. I should be able to run some descqa tests on this table/mini-catalog.  I have requested both of you as reviwers. The next steps are 1)to make it an add-on to the dc2-object catalog, since people might want to check other quantities and 2) to make cosmoDC2 an add-on so other tests can be run.  I have also pushed my first attempt at the config file for using the new reader as part of a composite catalog. It doesn't crash, but it is not giving me the result that I expected.  The object catalog and the matched table both have 7548496 rows, but when I read in the composite catalog (which is matched on objectId) I get 78649132 rows, so I have not understood something about how composite catalogs work. Could one of you take possibly take a look and see if you can spot the problem?


yymao:
Thanks @evevkovacs @fjaviersanchez --- about the questions @evevkovacs raised:

> 1) to make it an add-on to the dc2-object catalog

This can work by either matching the `object_id` column like you did, or, if the object catalog and the matched table have the same row order, use the `MATCHING_ORDER` method. (It seems that the matched table is a single fits file, so we can't use the`MATCHING_FORMAT` method, since it requires the add-on catalog to have exactly the same order.)

In the composite config, it appears that you are matching to `dc2_object_run2.1i_dr1` not `dc2_object_run2.1.1i`. This is probably what causes the issue? 

> 2) to make cosmoDC2 an add-on so other tests can be run

I now realized that this is a bit beyond the original design of the composite catalog, which was to add add-on columns to a main catalog by matching to a single column of the main catalog. Here, the cosmoDC2 needs to be matched to a different column (`truth_id`). 

While this possible, the implementation would look quite ugly unless we go change the underlying composite catalog code. But before we further expand GCR's functionality, we should really ask if we should instead explore using database instead.

GCRCatalogs/catalog_configs/dc2_object_run2.1.1i_matched_addon.yaml
```diff
@@ -0,0 +1,8 @@
+subclass_name: composite.CompositeReader
+description: DC2 Run 2.1.1i Object Catalog Matched Table addon catalog
+creators: ['Javi Sanchez', 'Eve Kovacs']
+catalogs:
+  - catalog_name: dc2_object_run2.1i_dr1
```
yymao: should this be
```suggestion
  - catalog_name: dc2_object_run2.1.1i
```
?

GCRCatalogs/dc2_matched_table.py
```diff
@@ -0,0 +1,70 @@
...
+            self._matched_galaxy_count = np.count_nonzero(self._galaxy_match_mask)
+            self._star_match_mask = (hdul[1].data[is_star]) & (hdul[1].data[match_flag])
+            self._matched_star_count = np.count_nonzero(self._star_match_mask)
+            
+        # modify native quantities
+        self._quantity_modifiers = {}
+        modified_quantity_list = [c for c in cols if not is_star in c and not match_flag in c]
+        for q in modified_quantity_list:
+            self._quantity_modifiers[q + '_galaxy'] =  (lambda x: x[self._galaxy_match_mask], q)
+            self._quantity_modifiers[q + '_star'] =  (lambda x: x[self._star_match_mask], q)
```
yymao: Line 48-50

So these new columns will have different number of rows than other columns. Is this intended? While this is allowed in GCR, it may cause confusion to users. Also, this may not work as intended when it becomes an add-on catalog because the composite reader does assume that all columns have the same row number.

evevkovacs: Ah yes, thanks for spotting that typo! 
@yymao Re : different number of rows issue. Yes, I have been thinking about this point, especially, as you say, when it becomes an add-on. My proposed solution is to create these as numpy masked arrays, so that the length of the vectors is the same for all columns. What do you think?

GCRCatalogs/dc2_matched_table.py
```diff
@@ -5,66 +5,147 @@
 based on a matching table provided by Javier Sanchez .
```
johannct: Is the class meant to deal with any matched catalog, or only the dpdd to truth matching that seems to be targeted here? It may be useful to expand the docstring to provide more information to the user as to how to use this class.

GCRCatalogs/catalog_configs/dc2_object_run2.1.1i_matched.yaml
```diff
@@ -1,5 +1,7 @@
 subclass_name: dc2_matched_table.DC2MatchedTable
-filename: '/global/projecta/projectdirs/lsst/groups/SSim/DC2/matched_ids_run2.1.1i_dpdd.fits.gz'
```
johannct: I propose that we use this pull request to fix the NERSC-specific paths problem. @yymao @boutigny @heather999 @wmwv

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_matched.yaml
```diff
@@ -0,0 +1,12 @@
...
+table_filename_template: matched_ids_run{}_{}_{{}}.fits.gz
+data_release: dr1b_dpdd
+version: '2.1i'
+truth_version: 'cosmoDC2_v1.1.4'
+object_id: 'objectId'
+truth_id: 'truthId'
+match_flag: 'is_matched'
+is_star: 'is_star'
+creator: 'Javier Sanchez'
+description: 'This is the table of matches between the object catalog and the extragalactic catalog'
```
johannct: should we specify what kind of matching was applied, and a link to the code which did the matching. Provenance seems not easy to track as it stands. Maybe the matching code "as run" could make it to a separate directory in GCRCatalogs?

fjaviersanchez: This is the code used to match the inputs and outputs: https://github.com/LSSTDESC/CatalogMatcher/blob/master/CatalogMatcher/match.py#L44

In particular I used `npix=3` and `max_deltamag=1.0`

fjaviersanchez: The complete script lives [here](https://gist.github.com/fjaviersanchez/787bb5cd6b598226174e1cd9661465ca)

johannct: thanks @fjaviersanchez ! Ideally for provenance we would have a tag of the script in the github repo you are using. Maybe a static reference to a perennial gist is also ok at this point, but I like less the idea.

GCRCatalogs/dc2_matched_table.py
evevkovacs: @johannct  This class is meant to deal with the matched tables for DC2 object catalogs provided by Javi. I am not aware of any other possible application at present. I have expanded the doc string to provide more description.

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_matched.yaml
evevkovacs: I added these locations to the doc strings in the code.

fjaviersanchez: I can put that code in the DC2-production repo

johannct: I am not sure what is best..... Other opinions welcome : @heather999 @yymao @wmwv

evevkovacs:
@yymao @fjaviersanchez  The code is tested and works but fails the travis check with a warning I don't fully understand. The offending line is 127.  The warning goes away if I add the argument  "handle" to the call signature for the native_quantity_getter, but then the code crashes. Can one of you please take a look and see if you know how to make travis happy. Thanks

GCRCatalogs/catalog_configs/dc2_object_run2.1i_dr1b_matched.yaml
heather999: I think the code belongs in a LSSTDESC repo which is tagged.  Unfortunately, we do not do a good job of tagging DC2-production, but I'm open to use that repo for this purpose... and then we should take a quick look at the outstanding PRs and make a DC2-production release tag.

evevkovacs:
@fjaviersanchez @yymao  I made one last attempt to fix the travis issue but it also failed with the same error. The offending line is now 135. Thanks

evevkovacs:
@fjaviersanchez  @yymao  I made one last attempt to fix the travis issue and this time it passed! So as soon as one of you approves, I can merge this code. Thanks a lot! I'd like to merge so that other people can run descqa tests if they want.

heather999:
@fjaviersanchez Please go ahead and add the matching scripts to DC2-production via a PR and then we should organize a tag of DC2-production.

PR #396, opened 2019-10-10
Update Object Table schema description psf_fwhm to arcsec.

wmwv:
The README Schema incorrectly listed this as being in pixels.  It's in arcsec.


wmwv:
Meta yaml files updated.
The {DIA,Forced,}Source Tables actually only have psf_fwhm_pixels, which is in, well, pixels.  So the fix there was to change the column name of the quantity.

wmwv:
@yymao ping

PR #402, opened 2019-11-05
U/sschmidt23/mockpdfreader

sschmidt23:
This pull request adds a composite reader for the subset of galaxies where redshift PDFs have been computed.  The base catalogs are  photoz_magerr_10y_estimate_image (cosmoDC2_v1.1.4_image truth catalog galaxies with mock 10-year depth errors added) and a new photoz_magerr_with_pdfs catalog.  

A quick demo notebook has been added in the `examples` folder, but this is mostly just a quick demo for Chris and Markus as they review, it should most likely not be included in the final merge, as it is not as useful to others and is not very well documented.


sschmidt23:
I noticed a mistake where I forgot to define a decorated function so that photoz_pdf_bin_centers was actually accessible for the catalog, I pushed that update, am now waiting for the CI check to re-run.

Issue #345, opened 2019-07-12
Homogenize the config keyword that indicates file location


yymao:
In an effort to make `gcr-catalogs` more easily installable outside NERSC (#271), we need to allow the root directory of file location to be configurable. The first step is to homogenize the config keyword that indicates file location. 

Proposal: 

Each catalog config that uses on-disk files must use `base_dir` to indicate the file location. Additional custom keywords can be used to indicate file names or file name patterns, but the catalog shall not access any file that does not sit in `base_dir` or its subdirectories.

wmwv:
I agree with the spirit of this.  Which catalog configs currently don't obey this proposal?

But I would like this to be a requirement for the catalog config files themselves, but not for the general behavior of specifying custom things in `config_overwrite` arguments when loading the catalog.

yymao:
Superseded by #448

PR #416, opened 2020-02-12
Implement ROOT_DIR for GCR.  Needs more testing

### LINKED ISSUE #346: Make root directory (common prefix of all `base_dir`s) customizable during setup
In an effort to make gcr-catalogs more easily installable outside NERSC (#271), we need to allow the root directory (the common prefix of all `base_dir`s, after #345 is done) to be configurable during package setup time. 

There are a few ways to do this: 

1. Require `base_dir` to always be a relative path. At run time, the reader prepends root directory path (which is set during package installation) to `base_dir`.

2.  Require `base_dir` to be in the form of `<ROOT_DIR>/relative/path`. At run time, the reader replaces `<ROOT_DIR>` with the actual root directory path (which is set during package installation). 

3.  Require `base_dir` to be in the form of `/global/projecta/projectdirs/lsst/production/catalogs/relative/path`. At run time, the reader replaces `/global/projecta/projectdirs/lsst/production/catalogs` with the actual root directory path (which is set during package installation). 

(1) appears more elegant and less hack-y. (2) is more explicit to the user (i.e., the user would know what will happen). (3) has the benefit that for NERSC users, they can easily find the files without needing to find out what `ROOT_DIR` is. 

Comments on this are welcome.

JoanneBogart:
Meant to fix #346


GCRCatalogs/register.py
```diff
@@ -33,6 +49,16 @@ def load_yaml(yaml_file):
...
 
+def _resolve_dict(d):
+    for (k,v) in d.items():
+        if k in _PATH_LIKE:
+            d[k] = os.path.join(str(_ROOT_DIR), str(v))
+        elif k == 'catalogs':
+            # list of items, each of which is catalog spec.
+            for c in v:
+                _resolve_dict(c)
+    return d
```
yymao:
```suggestion
def _resolve_dict(d):
    for k, v in d.items():
        if isinstance(v, dict):
            d[k] = _resolve_dict(v)
        elif k in _PATH_LIKE and is_string_like(v):
            d[k] = os.path.join(ROOT_DIR, v)
    return d
```

GCRCatalogs/register.py
```diff
@@ -12,6 +13,21 @@
...
+def _get_rootdir():
+    with open(os.path.join(os.path.dirname(__file__),'site_config/site_rootdir.yaml')) as f:
+        d = yaml.safe_load(f)
+        host = socket.getfqdn()
+        for (k,v) in d.items():
+            if k in host:
+                return v
+        return None
+
+_ROOT_DIR = _get_rootdir()
```
yymao:
```suggestion
_DEFAULT_ROOT_DIR = _get_rootdir()
ROOT_DIR = _DEFAULT_ROOT_DIR

def reset_root_dir():
	ROOT_DIR = _DEFAULT_ROOT_DIR
```

GCRCatalogs/register.py
```diff
@@ -12,6 +13,21 @@
...
+              'catalog_root_dir', 'header_file', 'repo')
+
+def _get_rootdir():
+    with open(os.path.join(os.path.dirname(__file__),'site_config/site_rootdir.yaml')) as f:
+        d = yaml.safe_load(f)
+        host = socket.getfqdn()
+        for (k,v) in d.items():
+            if k in host:
+                return v
+        return None
```
yymao: Make `_get_host_info` a separate function (in case in the future we want to change how to obtain host info)

If the host cannot be identified, use current directory as `ROOT_DIR`. 

```suggestion
def _get_host_info():
	return socket.getfqdn()

def _get_root_dir():
	host = _get_host_info()
    with open(os.path.join(os.path.dirname(__file__), 'site_config', 'site_rootdir.yaml')) as f:
        d = yaml.safe_load(f)
    for k, v in d.items():
        if k in host:
            return v
    return os.curdir
```

GCRCatalogs/register.py
```diff
@@ -55,7 +81,7 @@ def content(self):
             self._content = load_yaml_local(self.path)
         return self._content
 
-
+
```
yymao:
```suggestion

```

heather999:
@JoanneBogart @yymao I'd like to try this out in a docker container - is now a good time?  If I were to do that - can you provide a brief set of instructions to do such an install with the new ROOT_DIR?

GCRCatalogs/register.py
JoanneBogart: A couple comments:
1. I don't think this will handle the 'catalogs' case properly.  The structure of 'catalogs' is a list of dicts.  In a later version, not yet pushed, I made it a bit more general with a declaration 
`_DICT_LIST = ('catalogs')`
More generally, I'm somewhat torn about how general to make this code.  I considered just passing all lists and dicts through the code, but some of the config files are quite long, containing things which really don't need to be parsed, so I've compromised with tuples of known keywords of interest.
2. Originally I didn't cast _ROOT_DIR and v to str and one of the checks failed when I pushed to GitHub.  is_string_like seems obscure (the only place I found it so far is https://networkx.github.io/documentation/stable/reference/generated/networkx.utils.misc.is_string_like.html ).  I would rather check by using calls to isinstance.

GCRCatalogs/register.py
JoanneBogart: I like the idea of separating the root dir used from the default.  I'm not sure about exposing the variable, though.  If someone needs to override I'd rather they do it through a function call.

GCRCatalogs/register.py
JoanneBogart: Yes, originally I had it as a separate function.  I'll put it back.

I don't think there is a sensible fallback if host cannot be identified.  There should be a way for the user to explicitly set it, though, and an exception raised if someone tries a load when it isn't set (or at least when they try a load of something involving a relative path).

JoanneBogart:
@heather999  I don't think there is anything special you need to do.  I tried running the bit of code which determines site from within a shifter image and it worked; everything else follows from that.
   When GCRCatalogs is imported, that code runs to determine if site is nersc and, if so, sets the default root dir suitably.    The existing config files all have absolute paths, which will continue to work with this code.  To see it in action you should do a load of one of the test configs with name ending in _rel.

GCRCatalogs/register.py
yymao: We can raise a warning if `_DEFAULT_ROOT_DIR` cannot be set but it should not raise an exception. People may want to use GCRCatalogs in ways that do not require `ROOT_DIR`.

GCRCatalogs/register.py
yymao: Since there is really no real protected properties in Python, we do depend on  users being reasonable users. Hence, allowing ROOT_DIR to be a regular variable is a clear indication of that the users have control over it and they should be careful with it.

GCRCatalogs/register.py
yymao: Sorry, `is_string_like` is already in `GCRCatalogs.utils`. 

You're right about the list issue. Let me think about it.

yymao: How about this:

```python
def _prepend_roor_dir(value, key=None):
    if isinstance(value, dict):
        return {k: _prepend_roor_dir(v, k) for k, v in value.items()}
        
    if isinstance(value, list):
        return [_prepend_roor_dir(v, key) for v in value]
    
    if key in _PATH_LIKE and is_string_like(value):
       return os.path.join(ROOT_DIR, value)
   
    return value
```

JoanneBogart: As I said - yes, you can do something like this.  But it means parsing a lot more of the content of config files than will ever be necessary.  Maybe the parsing is fast and it doesn't matter.

GCRCatalogs/register.py
JoanneBogart: Code should not depend on inputs being reasonable or users being careful.   GCR should check that it is reasonable.

GCRCatalogs/register.py
JoanneBogart: If the path is absolute ROOT_DIR isn't needed.   Otherwise, I still think there is no sensible default. People may have been in the habit of depending on relative paths this way; that doesn't mean we have to continue allowing it. Having to make an extra call to set ROOT_DIR is not more trouble than having to cd to a particular place, and there is at least some record of what was done.

GCRCatalogs/register.py
yymao: The parsing has been done by yaml.safe_load anyway. Walking the dict is not costly. I think the `elif k == 'catalogs'` is a bit opaque as people may not know why it needs to be a special case. If we are going that route we need to add a bit more comment.

GCRCatalogs/register.py
yymao: My point is that users can always edit `_DEFAULT_ROOT_DIR` or `_ROOT_DIR` even if there exists a function called `set_root_dir`. So I would prefer having fewer ways to do the same thing. We also need a `get_root_dir` if we use `_ROOT_DIR` instead of `ROOT_DIR`.

GCRCatalogs/register.py
yymao: I am not sure if I follow. My point is that one should be able to import `GCRCatalogs` outside CC/NERSC and without the need to set `ROOT_DIR`. For example, a user may want to just load a reader for other purposes, and there is no need for the config files at all. 

So my request is that if there's no `ROOT_DIR` at all, `GCRCatalogs` should not raise an exception (but a warning is ok).

JoanneBogart: I was thinking of raising an exception only when someone tried to load a file without absolute path, not at import time.   If we go with explicitly including $ROOT_DIR in paths, then it would only be when loading a file for which $ROOT_DIR is part of the path.

I'd also be in favor of a warning at import time.

GCRCatalogs/register.py
JoanneBogart: Yes, people can modify variables, but, outside of personal scripts and interactive work, it's not a good thing.   I would rather provide a routine which can at least check that the value corresponds to an actual directory.    I don't mind also providing a one-line get routine.

GCRCatalogs/register.py
yymao: At catalog load time, usually the reader does raise an exception if there is no file at the given path. 

The code we are working here loads the config, modify it as needed, and then pass it to the reader (and the reader loads the catalog). So the question is whether an exception is needed during this process. Once the config is passed to the reader, I would prefer that we let each reader to handle what exception to raise.

GCRCatalogs/register.py
JoanneBogart: Oops!  I've eliminated the extra spaces.

GCRCatalogs/register.py
yymao: OK, so now there are two distinct questions: (1) the API for setting/getting `ROOT_DIR` and (2) whether `ROOT_DIR` should be checked to a valid path when set. 

What I have been discussing is (1). I prefer `GCRCatalogs.ROOT_DIR = ...` to `GCRCatalogs.set_root_dir(...)` for reasons I mentioned. But this is really just a question of API.

Question (2) is different --- I actually think it should _not_ be checked at set time. It could be checked later (as we discussed later). The reasons I think it shouldn't be checked at set time are:
- (a) `ROOT_DIR` is not used until the catalog is about to be loaded, and generally I don't want to prohibit the users to do whatever they want unless it is really going to cause an issue.
- (b) I can imagine cases where `ROOT_DIR`is set before the directory exist, but later at catalog load time, the directory becomes ready. 
- (c) I think `GCRCatalogs.set_root_dir(..)` should have the same behavior as `GCRCatalogs._ROOT_DIR = ...`. Otherwise users can just skip the API we design and opt to use `GCRCatalogs._ROOT_DIR`.

GCRCatalogs/register.py
JoanneBogart: My current draft  has
`_DICT_LIST = ('catalogs')`
and then a clause
```
        elif k in _DICT_LIST:
            # list of dicts, each of which may have _PATH_LIKE keys
            for c in v:
                _resolve_dict(c)
```
In choosing between this kind of thing and a more general handler like the one you propose there are two things to consider:

- is there a noticeable performance penalty?  The initial parsing is done by yaml.safe_load. GCR code would still have to look at every keyword with this approach, but maybe that's not a problem
- should we keep a tighter rein on the form of the config files?  
For example, someone could add a new keyword with a dict as value and one of the dict keys could happen to be in the _PATH_LIKE keys, but maybe not intended to be handled as those keys are at the top level or with a `catalogs` block.  

One way to avoid confusion would be to require $ROOT_DIR before any value which should be transformed.  If we go with the general handler, I think we need an explicit signal of some kind.

JoanneBogart: On the other point about `is_string_like`  - since it's in the package it's reasonable to use it, but I suspect the check will still complain about  `os.path.join` line unless I cast the value, whereas if I use `isinstance(value, str)` it probably won't.   What is the advantage of `is_string_like`?

JoanneBogart:
test_reader_modules.py is now failing for the "right" reason:  site is not found and my additional test configs require it since they have relative paths.
We might have to redesign this test.

GCRCatalogs/register.py
yymao: `is_string_like` is mostly for Python 2-3 compatibility, so it's not as relevant now TBH. I am using it merely because there's already a function in `utils`. 

It's ok to turn off some linter checks by comments.

yymao: On a different note, I suggest that we use `_PATH_LIKE_KEYS` or `_PATH_DEFINING_KEYS` instead just `_PATH_LIKE`.

yymao: The reason I don't quite like the proposed solution of `_DICT_LIST = ('catalogs',)` is that I find it quite obscure --- I think it is difficult for new devs to figure out why this particular keyword `catalogs` is in `_DICT_LIST`. The reason that I proposed a more generic solution is that it should brings up fewer questions IMO. It is also not clear what happen in a new reader use `catalogs` to specify file path.  

The keywords in config files are not controlled or checked (with the exception of `subclass_name`, `alias`, and `based_on`) in `register.py`. They are just passed to each reader. Hence, the types that keywords corresponds to are determined by the readers, not by `register.py`. That's why this `_DICT_LIST` check makes me uncomfortable, as it creates more cross-dependencies between `register.py` and readers. 

If we implement a set of standard keywords for the config files (related to #345), then I'd be more ok with more filtering on keywords here.

GCRCatalogs/register.py
JoanneBogart: My preference is to prevent users from stepping on their own toes where possible, when it doesn't entail any real loss of functionality. And also to catch errors earlier rather than later.  
 For a), something which is being done with no clear purpose in mind which could lead to failure later is better not done at all. For b) I don't think it's unreasonable to ask users to defer setting _ROOT_DIR until after the directory is created.
Concerning c), the whole reason for providing a function is that setting GCRCatalogs._ROOT_DIR directly really shouldn't be allowed.  It's unfortunate that Python **does** allow it and therefore some people may do it in inappropriate circumstances.  What we can do is provide a better interface, which is the one that is publicized.

yymao: Right, so I think we do disagree on some fundamental principles here. For example, I think we should use example and documentation rather than hard-coded exceptions to prevent users from stepping on their own toes. Another difference is that I think while we can catch errors early, we should not raise exception until it is really necessary. 

So, generally I am ok to defer to the author's preference, but I also want to make sure philosophy is not completely inconsistent with other parts of the code.

I am ok with get/set API, but still would not want to raise exception if when `ROOT_DIR` is set to a non-existent path. Raising a warning is fine.

GCRCatalogs/register.py
JoanneBogart: For individual testing and debugging the approach of waiting to raise an exception until it's absolutely necessary has its place, though in this case I would argue that asking the user to either fix the config or make one function call to establish an acceptable root_dir is not much of a burden.   For production running - and GCRCatalogs is used extensively in such circumstances - it's better to find out about potential problems early.

yymao: I fully agree with the distinction about interactive vs. production use. I would argue that `GCRCatalogs` is meant for interactive use, but I also have to agree that it has been used in production too.

But I am still not sure why `set_root_dir` need to raise exception. I think for production one won't need to use `set_root_dir`. If `set_root_dir` will raise exception when the path is not valid, should `config_overwrite` also raise exception when people overwrite the path with a invalid path? I would not want to go down that route.

JoanneBogart: I agree set_root_dir is not intended for production use and it's very unlikely it would be used in those circumstances, so a warning is adequate.  I'll make this change.

config_overwrite should do the same level of checking as is done for the original path, whatever that level is. 
 I would prefer to do whatever checking can be done easily in a uniform way in register rather than duplicating the logic in many readers.  That would include looking at paths (if an invalid path has been specified, chances are good it's a typo), but, for this project, I don't think it's appropriate to make wholesale changes, especially when we have philosophical differences.

GCRCatalogs/register.py
JoanneBogart: A couple things -
Which new devs do you mean?  I assume most people using using GCR don't look at register.py other than perhaps to see what services are available. People creating composite catalogs have to know about the 'catalogs' keyword.  Does anyone else?

We've already started down the road of greater control of keywords because of the special handling of the members of _PATH_DEFINING_KEYS.   `catalogs` also demands special handling because its dict may have keywords which are themselves in _PATH_DEFINING_KEYS _and are meant to be handled the same way_ as far as root dir is concerned.   If someone invents a new structure for some other config and that also happens to have a `filepath:`, `base_dir:`, etc. which is neither at the top-level or inside a `catalogs` block, there are two ways to approach it:

1. Insist that all occurrences of these keywords, no matter what the context, are treated the same; that is, always prepend root dir and don't allow absolute path in tagged versions. 
2. Only treat these keywords that way in circumstances where that is definitely the intent. Currently that means either at the top level or within a `catalogs` block.  If someone wants to invent a new block which has behavior like the `catalogs` block, they make a pull request so it can be added to _DICT_LIST

In either case we're imposing standards. I think the second approach gives developers a bit more freedom, but the first is also ok as long as everyone understands that all these keywords are behaving like reserved words, no matter what the context.

JoanneBogart:
@yymao  How do you feel about requiring some string (like ${ROOT_DIR}) when root dir is supposed to be prepended?   Earlier I believe you said it wasn't necessary and I thought the same, but now I'm not so sure.  The advantage is that it would allow people to use a relative path during development without interfering with access to production catalogs.  I can imagine this would be useful for someone developing a composite catalog or developing a new catalog of any sort for an application which also needs access to production catalogs in the usual place.

GCRCatalogs/register.py
JoanneBogart: It's now changed to warning.  Can we resolve this issue?

GCRCatalogs/register.py
JoanneBogart: Concerning config_overwrite: this strikes me as a place where it *is* reasonable for someone to specify an absolute path or a relative path which is relative to current dir rather than the root dir.   But it's also possible they'd want to specify something relative to root dir. This is another argument for starting the value of `base_dir` (or values of other path-like keys) with a special string when it's meant to be relative to root dir.

yymao: I'll reply to this below and resolving this thread since the original code is outdated.

yymao:
@JoanneBogart thanks for the comments/revision. Here's some general thoughts:

1. I still don't think `register.py` should take on the job of raising _exception_ when `_ROOT_DIR` does not exist. Each reader already makes sure the files exist and raises exception when one try to load the catalog the path is wrong. I am not convinced that there is a need to add one more layer that checks only parts of the path. 

2. I don't quite like the `${ROOT_DIR}` idea because we then mix in a new syntax (bash). It also lets users to wonder if they can put `${ROOT_DIR}` in the middle of the path or use other environment variables (which they can't). 

Let's review a few use cases:

- Yaml config using absolute path: Should not exist for the long term, but for temporary config files it should work as expected (i.e. load from the absolute path)
- Yaml config using relative path: `_ROOT_DIR` will be prepended to paths.
- Yaml config using relative path but someone wants to use a different root dir (e.g. current dir) for temporary reasons: Use `set_root_dir()` to customize `_ROOT_DIR`

- Using `config_overwrite` at run time with absolute path: load from the absolute path.
- Using `config_overwrite` at run time with relative path: the relative path will be prepended with `_ROOT_DIR`. 
- Using `config_overwrite` at run time with relative path but want to actually use relative path: the user can either use `set_root_dir` to customize `_ROOT_DIR`, or to simply use `os.path.abspath('relative/path')`

GCRCatalogs/register.py
```diff
@@ -33,6 +76,21 @@ def load_yaml(yaml_file):
...
+            if v.startswith('/'):
+                continue
+            elif _ROOT_DIR is None:
+                raise Exception('GCRCatalogs root dir not set')
+            else:
+                d[k] = os.path.join(_ROOT_DIR, v)
+        elif k in _DICT_LIST:
+            # list of dicts, each of which may have _PATH_LIKE keywords
+            for c in v:
+                _resolve_dict(c)
```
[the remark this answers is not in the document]
JoanneBogart: I believe one of the test cases (the one with a composite catalog) shows that this does modify the input dictionary.  All that needs to be changed is the value of certain keywords.  It is not necessary to copy the whole dictionary.

GCRCatalogs/register.py
```diff
@@ -3,24 +3,67 @@
...
+_DICT_LIST = ('catalogs')
+
+def _get_site_info():
+    '''
+    Return a string which, when executing at a recognized site with 
+    well-known name, will include the name for that site
+    '''
+    return socket.getfqdn()
+
+def _get_default_rootdir():
```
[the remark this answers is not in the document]
JoanneBogart: I have no objection, but there is still the possibility that the user won't have set it.

GCRCatalogs/register.py
```diff
@@ -3,24 +3,67 @@
...
+    with open(os.path.join(os.path.dirname(__file__),'site_config/site_rootdir.yaml')) as f:
+        d = yaml.safe_load(f)
+        site_info = _get_site_info()
+        for (k,v) in d.items():
+            if k in site_info:
+                return v
+        warnings.warn('No default root dir found')
+        return None
```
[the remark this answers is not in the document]
JoanneBogart: This is the same (still unresolved) issue about using current directory as default, which is also connected with explicit - or not -  indication that rootdir should be prepended.

JoanneBogart:
Concerning your general comments above:

> I still don't think register.py should take on the job of raising exception when _ROOT_DIR does not exist. Each reader already makes sure the files exist and raises exception when one try to load the catalog the path is wrong. I am not convinced that there is a need to add one more layer that checks only parts of the path.

My thinking is that _ROOT_DIR is the responsibility of register.py.   If it doesn't have a sensible value *and* there is an attempt to use it (where here "use it" means directly or indirectly invoking `resolve_config`), that is an error.  I prefer looking for that error in one place rather than depending on all maintainers of readers to do the right thing.   Of course if someone tries to actually  open a file using a bad path they'll get some sort of error message, but it will not pinpoint the source of the problem if that problem really had to do with _ROOT_DIR.

I agree that using a string that looks like a shell environment variable is not a good idea because it's subject to misinterpretation.  I need to come up with a better alternative.   But I do think the concept is valuable so that a) the paths in config files  mean what they appear to mean (relative path is always relative to current directory) and b) users can refer to catalogs which are stable and in production as well as those under development in the same process without having to resort to absolute paths for the development ones.  You suggested a way to handle this with `config_overwrite` in your last bullet.  That would work, but as a developer I wouldn't want to have to change my code to accomplish this (and have to remember to change it back when the catalog becomes production). It's much easier to keep track of it in the config file, especially since any relative path in the config would have to change before it could become part of a tag.

yymao:
Again, my fundamental request is that `register.py` should not raise exception regarding the path being invalid. Unless we are changing the fundamental philosophy here, I really want to maintain the idea that we allow the readers to be in charge of doing whatever they want. The job of `register.py` is to pass the config files to the reader, and it's not it's job to say if the path or part of the path is not valid. 

As for supporting relative paths, I feel like we are creating imaginary use cases here. I am not sure we've ever encountered such a need. And if it we do, I think it is rare enough that it is not unreasonable to ask them to specify absolute path during development stage. 

If you think it is important that we must support such use case, that's fine by me, and my request would be not to use a syntax that is suggestive that users can do more than that can (which I feel it's really difficult to avoid).

GCRCatalogs/register.py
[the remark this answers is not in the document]
yymao: you're right. We should make it clear that `c` is a dictionary so that this is clear.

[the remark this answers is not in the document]
JoanneBogart: I think it's clear as it stands.  There is a comment saying k is a list of dicts and the routine is called _resolve_dict.    I'd like to resolve this conversation.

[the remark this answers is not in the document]
yymao: That's fine. 

Please add the checks (`isinstance(v, list) and all(isinstance(c, dict) for c in v)`) to avoid keyword collision (i.e., some other readers use `catalogs` for a different use).  Thanks,

JoanneBogart:
I think it will be less confusing if something that looks like a relative path has its usual meaning: relative to current directory.
How about using `^` to represent root dir?   This character has no special meaning for YAML but is very unlikely to appear in a filepath.  The three cases would then be

* /an/absolute/path
*  relative/to/current/directory
* ^/relative/to/rootdir

yymao:
That's fine. I feel "relative to current directory" use case would be much more rare than the "relative to rootdir" use case? How about using "./" for the former? i.e., 

- /an/absolute/path
- relative/to/rootdir
- ./relative/to/current/directory

JoanneBogart:
I agree that relative-to-rootdir is going to be much more common than relative-to-current-dir, but I think clarity is more important than saving people typing an extra character.  At the command line `relative`   and   `./relative` give you the same thing so  people's expectation would be that they translate to the same thing in the config file as well.   The one thing that *can't* be represented in any standard fashion at the command line is the relative-to-rootdir case, so I think that's the one that needs something special.  
I'd like to stick with `^` for relative-to-rootdir if that's ok.

yymao:
It's ok. But I think it'll hard to guess that `^/` means by just looking at the config files. 

I anticipate this is going to cause some confusion (in fact, any of these proposals will) but I don't have better idea so go ahead.

JoanneBogart:
Yes, it is by no means obvious!   The best we can hope for is to not be misleading.  We need to document it clearly and extensively. When I change all the existing config files I can add a comment at the top of each explaining.  Perhaps we can also add one or more example files either in the catalog_configs directory or somewhere else in the repo.

JoanneBogart:
@heather999 The code in branch u/jrbogart/root_dir is getting close to final.  Before it can be used as intended we need a directory to act as the root dir for catalogs which will probably mostly contain sym links to wherever each catalog actually is.  Where should this directory be? Once this is determined I can make the sym links and update config files (still on the branch until we're sure it's working).

GCRCatalogs/register.py
```diff
@@ -33,6 +76,21 @@ def load_yaml(yaml_file):
...
 
+def _resolve_dict(d):
+    for (k,v) in d.items():
+        if k in _PATH_LIKE_KEYS and isinstance(v, str):
+            if v.startswith('/'):
+                continue
+            elif _ROOT_DIR is None:
+                raise Exception('GCRCatalogs root dir not set')
+            else:
+                d[k] = os.path.join(_ROOT_DIR, v)
```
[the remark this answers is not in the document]
JoanneBogart: This code has been modified in the latest version.  In case _ROOT_DIR is referenced in a path but not set it issues a warning.

GCRCatalogs/register.py
```diff
@@ -3,24 +3,77 @@
...
+# Substitute for the following string if it starts a value for a path-like
+# keyword
+_ROOT_DIR_SIGNAL = "^/"
 _YAML_EXTENSIONS = ('.yaml', '.yml')
 
+# Keys appearing in yaml files whose values may be paths relative to _ROOT_DIR
+_PATH_LIKE_KEYS = ('filename', 'addon_filename', 'base_dir', 'root_dir',
+                   'catalog_root_dir', 'header_file', 'repo', 'table_dir')
+
+_DICT_LIST = ('catalogs')
```
yymao, line 22: I think this should be a tuple, right?
```suggestion
_DICT_LIST = ('catalogs',)
```

GCRCatalogs/register.py
```diff
@@ -33,6 +86,24 @@ def load_yaml(yaml_file):
         config = yaml.safe_load(r.raw)
     return config
 
+def _resolve_dict(d):
```
yymao, line 89:
```suggestion
def _resolve_dict(d):
    """
    input dictionary `d` will be modified in-place
    """
```

GCRCatalogs/register.py
```diff
@@ -54,15 +125,15 @@ def content(self):
...
-
+    
 class ConfigRegister():
     def __init__(self, config_dir):
         self._config_dir = config_dir
         self._configs = dict()
         self._configs_resolved = dict()
 
         for config_file in os.listdir(self._config_dir):
+            #print(config_file)                    ##  DEBUG
```
yymao, line 136:
```suggestion
```

GCRCatalogs/register.py
```diff
@@ -102,6 +173,11 @@ def resolve_config(self, config, past_refs=None):
 
                 return self.resolve_config(base_config, past_refs)
 
+        # Finally resolve relative paths in the config
+        config = _resolve_dict(config)
+
+        # temp for debugging: write out resolved config
+        print(yaml.dump(config))
```
yymao, line 180:
```suggestion
```

GCRCatalogs/site_config/site_rootdir.yaml
```diff
@@ -0,0 +1,2 @@
+nersc:  '/global/projecta/projectdirs/lsst/groups/CI/gcr_rootdir_test'
+in2p3:  '/sps/lsst/data/desc/gcr_rootdir'
```
yymao, line 2:
```suggestion
nersc: '/global/projecta/projectdirs/lsst/groups/CI/gcr_rootdir_test'
in2p3: '/sps/lsst/data/desc/gcr_rootdir'
```

setup.py
```diff
@@ -40,5 +40,5 @@
         'focal_plane': ['scikit-image', 'pandas'],
         'full': ['h5py', 'sqlalchemy', 'pymssql<3.0', 'pandas', 'tables', 'scikit-image', 'healpy', 'pyarrow'],
     },
-    package_data={'GCRCatalogs': ['catalog_configs/*.yaml']},
+    package_data={'GCRCatalogs': ['catalog_configs/*.yaml','site_config/*.yaml']},
```
yymao, line 43:
```suggestion
    package_data={'GCRCatalogs': ['catalog_configs/*.yaml', 'site_config/*.yaml']},
```

tests/test_reader_modules.py
```diff
@@ -3,6 +3,12 @@
 """
 import pytest
 import GCRCatalogs
+import os
+
+if GCRCatalogs.register._ROOT_DIR is None:
+    # Adust so test can be run anywhere
+    GCRCatalogs.register.set_root_dir(os.getenv('HOME'))
+    GCRCatalogs.register._ROOT_DIR = str(os.getenv('HOME'))
```
yymao, line 11: I think eventually we agree that `GCRCatalogs` and its modules should be importable even if ROOT_DIR is not set?
```suggestion
```

GCRCatalogs/register.py
yymao: Right, but the issue is that, without the comma, `_DICT_LIST` would be a string rather than a tuple.
```python
>>> _DICT_LIST = ('catalogs')
>>> type(_DICT_LIST)
str
```

GCRCatalogs/site_config/site_rootdir.yaml
JoanneBogart: It's hard to tell from the "suggested  change" display above, but I *think* you're asking that the two spaces following : be changed to a single space.   I've made that change.  If you want all yaml to conform to a very specific style like this I suggest you document it somewhere in the package (or is it already there and I've missed it?)

setup.py
JoanneBogart: Similar to above.  I guess you want a space following the comma?  I've added it.

GCRCatalogs/site_config/site_rootdir.yaml
yymao: Yes I was removing the extra space. It's not a big deal though. We don't enforce a strict formatting rule and I removed them only because I happen to notice them.

JoanneBogart:
Thanks for the tip - I've started up pytest.    If the run looks ok I will go ahead and merge.

PR #421, opened 2020-03-16
Several minor bug fixes for v0.15.0

yymao:
A full test run reveals some minor bugs in our planned release v0.15.0. 

- The `_lsst` bands are not available in  Buzzard v1.6. This PR removes them. cc @j-dr 
- The underlying repo for `dc2_single_visit_run2.1i_237263r` is no longer compatible with the updated desc-stack. I think it's ok to just remove this catalog unless @fjaviersanchez disagrees. 
- Removing `dc2_object_run1.2i_tract5063_with_metacal` because `dc2_metacal_run1.2i_tract5063` never exists. cc @EiffL


EiffL:
ok that looks good, but for my own education, why does `dc2_object_run1.2i_tract5063_with_metacal` never exists?

yymao:
@EiffL `dc2_object_run1.2i_tract5063_with_metacal` does exist, but it is a composite catalog which makes use of `dc2_metacal_run1.2i_tract5063` --- the latter never exists in GCRCatalogs, and hence I am removing the former. 

As to why `dc2_metacal_run1.2i_tract5063` never exists that's beyond me...

EiffL:
ok, fair enough, but in that case, shouldn't we instead add a `dc2_metacal_run1.2i_tract5063` entry? If I understand correctly, there is  no other small test catalog for metacal

yymao:
Certainly, but only if you're up to help with that :wink:

EiffL:
Hum....  yeah no, ok. It's hold stuff anyway, there is no metacal for the rest of this run, only for this small patch. I'm ok with your modifs

GCRCatalogs/buzzard.py
```diff
@@ -175,42 +172,25 @@ def _subclass_init(self,
...
                                        'g_lsst', 'r_lsst', 'i_lsst', 'z_lsst', 'y_lsst', 'Y_wfirst',
                                        'J_wfirst', 'H_wfirst', 'K_wfirst']):
-
                     i += 6
-
                     self._quantity_modifiers['Mag_true_{}_z0'.format(b)] = (_abs_mask_func, 'auxmag/AMAG/{}'.format(i))
                     self._quantity_modifiers['mag_true_{}'.format(b)] = (_mask_func, 'auxmag/TMAG/{}'.format(i))
                     self._quantity_modifiers['mag_{}'.format(b)] = (_mask_func, 'auxmag/LMAG/{}'.format(i))
 
-            else:
```
EiffL: this looks like a severe delete, but I trust you have removed them for good reasons :-)

yymao: Yes, thanks for pointing this out -- I should have explained it better.

So for lines where you see `truth/OMAG`, `truth/OMAGERR`, `truth/AMAG`, and `truth/TMAG` -- these for all for DES filters, and they exist for all versions. Hence they are moved to the new lines 175-179. 

For the two lines where you see `lsst/AMAG` and `lsst/TMAG` --- those are only available in v1.9.2 and above, and they have been dealt with in lines 181--188. 

There's one line where it maps `mag_true_x` to `mag_true_x_lsst`. This should be done for all versions, hence I move it to line 190-193. 

I think this explanation also works as me double checking my work --- so thanks for prompting it.

EiffL: Ok thanks Yao!

yymao:
Thanks @EiffL --- I also just finished another full test run on this PR and all tests passed! Merging now and will make a release.

PR #422, opened 2020-03-25
Update setup.py

boutigny:
Copy the site_config directory into the miniconda directory tree


heather999:
Please disregard the `desc-env-test` check, that workflow wasn't meant to be run in this circumstance of a PR on a branch.  The failure in no way indicates that the change requested won't work.  I've since updated the workflow to only run when changes are pushed to master.

PR #426, opened 2020-05-01
U/sschmidt23/run2.2i pz

sschmidt23:
This PR adds a Run2.2i photoz catalog (photoz_run2.2i_dr3.yaml), an associated subclass to handle parquet rather than hdf5 files (dc2_photoz_parquet.py), and a Run2.2i composite catalog that matches to the main dc2_object_run2.2i_dr3 catalog row by row (dc2_object_run2.2i_dr3_with_photoz.yaml).


sschmidt23:
I did a fresh clone of the repo to make sure it worked and everything checked out so I think it's ready to merge.  Thanks again for tracking down that one issue with the pdf objects.

PR #427, opened 2020-05-01
Update to v0.16.1

yymao:
Let's make a new release with the newly added photo-z catalog (#426). 

Is there any other features and/or additions that this release should wait for? If not, I'll tag it by EOB Monday. 

cc @sschmidt23 @JoanneBogart @heather999 @johannct


evevkovacs:
@yymao  I just submitted a PR to add yaml files for several versions of baseDC2 that I have been testing with DESCQA. It would be good to include them in the new release. I believe that I need to modify the absolute paths to relative ones, but otherwise the yaml files should be correct. No other code modifications are required. Please let me know about the paths.

yymao:
@evevkovacs are these catalogs already on CFS? If so, you can supply an absolute path in the config for now, as right now the default root dir is still pointing to projecta. We will update all configs that still contain absolute paths once the CFS space is fully ready. 

(However, let me just cc @JoanneBogart here --- she knows the situation better so if she gives you any instructions that contradict mine, please follow her lead :slightly_smiling_face:)

evevkovacs:
@yymao @JoanneBogart  Yes, these catalogs are on CFS already. So I will leave the yaml files as is, and we can fix them later.

johannct:
yes modifications to DR6 yaml files (absolute paths -> relative paths)

yymao:
@johannct Do you mean Run 2.2i DR6 object catalogs? Are they already on NERSC? I don't think they have been included in GCRCatalogs yet.

johannct:
@yymao https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/dc2_metacal_run2.2i_dr3.yaml and the like have absolute paths

yymao:
OK, yes, DR3 catalogs have absolute paths. But we haven't changed our global root dir to cfs yet.

@JoanneBogart should we take care of this now or wait until the `catalogs` directory organization is done?

PR #428, opened 2020-05-01
yaml files for baseDC2_5000 catalogs

evevkovacs:
Added yaml files to define various versions of baseDC2_5000. These are needed for running tests and comparing with the baseDC2 versions underlying cosmoDC2.
Do I need to switch to using catalog root_dir instead of absolute path now?


yymao:
Thanks @evevkovacs -- I think this looks fine for now. As I mentioned in #427, in the future we'll need to update the absolute path. 

I'll let @JoanneBogart review this just to be sure.

evevkovacs:
@yymao Changes have been made (Commit message got truncated because I used a back-quote) Should be ready to merge

PR #430, opened 2020-05-07
Refactor register.py

yymao:
I refactored `register.py` in this pull request. The motivation is that there are now lots of functions and module variables in `register.py`, but many of them are in fact separable. Hence, a refactor can make future maintenance and feature implementation easier. 

The diff in the PR is not going to be very helpful, so let me outline the new structure here. I suggest reading the new file directly when reviewing. 

During runtime all the configs are stored in a module-level variable `_config_register`, which is an instance of `ConfigRegister`. With the exception of a few helper functions at the beginning, all other module-level functions should be functions that access or manipulate `_config_register`. 

Previously, `root_dir` is stored as a module-level variable, but it really should be a property of `_config_register`, because conceptually, a user can create two instances of `ConfigRegister` that both have the same set of configs but different `root_dir`'s. 

(This is actually the _real_ motivation of this refactoring, as I noticed a bug -- if one loads a catalog, and then sets a new runtime `root_dir`, and then reload the same catalog again, the newly set `root_dir` won't take effect because it was cached. While this is not a common use case, it highlights the inconsistency between the use of module-level variables and object attributes.) 

So the goal is to move everything into object attributes. Instead of just putting all the `root_dir`-related functions into `ConfigRegister`, I put them in a new `RootDirManager` class, which contains all the functionality regarding changing and setting `root_dir`. 

The `ConfigRegister` class is now a mixin of `RootDirManager` and `ConfigManager`, the latter deals with identifying all the config files, and resolve `alias`/`based_on`. 

Individual config is wrapped by the `Config` class like before. But in this refactoring, I make sure that all the single-config-related functions are actually in `Config` (previously some are methods of `Config`, but others are module-level functions). 

For example, if `config` is an instance of `Config`, then it would be ideal if I can simply access `config.resolved_content` to get the resolved dictionary. Originally once has to call `_config_register.get_resolved("catalog_name")` to resolve alias, but to call `_resolve_dict(catalog_dict)` to resolve root dir. This is now homogenized. `ConfigManager` has a `resolve_refernce` method for resolving `alias` and `based_on` and `RootDirManager` has `resolve_root_dir`. Both methods are added to `Config.resolvers` (see around line 418). This way, when one accesses `config.resolved_content` on a single config instance, it knows how to resolve both `alias`/`based_on` and also `root_dir`. 

And, last but not least, this PR adds the concept of a "pseudo" config file. If a `is_pseudo_entry` key exists, then the config does not need to be loadable (i.e., it doesn't need to have `subclass_name`). 

In addition, many more convenient options are added to `get_available_catalogs` (they are passed to `ConfigManager.get_configs`) to help selecting specific sets of config files.

### DIFF
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -507,21 +507,21 @@ def reset_root_dir():
 
 def get_available_catalogs(
     include_default_only=True,
-    names_only=True,
+    names_only=False,
     name_startswith=None,
     name_contains=None,
     **kwargs
 ):
     """
-    Returns a list of all available catalogs.
+    Returns a dictionary of all available catalogs and their corresponding configs.
 
     Parameters
     ----------
     include_default_only: bool, optional (default: True)
         When set to False, returned list will include catalogs that are not in the default listing
         (i.e., those may not be suitable for general comsumption)
-    names_only: bool, optional (default: True)
-        When set to False, ruturns a dictionary where keys are catalog names and values are corresponding configs.
+    names_only: bool, optional (default: False)
+        When set to True, ruturns just a list of catalog names.
     name_startswith: str, optional (default: None)
         If set, only return catalogs whose name starts with *name_startswith*
     name_contains: str, optional (default: None)
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -69,7 +69,7 @@ def load_catalog_from_config_dict(catalog_config):
     Loads and returns the catalog specified in *catalog_config*.
     """
     return import_subclass(
-        catalog_config["subclass_name"], __package__, BaseGenericCatalog
+        catalog_config[Config.READER_KEY], __package__, BaseGenericCatalog
     )(**catalog_config)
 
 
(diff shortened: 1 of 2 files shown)


GCRCatalogs/register.py
```diff
@@ -1,301 +1,492 @@
...
+        except PermissionError:
+            warnings.warn("root_dir has been set to '{}' but you have no permission to access it".format(path))
+        except OSError as e:
+            warnings.warn("root_dir has been set to '{}' but errors may occur when you try to access it: {}".format(path, e))
+        self._custom_root_dir = path
+
+    def reset_root_dir(self):
+        self._custom_root_dir = None
+
+    def resolve_root_dir(self, config_dict, config_name=None, record=None):  # pylint: disable=unused-argument
```
JoanneBogart, line 146: Add something about optional arguments `config_name` and `record` to header comments.  Their purpose is obscure.

GCRCatalogs/register.py
```diff
@@ -1,301 +1,492 @@
...
+    @property
+    def is_pseudo(self):
+        return self.get("is_pseudo_entry")
+
+    @property
+    def is_alias(self):
+        return self.get("alias")
+
+    @staticmethod
+    def _has_reference_keys(d):
```
JoanneBogart, line 248: Suggest you add a static to the class, e.g. `REF_KEYS = ("alias", "based_on")` since they appear in two or three different places in the code, and it's conceivable at some point we might add another reference key needing resolution.

GCRCatalogs/register.py
yymao: Good point. Expanded docstring to make it clear.

GCRCatalogs/register.py
yymao: Sure. While I am at this, I made all special key names static attribute to `Config`.

yymao:
Thanks @JoanneBogart. I've addressed your comments and this PR is ready for re-review.

yymao:
I have added a new module-level function `retrieve_paths` that can be called like:

```python
>>> GCRCatalogs.retrieve_paths(name_starswith="buzzard")
[('buzzard_high-res_v1.1',
  '/global/projecta/projectdirs/lsst/groups/CS/Buzzard/Buzzard-highres',
  '/global/projecta/projectdirs/lsst/groups/CS/Buzzard/Buzzard-highres'),
 ('buzzard_v1.6',
  '^/Buzzard/Buzzard-0/addgalspostprocess',
  '/ROOT_DIR/Buzzard/Buzzard-0/addgalspostprocess'),
....
]
```

yymao:
Thank you @JoanneBogart. I'll merge it tomorrow.

Issue #425, opened 2020-04-18
Add deprecation warnings when users load older catalogs that we plan to drop support


yymao:
As the number of catalogs in `GCRCatalogs` grows, we probably would want to start to drop support of the older versions of some catalogs, especially those that were created for testing or other temporary purposes. 

Before we do that, `GCRCatalogs` should print out clear deprecation warnings when users try to load them, so that we can make sure those catalogs are safe to be removed.

yymao:
Let's at least start to collect the list of the "soon-to-drop-support" catalogs. I'd like to ask:

@evevkovacs - for protoDC2 and cosmoDC2
@j-dr - for buzzard
@wmwv @fjaviersanchez @johannct - for DC2 catalogs

heather999:
Can we consider deprecating DC2 Run1.1p, as in:
https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/dc2_object_run1.1p.yaml  ?
I'm not aware anyone would be seriously utilizing Run1.1p data at this point.  Soon I could imagine doing the same for Run1.2x, but there's some possibility people are still active on that data.

katrinheitmann:
What does "deprecate" entail? The 1.2p/i runs are the only ones where we have the phosim/imsim pair. Maybe there is not a lot of traffic right now, but it is an important dataset. We are also still going to have it in the paper.

johannct:
I think the initial phrasing of @yymao is correct : drop support of "older versions of some catalogs". In the case of 1.2, there is only one version. In general deprecation results from either the replacement of a functionality with another, or the dropping of features that cannot be supported anymore. In our case deprecation will thus results : 1/ in the first case from a new version of a given catalog (say we feel like reprocessing 1.2 with stack 19.0.0) or 2/ in the second case we just find ourselves short on storage and need to stage some low-usage catalogues.

johannct:
For 1.1 specifically, I am not sure whether it can be considered as an early version of 1.2p, I just can't remember. If it can, we should deprecate it, as if asked by someone about using it, we would immediately answer "rather use 1.2p".

evevkovacs:
@yymao protoDC2_v2.1.2, protoDC2_v3.0 and protoDC2_v5.0 should be kept.  All versions of cosmoDC2 currently in the catalogs area and previously listed by Heather and me also need to be retained for the forseeable future.

yymao:
[...]

PR #433, opened 2020-05-15
allows deprecated key in config

yymao:
Per the discussion in #425, this PR adds support of a new `deprecated` keyword in the catalog config files. When a user attempts to load a catalog that has `deprecated` key, a `DeprecationWarning` will be displayed. 

For now this PR has not yet added the `deprecated` key to any configs yet. I'd like to have this technical part review and merged first, and have a different PR to actually adding these  `deprecated` key to catalog configs.


yymao:
Thanks @JoanneBogart

PR #434, opened 2020-05-15
Fix two issues in a recent test run

yymao:
This PR fixes two issues that @JoanneBogart noticed after running the test:

   1. Missing filter warnings. This is due to a non-ideal implementation of `_generate_modifiers`. I use this opportunity to make the implementation more robust. In particular, a new `_detect_available_bands` method is now implemented for several catalogs to automatically detect available filter names.

   2. AttributeError during `DC2ObjectCatalog` cleanup. This PR adds a guard to skip the cleanup if it's already done.


yymao:
Thanks @JoanneBogart!

PR #436, opened 2020-05-16
More bug fixes to legacy catalogs

yymao:
This PR contains a few more bug fixes to legacy catalogs:

 1. Remove `hsc-pdr1-xmm` since the files have been purged.
 2. Comment out unfinished modifiers in `dc2_dia_object.py` and fix a f-string bug.
 3. Check the availability of modelfit flux and flag in legacy object catalog.  
 4. Update path to instance catalog examples and fix a deprecation warning.

### DIFF
--- a/GCRCatalogs/dc2_dia_object.py
+++ b/GCRCatalogs/dc2_dia_object.py
@@ -105,8 +105,8 @@ class DC2DiaObjectCatalog(DC2DMTractCatalog):
             modifiers[f'magMean_{band}'] = (convert_nanoJansky_to_mag,
                                             f'psFluxMean_{band}')
             modifiers[f'magMeanErr_{band}'] = (convert_flux_err_to_mag_err,
-                                               'psFluxMean_{band}',
-                                               'psFluxMeanErr_{band}')
+                                               f'psFluxMean_{band}',
+                                               f'psFluxMeanErr_{band}')
             modifiers[f'magMeanStd_{band}'] = (convert_flux_err_to_mag_err,
                                                f'psFluxMean_{band}',
                                                f'psFluxSigma_{band}')
--- a/GCRCatalogs/dc2_dia_object.py
+++ b/GCRCatalogs/dc2_dia_object.py
@@ -70,26 +70,26 @@ class DC2DiaObjectCatalog(DC2DMTractCatalog):
             'diaObjectId': 'diaObjectId',
             'ra': (np.rad2deg, 'coord_ra'),
             'dec': (np.rad2deg, 'coord_dec'),
-#            'radecCov':  # A covariance matrix for the uncertainty in ra, dec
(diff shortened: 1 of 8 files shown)


yymao:
A full test run with this PR is done and all tests have passed.

GCRCatalogs/dc2_dia_object.py
```diff
@@ -70,26 +70,26 @@ def _generate_modifiers(dm_schema_version=4, bands='ugrizy', **kwargs):  # pylin
...
+            # 'lcNonPeriodic': '',
+            # # These are possible to calculate if you require the Object Table
+            # 'nearbyObj': '',
+            # 'nearbyObjDist': '',
+            # 'nearbyObjLnP': '',
         }
 
-        modifiers['good'] = (create_basic_flag_mask,)
-        modifiers['clean'] = modifiers['good']
+        # modifiers['good'] = (create_basic_flag_mask, )
```
JoanneBogart, line 91: Why are these now commented out?  Should they have not been there in the first place?

GCRCatalogs/instance_catalog.py
```diff
@@ -266,7 +266,7 @@ def _generate_native_quantity_list(self):
         return native_quantities
 
     def _pd_read_table(self, obj_type, **kwargs):
-        return pd.read_table(
+        return pd.read_csv(
```
JoanneBogart, line 269: From the documentation, it looks like read_csv and read_table behave identically when delim_whitespace is True.   Why did you change it?

GCRCatalogs/dc2_dia_object.py
yymao: This line appears to be an unfinished line. There should be columns names following the `create_basic_flag_mask` function ([example](https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/dc2_object.py#L360)). 

However, I checked and could not find these columns needed to define `good` in the underlying file, so I commented out them for now. We'll need to check with @wmwv if we want to properly resolve this.

GCRCatalogs/instance_catalog.py
yymao: `read_table` was deprecated at one point but apparently it has been "un-deprecated" now (see https://github.com/pandas-dev/pandas/issues/25220). 

The version of pandas that is currently in our `desc-stack` is not up-to-date enough so it still prints out a deprecation warning for `read_table`. However, it's true that this edit is not necessary.

JoanneBogart: Thanks; that explains it.

GCRCatalogs/dc2_dia_object.py
JoanneBogart: Good enough for now.  Thanks.

PR #442, opened 2020-05-20
Warn only if site from env differs from socket

yymao:
I am asking Heather to set the environment variable `DESC_GCR_SITE` in our python environments on NERSC (https://github.com/LSSTDESC/desc-python/issues/39). I then realize that this would cause a warning to be always printed out. 

This PR is to make it only warn if the site name from DESC_GCR_SITE is inconsistent with the site name from socket.

### DIFF
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -128,7 +128,7 @@ class RootDirManager:
         """
         site_from_env = os.getenv(self._DESC_SITE_ENV)
         site_from_socket = socket.getfqdn()
-        if site_from_env and site_from_env not in site_from_socket:
+        if site_from_env and site_from_socket and site_from_env not in site_from_socket:
             warnings.warn("Site determined from env variable {} = {}".format(self._DESC_SITE_ENV, site_from_env))
             return site_from_env
         return site_from_socket
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -128,8 +128,13 @@ class RootDirManager:
         """
         site_from_env = os.getenv(self._DESC_SITE_ENV)
         site_from_socket = socket.getfqdn()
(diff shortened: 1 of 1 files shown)


GCRCatalogs/register.py
```diff
@@ -126,12 +126,12 @@ def _get_site_info(self):
...
-        if v:
-            warnings.warn("Site determined from env variable {}".format(self._DESC_SITE_ENV))
-            return v
-        return socket.getfqdn()
+        site_from_env = os.getenv(self._DESC_SITE_ENV)
+        site_from_socket = socket.getfqdn()
+        if site_from_env and site_from_socket and site_from_env not in site_from_socket:
+            warnings.warn("Site determined from env variable {} = {}".format(self._DESC_SITE_ENV, site_from_env))
+            return site_from_env
+        return site_from_socket
```
JoanneBogart, line 134: I agree there should be no warning in case `site_from_env` is not empty and is in `site_from_socket` but, to make this more foolproof I prefer
* if `site_from_env` has a value but `site_from_socket` doesn't, return `site_from_env` (just in case it's possible for socket.getfqdn() to return None or empty string)
* they both have a value and `site_from_env  in site_from_socket`, return `site_from_env` (without any warning) rather than `site_from_socket`.  We know the return value from getfqdn( ) includes extraneous stuff, conceivably it can include more than one legitimate site value.  It's not likely, but I think it's even less likely that someone would provide a value in the environment value with this problem.

yymao: Thanks. That sounds reasonable. I'll update the logic here.

yymao:
OK I implement the logic as you described. 

I also try to skip warning when `site_from_env == "nersc" and site_from_socket.startswith("nid")`. I am on the fence about this. What do you think?

yymao:
Thanks @JoanneBogart !

PR #446, opened 2020-05-25
Add `deprecated` and `addon_for` to configs

### LINKED ISSUE #425: Add deprecation warnings when users load older catalogs that we plan to drop support

yymao:
This PR adds the `deprecated` and `addon_for` keywords to catalog config yaml files as appropriate. 

Keyword `deprecated` is added to older catalogs as specified in #425. A deprecation message is included when an alternative catalog is available. 

Keyword `addon_for` is added to catalogs that are not supposed to be used as a standalone catalog (i.e., those are parts of a composite catalog). 

`register.py` is modified so that by default, deprecated and addon catalogs will not show up when user calls `get_available_catalogs`. Deprecated and addon catalogs still can be loaded with `load_catalog`. 

This PR also adjusts the `include_in_default_catalog_list` keyword for several commonly used catalogs. 

`README.md` is updated to reflect these changes (in particular, deprecated catalogs are removed). 

Fix #425.


yymao:
cc several catalog creators in case they want to double check the edits: @evevkovacs @j-dr @EiffL @sschmidt23 @erykoff @johannct @wmwv

yymao:
Thanks @JoanneBogart.  Documenting reserved keywords in `CONTRIBUTING.md` sounds like a good idea, and I think it makes sense to include that in this PR. I'll work on that. 

I actually thought of adding code to check the contents of `addon_for` or `deprecated`, but later concluded it's not as helpful as what I have thought.
Mainly the reason is that the purpose of these two keywords is mainly to _document_ the catalogs, and they do not alter the catalog loading behavior nor trigger any automation. 

For example, the `addon_for` key is really meant to let the users know that this catalog is not supposed to be used as a standalone catalog. The code would not automatically generate a composite catalog when detecting the `addon_for` keyword (and it can't, since more information is needed to generate a composite catalog). 

Similarly for the `deprecated` key, the deprecation message is meant for human (as sometimes multiple alternatives, each with different suggested use), and the alternative catalog won't be load automatically. 

If we are going to implement some checks, I am also not sure if runtime checks make sense. Since these keywords do not alter the catalog loading behavior, there is no reason to raise errors or warning during runtime if they are set wrong. Maybe implementing them in a test would make more sense. 

Anyways, I didn't come up with a good enough reason to convince myself to put effort in implementing these checks. Further thoughts are still welcome. If it becomes more involved we can open a new issue.

JoanneBogart:
@yymao I agree a test might be a better place to do those checks.  It's most likely to be useful when a new release is imminent, to be sure that everything is still consistent.

yymao:
I can think of a useful test for `addon_for`, but don't yet have a good idea for `deprecated`.

JoanneBogart:
For `deprecated` I was thinking thing value should be structured, e.g.
```
deprecated: 
    status: true
    recommended: another-catalog
    message: anything else the user should know
```
where the `recommended` and `message` fields would be optional. If `recommended` is present, the test would check to see if `another-catalog` is registered, and, at load time, a warning would be issued suggesting its use.   If `message` is present, its value would also be put out as a message at load time (in addition to the "suggested" message, if there is one)

yymao:
Thanks for the suggestion @JoanneBogart.  Maybe we should step back and consider what our goals are. I can think of three goals for now:

1. To make all configs more structured and uniform in general.
2. To avoid human error in the config content. 
3. To enable potential future automation of certain tasks. 

I am certainly in favor of all three. I think (1) is the overall direction we want to move to, but would require some careful thinking in the design.  Even just for the `deprecated` key, there are cases where there are multiple recommended alternatives (like buzzard) and even cases where there are multiple recommended alternatives, each of which should have a different note (like protoDC2 and several Run1.x). I bet we can still come up with a solution, but I would prefer that we take time and actually think about this (in the context of all other keywords as well, if we really want to come up with a consistent design). 

(2) and (3) are both important too. But I am struggling with them because neither is really important for `deprecated` and `addon_for`, since neither keywords changes how catalog is loaded. So a typo in `deprecated` is similar to a typo in `description` or `creators`. Not that we shouldn't watch out for them, but whether it is useful to invest in auto check on them. 

Also, I should point out that we always need human check on the content of `deprecated` and `addon_for`. Even if the automation check can detect whether the catalog specified exist, it won't know if that is indeed the catalog that should be recommended. 

I am not saying this is a reason for not having automatic checks. Both are important. The real question I am asking is, given that (a) human check is always needed for `deprecated` and `addon_for`, and that (b) we will restructure/uniformize all the config keywords in a not-so-distant future, is it worth the effort to invest automatic checks on `deprecated` and `addon_for` right now?

JoanneBogart:
@yymao I agree with those goals.  
It did occur to me that there could be more than one suggested alternative, or that there could be other types of complications.  To support multiple recommendations the value for `recommended` could be a list.  I think there is some value in a test which would see if they exist (there could be a typo in the recommended config name or a previously-recommended catalog might no longer exist or could itself be deprecated - that is, basically helping with goal (2) ), but agree it may not be worth putting too much effort into this kind of thing now if more comprehensive changes are coming soon.

yymao:
Yep I certainly see the value, but just feel that we may need to give it some more thinking so that the solution is indeed useful and still elegant enough. 

For example, when I was reading your comment I also remembered that we need to check the recommended catalogs themselves are not deprecated. 

I was also thinking if there is a way to keep the current, more human readable form, instead of introducing a nested structure. For example, we can require catalog names to be surrounded by a "`", such as:

```yaml
deprecated: Use `dc2_object_run1.2p` (most compatible) or `dc2_object_run2.2i` (latest) instead.
```

This way the test can parse the catalog names from a human readable sentence.

JoanneBogart:
There are trade-offs depending on degree of structure; I don't know what the optimal solution is. 

If it's more formal, more of the work is done by the YAML parser rather than GCR code.  I'm guessing that the number of people who read the config file in its raw form is small, so as long as the documentation on how to *write* a `deprecated` section is clear and keywords are carefully chosen to do a good job of suggesting the intent, I think we're ok. 

 But we shouldn't go overboard, building a complicated structure for a secondary feature.

yymao:
> If it's more formal, more of the work is done by the YAML parser rather than GCR code.
I see what you are saying. I wasn't worried about people reading the raw yaml config, but how to handle the deprecation warning printout. We'll also need to introduce some code to format the structured information to human-readable printouts.

yymao:
@JoanneBogart I've updated `CONTRIBUTING.md` ([rendered version](https://github.com/LSSTDESC/gcr-catalogs/blob/issues/425/add-deprecation-flag/CONTRIBUTING.md)). Do you mind giving it a review? 

After writing this, I feel I am more inclined to punt on writing config tests for now. I am actually more convinced that we should have config tests, but (1) we should do it systematically (i.e., not just for `deprecated` and `addon_for`), and (2) we may want to also just unify our keyword requirements at once. So unless there is a strong feeling from others that we must have tests for `deprecated` and `addon_for` right now, I would suggest that we open a new issue and proceed with this PR. Comments on this point are still welcome.

JoanneBogart:
Thanks @yymao That sounds like a good strategy.  I don't see any urgency in implementing new tests. Holding off in favor of a more comprehensive approach is fine as long as it's not forgotten.  A new issue will take care of that.

CONTRIBUTING.md
```diff
@@ -1,57 +1,104 @@
...
+# Reserved keyword
+alias: another_catalog_name  # When set, this yaml file will act as an alias to another_catalog_name.
+                             # All other keywords in this config will be ignored.
+based_on: another_catalog_name  # When set, this yaml file will inherit the config content of another_catalog_name.
+                                # All other keywords in this config will overwrite the inherited content.
+include_in_default_catalog_list: true  # When set, this catalog will show up in the "recommended" catalog list.
+                                       # Only catalogs that are intended for general comsuption should set this keyword.
+addon_for: another_catalog_name  # Set this keyword to indicate that this catalog is intended to be used ONLY as
+                                 # an addon catalog for another_catalog_name, and is not for standalone use.
+                                 # Note that setting this keyword does not prohibit users from load this catalog.
```
JoanneBogart, line 46: Change `load` to `loading`

CONTRIBUTING.md
```diff
@@ -1,57 +1,104 @@
...
+deprecated: "A short deprecation message."  # Set this keyword if this catalog has been deprecated and should
+                                            # no longer be used.
+                                            # Setting this keyword does not prohibit users from load this catalog.
+is_pseudo_entry: true  # Set if this config is an pseudo entry and should be ignored by the register.
```
JoanneBogart, line 50: Explain a little more what's meant by "pseudo entry", e.g.
```
Set if this config is a pseudo entry (not intended to be loaded via GCR; no reader required) and..
```

CONTRIBUTING.md
```diff
@@ -1,57 +1,104 @@
...
+For example, if the reader class asks for `catalog_root_dir` as an input argument to specify the location of the
+catalog files, you need to include `catalog_root_dir` as a keyword in the corresponding yaml config file,
+and set it to the correct location.
+All keywords in the yaml config file will be passed to the reader class.
+
+Below is a list of required, recommended, or reserved keywords that may appear in a yaml config file.
+
+```yaml
+# Required keywords
+subclass_name: <reader_module_name>.<ReaderClassName>  # Indicate which reader to use.
```
JoanneBogart, line 31: Need to mention exceptions.  E.g.
```
Note that subclass_name is not required when certain other keywords are present:
* alias, pseudo_entry - not required; will be ignored
* based_on - not required but will be used if supplied
See descriptions of these keywords below for more information on their use
```

yymao:
@JoanneBogart thanks! I've addressed all of your comments. I also decided to reformat the section since putting longer comments inline is making it a bit difficult to read. Hopefully the [new format](https://github.com/LSSTDESC/gcr-catalogs/blob/issues/425/add-deprecation-flag/CONTRIBUTING.md) also improves readability.

yymao:
I know it's not a good practice to mix different things in one PR.. but when I was rewriting the documentation I realized that we didn't actually prevent users from loading pseudo-entries, so I fixed that in f97a1bb.

yymao:
Thanks @JoanneBogart --- I just noticed a minor issue: the reader list should include deprecated and addon catalogs (since those are still using active readers), so I pushed a new change 6d22096. Do you mind re-review it? Thanks!

Issue #331, opened 2019-06-26
clarify flavor of photometry in object catalogs


rmandelb:
It would be beneficial to document in some easily-findable place (schema or possibly elsewhere?) what type of photometry is recorded in the object catalogs: forced photometry from the forced source table versus non-forced photometry from the meas table.

yymao:
@rmandelb sorry to suddenly get back to your almost-one-year-old question but isn't object catalog _by construction_ using forced photometry?

rmandelb:
I can give you an answer based on what I vaguely recall my concern being from a year ago, but I don't know if the reasons behind it still apply.   At that time, I believe I was considering the perspective of a newbie to DC2 products (thinking ahead to the collaboration meeting and potential new users) seeing there were many different types of catalogs here and in the tutorials, and wanting to understand their key properties:

The object catalog uses forced photometry, and my question was whether this was indicated in some reasonable place in this repo.  If somebody doesn't know that's definitionally true for an object catalog (since we refer to "object catalogs" in many places and of course we are not going to give the definition each time), they could have this question given that it is possible _in principle_ to make a catalog with non-forced photometry.

This was not a suggestion that we make any kind of major changes -- just that in the few key places where users may start to interact with / learn about these files (which I was thinking could happen on confluence, in the GCR catalog list or schema, and the tutorials), they should get an indication of the flavor of the photometry, enough that they could understand how it may be used.  Again, I could be misremembering or our documentation may have changed in the past year, or maybe the GCR repo is simply not an important place to do this, so I understand if you prefer to close this issue.

yymao:
Thanks for the clarification, @rmandelb. I agree that we use the term "Object Catalog" as a jargon, and it may not be obvious to new users what it means. 

I added a one-line description of Object Catalogs in #449 ([quick link to the rendered version](https://github.com/LSSTDESC/gcr-catalogs/blob/u/yymao/update-schema-md/GCRCatalogs/SCHEMA.md#schema-for-dc2-object-catalogs)). Would this alleviate your concern?

rmandelb:
That would be a reasonable way to do it, though I'm not sure whether the schema or the README is the right place to do the job.  

Going back to my original question, I think it's valuable to give some indication (again in ~a sentence) which quantities are measured in a consistent way across bands versus completely independently.

yymao:
[...]

PR #449, opened 2020-05-28
Re-organize SCHEMA.md

yymao:
This PR slightly re-organizes `SCHEMA.md`. In particular, I made it clear which parts are for Extragalactic Catalogs, and which parts are for DPDD-like catalogs. I also reduced some duplications and fixed several typos. 

While I am on this, I added short description (taken from [LDM-153](https://ls.st/LDM-153)) for Object, Source, and Force Source Catalogs, as an effort to address the issue @rmandelb mentioned in #331. 

[A rendered version can be found here](https://github.com/LSSTDESC/gcr-catalogs/blob/u/yymao/update-schema-md/GCRCatalogs/SCHEMA.md).

### DIFF
--- a/GCRCatalogs/SCHEMA.md
+++ b/GCRCatalogs/SCHEMA.md
@@ -96,7 +96,8 @@ Quantity Label | Unit | Definition
 
 ## DC2 DPDD-like Catalogs
 
-The schema for DC2 DPDD-like Catalogs, such as Object, Source, and Force Source Catalogs, generally follow these rules:
+The schema for DC2 DPDD-like Catalogs (such as Object, Source, and Force Source Catalogs, as defined in [LDM-153](https://ls.st/LDM-153)),
+generally follow these rules:
 
 - For quantities that are defined in [LSST DPDD](https://lse-163.lsst.io/dpdd.pdf), we follow DPDD's naming scheme.
 - For quantities that are defined in the above "Schema for Extragalatic Catalogs", we follow Extragalatic Catalogs' naming scheme ('GCRbase' below).
--- a/GCRCatalogs/SCHEMA.md
+++ b/GCRCatalogs/SCHEMA.md
@@ -115,6 +115,9 @@ In the tables below we list the name of the quantity, its units and definition a
 ### Schema for DC2 Object Catalogs
 
 Object Catalog contains information about static astronomical objects measured on a stacked (coadd) image.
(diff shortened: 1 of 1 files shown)


GCRCatalogs/SCHEMA.md
```diff
@@ -144,22 +153,12 @@ Quantity Label | Unit | Definition | GCRbase | DPDD
...
-- For quantities that are defined in neither and are newly defined for the coadd catalogs, we generally follow Extragalatic Catalogs' naming style.
-
-For quantities that are not yet documented in the table above, we document them below:
-- In the table below we list the name of the quantity, its units and definition and whether the name is defined in the GCRbase or DPDD.
-- Items marked with 'xx' are not exactly defined in the DPDD, but their name is taken from a related column in a different table.  E.g.
-   * there is no `x`, `y` in the DPDD Object table, but these are called `x`, '`y` in the DPDD Source table.  We don't have `xyCov` so we separately list `xErr` and `yErr`.
-   * `radec` is a pair in the DPDD, but we separate out into `ra`, `dec` here.
-   * The DPDD says `psCov`, but we only have the diagonal terms, so we call it `psErr`.
-- Quantities named with `x` or `y` are in the local coordinate system of the tract+patch.
+Forced Source Catalog contains photometry measurements about low signal-to-noise detections done on individual
```
wmwv, line 160: The Forced Source Catalog contains information on all Objects.

I think mentioning "low signal-to-noise detections" is confusing here, both "low signal-to-noise" and "detections".  

It's potentially of significant interest for low SNR things that may or may not have yielded detections on the individual image.
In the high SNR regime, the Forced Source Catalog entries should agree with the Source Catalog entries.

"Forced Source Catalog contains photometry based on fixed-position measurements from objects in the Object Table."

yymao: Thanks, @wmwv --- do we want to mention that it is done on individual exposures? 

"Forced Source Catalog contains photometry from individual exposures based on fixed-position measurements from objects in the Object Table."

wmwv: How about:"Forced Source Catalog contains fixed-position photometry measurements performed on individual exposures based on the positions from objects in the Object Table."

yymao: Updated. Ready for re-review. Thanks!

PR #450, opened 2020-05-29
Issue/447/pznaming

### LINKED ISSUE #447: Better naming scheme for photo-z addon/composite catalogs?
When working on #446, I noticed that some photo-z addon/composite catalogs have inconsistent names. 

While we haven't enforced a strict rule for naming addon and composite catalogs, they should generally follow these example patterns below:

- Main catalog: `cosmoDC2_v1.1.4`
- Addon catalog: `photoz_v1_for_cosmoDC2_v1.1.4` (not for standalone use)
- Composite catalog: `cosmoDC2_v1.1.4_with_photoz_v1` (joining main and addon catalogs)

Consider those example patterns, I identified a few catalogs whose names are not consistent with the patterns:

- `cosmoDC2_v1.1.4_image_photoz_magerr_10y_PDFcat`: name suggests it's a composite catalog, but currently it's an addon for `photoz_magerr_10y_estimate_image`
- `photoz_magerr_10y_estimate_image`: name suggests it's an addon, but not clear for what (it's for `cosmoDC2_v1.1.4_image`). While the postfix `_image` gives a hint, it may actually be doubly confusing for people who are not familiar with these catalogs.
- `photoz_magerr_10y_with_pdfs`: name suggests it's an addon, but not clear for what catalog. However, this is actually a standalone composite catalog(! shouldn't this be an addon for cosmoDC2?) 

And then there's another kind of inconsistency:

- `pz_calibrate_references_cosmoDC2_v1.1.4_image` and `pz_calibrate_references_cosmoDC2_v1.1.4_small`: correct addon naming scheme but prefix should be `photoz_`. 

Since we are doing some catalog deprecation, it may be a good time to fix some of these inconsistency. Thoughts, @sschmidt23?

sschmidt23:
This is to address and fix #447 on catalog names. 
The PR also fixes an inconsistency in the names (ODDS vs odds and pdf_bins vs pdf_bin_centers) in the Run2.2i parquet subclass.

 I updated the names on several catalogs and tested on a local branch that I could still grab a few columns.  I believe the names are more straightforward, the only tricky bit are due to the "intermediate" mock 10-year depth ugrizy magnitudes and errors catalog for cosmoDC2_v1.1.4 that were created before computing photo-z's.  We have composite catalogs matching each of the three pairs of these files, which may be confusing, but I can see all three of these combinations as being useful.  Not sure what to do to clarify, other than documenting well, though obviously open to suggestions (including on alternate names).

Files in question:
cosmoDC2_v1.1.4_image_with_photozs_v1.yaml
(composite catalog of cosmoDC2_v1.1.4_image and photoz PDF files)

and the two renamed yaml files: 
cosmoDC2_v1.1.4_image_photoz_magerr_10y.yaml → cosmoDC2_v1.1.4_image_with_10y_magnitude_error_mocks.yaml
(composite catalog of cosmoDC2_v1.1.4 with the mock ugrizy magnitudes and errors at 10 year depth)

and 

photoz_magerr_10y_with_pdfs.yaml → photoz_mock_mags_errors_pdfs_for_cosmoDC2_v1.1.4_image.yaml
(composite catalog of the mock ugrizy magnitudes and errors at 10y depths with the photoz PDFs)


yymao:
Thanks, @sschmidt23! The merging conflicts are likely due to a new merged PR of mine. I'll be happy to resolve the conflicts. 

The composite feature actually allows you to join multiple catalogs into one. Would that be useful to fix the issue you mentioned?

sschmidt23:
I think I tried a composite catalog of composite catalogs, and it occasionally caused crashes on jupyterlab because of the size.  That was a while ago, so it may be worth trying again, I can try that now.

yymao:
Yep, a composite of a composite is not going to work well because every layer of composite sacrifices some efficiency. 

What I meant is the you can have multiple catalogs in a single composite. The config would look like:

```yaml
catalogs:
  - catalog_name: cosmoDC2_v1.1.4_image
  - catalog_name: photoz_1
    matching_method: MATCHING_FORMAT
  - catalog_name: photoz_2
    matching_method: MATCHING_FORMAT
```

sschmidt23:
One potential issue: two of the catalogs are identical in length and would use MATCHING_FORMAT, the third only has a subset of entries and needs to match on galaxy_id, so I'm not sure how to set up that yaml file.

sschmidt23:
Ah, never mind, looks like what I tried just worked.  Let me run a couple more tests to be sure it's fully functional, if so I'll replace the composites with the triple composite file.  Thanks @yymao!

sschmidt23:
Yes, everything works, though I am now thinking "scatmag_[band]" and "scaterr_[band]" are terrible variable names, I should create a better modifier for that in the photoz_magerr subclass.  How does "photoz_mag_[band]" and "photoz_mag_err_[band]" sound?

yymao:
Great that it works! 

How about `mag_err_{band}_photoz`? We hav `mag_err_{band}_lsst` etc so they all look similar?

sschmidt23:
Yes, that also makes sense, I was thinking of putting photoz first in the name to make sure they stood apart from the other magnitude estimates, but consistency probably outweighs that.  I'll update the photoz_magerr.py and push in a moment.

sschmidt23:
Ok, it all works with newly renamed e.g. mag_i_photoz and mag_err_i_photoz.  I think this is all good now.  But, before we merge maybe I should update the Confluence documentation page for any users that are caught by the name changes in old scripts.

yymao:
Thanks, @sschmidt23. Let me resolve the conflicts and review it.

PR #451, opened 2020-05-29
Deprecate Run 2.1

yymao:
Based on our recent discussion in #desc-dc2-data-access, this PR marks Run 2.1 catalogs in GCRCatalogs deprecated. All the marked catalogs will still be accessible in the foreseeable future, but won't be listed in the default catalog lists.


yymao:
Thanks, @JoanneBogart!

PR #454, opened 2020-06-01
Add redmapper 0.5.7 catalog for cosmodc2 1.1.4

erykoff:
At long last, I am putting the redMaPPer 0.5.7 catalog from cosmodc2 1.1.4 in gcr-catalogs.  Many apologies for the delay.

I also need to have the catalogs copied into the proper directory.  The files in `/global/cscratch1/sd/erykoff/redmapper_staging` need to be copied to `/global/cfs/cdirs/lsst/shared/xgal/cosmoDC2/addons/redmapper_v1.1.4` before this can be tested and merged.


yymao:
Thanks @erykoff. 

With this new addition, should we deprecate `cosmoDC2_v1.1.4_redmapper_v0.2.1py` then? 

I think you can open a ticket [here](https://github.com/LSSTDESC/desc-help/issues) to make a file copy request. You can reference this new config so that they'll know where to copy the file to.

erykoff:
I have filed https://github.com/LSSTDESC/desc-help/issues/18 for the copy.
Yes, the old ones should be deprecated.  I will add that to the PR.

Issue #394, opened 2019-10-09
add how to use `config_overwrite` in README


yymao:
add how to use the `config_overwrite` option in README so that people who need cutting-edge catalogs can use it.

yymao:
This should be relative simple to fix --- if anyone is interested in helping that'd be great.

PR #456, opened 2020-06-03
Add Run2.2i DR6

yymao:
This PR adds Run2.2i DR6. It should work once the files are ready, but we may want to do a test run to make sure all tract files are valid.

One minor issue is that we used to have alias like `dc2_object_run2.2i` pointing to `dc2_object_run2.2i_dr3`. Since DR6 and its associated data products are not yet fully available, I keep the aliases point to DR3 still. But I remove them from the default listing (and README) and promote the configs that have explicitly spell out `dr3`. I thought this would reduce confusion, but comments on how to make this more clear is welcome.


heather999:
@yymao We now have 78 object catalogs for DR6 available at NERSC: `/global/cfs/cdirs/lsst/shared/DC2-prod/Run2.2i/dpdd/Run2.2i-dr6/object_table_summary`

johannct:
we also need to build the dpdd for Stéphane

yymao:
I'll first run a test to make sure all tract files are valid and have entries.

yymao:
I tested this PR and confirmed that there are 78 tracts, all of which are valid and can be read in without issue. Can one of you take a look at this PR and approve it? Thanks!

yymao:
@johannct sorry that I pushed some more changes after you've reviewed the PR. Here's a summary of the new changes --- 

Based on the discussion with @wmwv, I've updated how the configs are named and specified for object catalogs. Here's the "rules" (currently apply to Run 2.2i only, I don't think we should apply them to earlier runs)

1. Always use `drX` in the catalog name. So there are `dc2_object_run2.2i_dr6` and `dc2_object_run2.2i_dr3`, but not `dc2_object_run2.2i` since it's ambigious.
2. If a data release has a partial release, add `a`, `b`, etc after `drX` (e.g., `dc2_object_run2.2i_dr3a`)
3. Config file for a "partial release (i.e, those with `a` `b` etc), should always specify the list of tracts. 
4. Config file for a regular release (i.e., those without `a` `b` etc) would include _all_ current available tracts. 

As a result, `dc2_object_run2.2i_dr6` and `dc2_object_run2.2i_dr6a` right now will give you exactly the same set of catalogs. But this is _not_because `dc2_object_run2.2i_dr6` is a alias, but because in that object catalog folder, the available tract files is exact the set that are also specified in `dc2_object_run2.2i_dr6a`. 

Later when we add in more tracts, `dc2_object_run2.2i_dr6` will automatically include the new tracts, while `dc2_object_run2.2i_dr6a` will remain what it is right now. 

Does this make sense, @johannct @wmwv? Further thoughts welcome. 

And as part of this change, I also update any deprecation messages that have made a reference to `dc2_object_run2.2i` but without `drX`. And that's why you see a large set of files been updated.

I also updated the README accordingly. It maybe easier to review the [rendered version here](https://github.com/LSSTDESC/gcr-catalogs/blob/u/yymao/add-run22i-dr6/README.md). While I was updating the README, I fixed #394 (sorry, bad habit). 

This PR is now ready for re-review.

yymao:
Yes, the idea is that general users will just use all tracts that are available i.e. `dr6`, and `dr6a` is used only when one needs to track down when provenance is important (i.e. when tracking down "what we did then?" kind of questions).

yymao:
Thank you both @johannct @wmwv !

PR #459, opened 2020-06-09
Remove ambiguous Run 2.1i configs

yymao:
As discussed in #456, I am removing Run 2.x config that do not have a clear DR designation. I tried to avoid breaking backward compatibility, but when preparing the new release, I noticed these three configs are added after v0.17.0, so they all can be removed without breaking backward compatibility (and one of them even has a typo in the catalog name :sweat_smile:).


yymao:
Thank you Michael!

Issue #465, opened 2020-06-26
cosmoDC2_image photo-z composite catalogs often fail to load


sschmidt23:
Due to the large sizes of the photo-z add_on catalogs for cosmoDC2_v1.1.4_image and the need to load the entire catalog because only a subset of the rows have photo-z's, and those are matched by galaxy_id, the catalog often causes memory problems when `photoz_pdf` (an array or 300 floats for every galaxy) is included in the queries.  The matching scheme currently used means that this problem is not solved with return_iterator=True or using a native filter.

A potential solution is to switch the matching scheme to MATCHING_FORMAT and also reading in a pre-computed mask.  This would enable using the iterator and native_filters, and generally improve functionality.

yymao:
Fixed by #466

PR #466, opened 2020-06-30
Issue/465/dc2pzcatalogfix

sschmidt23:
This PR addresses issue #465 where the match by galaxy_id meant that the large photo-z composite catalogs for the extragalactic catalog failed to load if photoz_pdf was selected, even if return_iterator or native_filter were used.  This update switches the matching to MATCHING_FORMAT, even though the photoz catalog has a different number of rows.  To get the correct masked arrays I've computed a `photoz_mask` column which is added to the `photoz_mock_magnitudes_with_errors_for_cosmoDC2_v1.1.4_image` catalog (also loaded in the `cosmoDC2_v1.1.4_image_with_photozs_v1` composite) which will properly mask the arrays.  This fix also fixes the same issue for the FlexZBoost extragalactic photo-z catalog that will be added in the very near future.

In parallel, I am creating an extragalactic photoz tutorial notebook that will show how to use these catalogs and this new mask.

### DIFF
--- a/GCRCatalogs/catalog_configs/photoz_mock_magnitudes_with_errors_for_cosmoDC2_v1.1.4_image.yaml
+++ b/GCRCatalogs/catalog_configs/photoz_mock_magnitudes_with_errors_for_cosmoDC2_v1.1.4_image.yaml
@@ -3,7 +3,6 @@
 subclass_name: photoz_magerr.PZMagErrCatalog
 catalog_name: photoz_mock_magnitudes_with_errors_for_cosmoDC2_v1.1.4_image
 base_dir: ^/xgal/cosmoDC2/addons/photoz_v1.1.4/IMAGE/10_year_error_estimates
-filename_pattern: 'z_(\d)\S+withmask.healpix_(\d+)_magwerr\.h5'
 creators: ['Sam Schmidt']
 description:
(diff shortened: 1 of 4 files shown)


GCRCatalogs/catalog_configs/photoz_mock_magnitudes_with_errors_for_cosmoDC2_v1.1.4_image.yaml
```diff
@@ -3,6 +3,7 @@
 subclass_name: photoz_magerr.PZMagErrCatalog
 catalog_name: photoz_mock_magnitudes_with_errors_for_cosmoDC2_v1.1.4_image
 base_dir: ^/xgal/cosmoDC2/addons/photoz_v1.1.4/IMAGE/10_year_error_estimates
+filename_pattern: 'z_(\d)\S+withmask.healpix_(\d+)_magwerr\.h5'
```
yymao, line 6: Is this line still needed given that you have edited the default `FILE_PATTERN`?

sschmidt23: Ah, good catch, I put that in before remembering I could change the default and forgot to take it out.  I just removed it and pushed that change.

Issue #457, opened 2020-06-04
Add ML photoz for cosmoDC2_v1.1.4_image


sschmidt23:
As an alternative to template-based BPZ, and to have an example of an machine learning based photometric redshift code, add photo-z's computed with FlexZBoost to gcr-catalogs for cosmoDC2_v1.1.4_image (at least the subset with i<26.5, as is done for BPZ)

yymao:
Fixed by #467

PR #467, opened 2020-07-02
Issue/457/add fzboost

sschmidt23:
This PR addresses issue #457 to add FlexZBoost photo-z's for the extragalactic catalog.  As the new files are in parquet rather than hdf5 this required a new addition of a new subclass to the dc2_photoz_parquet file, but it is otherwise fairly straightforward.  The catalogs use the same "trick" of loading using MATCHING_FORMAT as was addressed in issue #465, and functions almost identically to the existing template-based photo-z catalog, so it should be very easy to use for people already familiar with the existing catalogs.


yymao:
@sschmidt23 thanks for the PR. I am trying to understand the new class `DC2PhotozGalaxyCatalog`. The methods in this new class seem to be duplicating methods in `DC2PhotozCatalog` and `CosmoDC2ParentClass`. I don't feel very comfortable with just repeating code. Does it make sense to make `DC2PhotozGalaxyCatalog` a mixin of `DC2PhotozCatalog` and `CosmoDC2ParentClass`? 

I also wonder if you indeed need all the features that `CosmoDC2ParentClass` provides? Or if what's in `PZCalibrateCatalog` sufficient and that `DC2PhotozGalaxyCatalog` should be a mixin of `DC2PhotozCatalog` and `PZCalibrateCatalog`?

sschmidt23:
Yes, I was being lazy and copied several functions while making sure I was following how things worked.  I'll go back and do it with a mixin class, which is a much better idea.

sschmidt23:
I separated out a couple of duplicate functions into a simple mixin class and made it so that DC2PhotozGalaxyCatalog inherits from CosmoDC2ParentClass, which saves some duplication.  There are still a couple of functions that are very similar (with slight changes due to healpix/reshift_lower vs tract indexing) to functions in DC2DMCatalog, but I didn't see an easy way to have the multiple inheritance work properly between the two classes and with the tract vs healpix difference, so those functions are still present in DC2PhotozGalaxyCatalog.

PR #469, opened 2020-07-12
Refactor dc2_photoz_parquet.py

yymao:
@sschmidt23 I've attempted to refactor `dc2_photoz_parquet.py` (as part of the review for #457). 

It turns out my suggestion of subclassing `CosmoDC2ParentClass` was not great; as it has too much custom methods specific for cosmoDC2. Since you are using parquet files, I found that simply subclassing `DC2DMCatalog` and implementing the healpix/zlo native filters work well. 

Hence I created a new `CosmoDC2Parquet` that is a subclass of `DC2DMCatalog`. `CosmoDC2Parquet` is similar to `DC2DMTractCatalog` but use healpix/zlo instead of tract. With this new class, `DC2PhotozGalaxyCatalog` is then just a very simple mixin of `DC2PhotozMixin` and `CosmoDC2Parquet`. 

I have not tested my implementation since NERSC is down.

A side note for @JoanneBogart --- When implementing `CosmoDC2Parquet` I was hoping to use the new `PathInfoParser` but not sure if it works for this case. In particular, the filename pattern is `some_prefix_z_(\d)_(\d).step_all.healpix_(\d+).parquet`, but I want to call the first integer `redshift_block_lower`, ignore the second integer, and call the third integer `healpix_pixel`. I'm not sure if I can customize `PathInfoParser` to this extend?

### DIFF
--- a/GCRCatalogs/catalog_configs/photoz_fzboost_for_cosmoDC2_v1.1.4_image.yaml
+++ b/GCRCatalogs/catalog_configs/photoz_fzboost_for_cosmoDC2_v1.1.4_image.yaml
@@ -2,9 +2,7 @@
 
 subclass_name: dc2_photoz_parquet.DC2PhotozGalaxyCatalog
 base_dir: ^/xgal/cosmoDC2/addons/photoz_v1.1.4/FlexZBoost
-catalog_root_dir: /global/cscratch1/sd/schmidt9/CosmoDC2v114_image/FlexZPipe/outputs
-healpix_pattern: fzboost_photoz_pdf_z_{}_{}.step_all.healpix_{}.parquet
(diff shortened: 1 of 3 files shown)


yymao:
Thanks for testing this and fixing the bugs @sschmidt23 !

JoanneBogart:
@yymao Concerning PathInfoParser:  you're right; it doesn't (yet) allow for that degree of customization.  Right now there is a small set of  names (including `healpix` but not `healpix_pixel`) which are known to have format \d+ and values for anything else enclosed in braces { }  is just expected to have the form of an alphanumeric name.   It's also possible to just put in the full re syntax for named groups yourself. Any named groups will be treated as native filters. (Well, that's the intent.  I don't think I tested it.)  But maybe there should be a simplified way to do this without having to resort to the full arcane re syntax.   I can make a separate issue for this.

PR #478, opened 2020-08-04
faster access to number of rows for parquet files

### LINKED ISSUE #477: len(catalog) fails for Parquet catalog dc2_object_run2.2i_dr6c
I get the error below when trying to find the length of the catalog `dc2_object_run2.2i_dr6c`.  
Might just be a missing pair of brackets?

```
In [1]: import GCRCatalogs                                                                                                                     

In [2]: print(GCRCatalogs.__version__)                                                                                                         
0.19.0

In [3]: cat = GCRCatalogs.load_catalog("dc2_object_run2.2i_dr6c")                                                                              

In [4]: print(len(cat))                                                                                                                        
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-7c633736d7f4> in <module>
----> 1 print(len(cat))

/opt/lsst/software/stack/python/miniconda3-4.7.10/envs/lsst-scipipe-4d7b902/lib/python3.7/site-packages/GCRCatalogs/dc2_dm_catalog.py in __len__(self)
    269         if self._len is None:
    270             # pylint: disable=attribute-defined-outside-init
--> 271             self._len = sum(len(dataset) for dataset in self._datasets)
    272         return self._len
    273 

/opt/lsst/software/stack/python/miniconda3-4.7.10/envs/lsst-scipipe-4d7b902/lib/python3.7/site-packages/GCRCatalogs/dc2_dm_catalog.py in <genexpr>(.0)
    269         if self._len is None:
    270             # pylint: disable=attribute-defined-outside-init
--> 271             self._len = sum(len(dataset) for dataset in self._datasets)
    272         return self._len
    273 

/opt/lsst/software/stack/python/miniconda3-4.7.10/envs/lsst-scipipe-4d7b902/lib/python3.7/site-packages/GCRCatalogs/parquet.py in __len__(self)
     55 
     56     def __len__(self):
---> 57         return int(self.handle.scan_contents)
     58 
     59     def __contains__(self, item):

TypeError: int() argument must be a string, a bytes-like object or a number, not 'method'
```

yymao:
This PR adapts the method suggested by @joezuntz in #477. Fix #477. 

@JoanneBogart can you check if it works as expected for catalogs that contain row groups?

### DIFF
--- a/GCRCatalogs/parquet.py
+++ b/GCRCatalogs/parquet.py
@@ -54,7 +54,7 @@ class ParquetFileWrapper():
         self._handle = None
 
     def __len__(self):
-        return int(self.handle.scan_contents)
+        return self.handle.metadata.num_rows
 
     def __contains__(self, item):
         return item in self.columns


JoanneBogart:
It gives the right answer and is maybe a little faster than the original method.  I'm surprised it isn't much faster, though.

yymao:
Thanks, @JoanneBogart !

PR #480, opened 2020-08-05
speed up len calculation for dc2_truth_parquet

JoanneBogart:
Follow-on to issue #477
The dc2_truth_parquet reader was not taking advantage of the improvement to parquet.py. This PR fixes that.


JoanneBogart:
Thanks, @yymao !    I was about to ask you to review but you're too fast for me :-)

Issue #479, opened 2020-08-05
Add reader for random forest photo-z's for subset of cosmoDC2_v1.1.4_small


sschmidt23:
@MarkusMichaelRau asked about creating a catalog reader for a catalog of photo-z's created using a random forest algorithm for a subset of cosmoDC2_v1.1.4_small.  The catalog has a slightly different set of columns and is all stored in a single parquet file, so it will need a few tweaks compared to the nominal setup for the other extragalactic catalog readers.

yymao:
Closed by #481

PR #481, opened 2020-08-20
add reader for rand forest PZ for CSDC2_114_small

sschmidt23:

This PR addresses issue #479 to make a reader for the subset of cosmoDC2_v1.1.4_small galaxies with photo-z's generated by Olivia/Markus.  Added a new subclass to deal with the unique quantity modifiers and single file storage.  Things seem to work properly, and were tested by Olivia for functionality.  

Right now the data file is specified with a hard link in the config yaml and resides in the PZ directories at NERSC.  We can either keep this, or copy the file to the shared directory and update to a relative path, not sure which is better.

### DIFF
--- a/GCRCatalogs/catalog_configs/photoz_SKRF_v1_for_cosmoDC2_v1.1.4_small.yaml
+++ b/GCRCatalogs/catalog_configs/photoz_SKRF_v1_for_cosmoDC2_v1.1.4_small.yaml
@@ -1,6 +1,6 @@
 # test for olivia photo-z's
 
 subclass_name: dc2_photoz_parquet.PZSKRFCatalog
-base_dir: /global/cfs/cdirs/lsst/shared/xgal/cosmoDC2/addons/photoz_v1.1.4/10_year_error_estimates/OLIVIA_WORK
+base_dir: ^/xgal/cosmoDC2/addons/photoz_v1.1.4/10_year_error_estimates/OLIVIA_WORK
 filename: final-result.parquet
 addon_for: cosmoDC2_v1.1.4_small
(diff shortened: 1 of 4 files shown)


sschmidt23:
Ok, I created a request to copy the file.  Agreed on the proliferation of readers, I think we've settled on parquet files, and I'm hoping that all new catalogs will follow the pattern (I plan to update some of the older catalogs to the parquet format and "retire"/delete those outdated readers, but it hasn't been a high priority since nothing is actually broken currently).

sschmidt23:
Ok, I've changed the path to the newly copied file in the shared area.  I also had to delete a pylint disable that appears to have been removed (google search showed some bug reports for `pylint: disable=C0330`) so that pylint checks would pass.  
I think things should be ready to go now.

GCRCatalogs/catalog_configs/photoz_SKRF_v1_for_cosmoDC2_v1.1.4_small.yaml
```diff
@@ -0,0 +1,6 @@
+# test for olivia photo-z's
+
+subclass_name: dc2_photoz_parquet.PZSKRFCatalog
+base_dir: /global/cfs/cdirs/lsst/shared/xgal/cosmoDC2/addons/photoz_v1.1.4/10_year_error_estimates/OLIVIA_WORK
```
yymao, line 4: Need to use the root_dir symbol here: 
```suggestion
base_dir: ^/xgal/cosmoDC2/addons/photoz_v1.1.4/10_year_error_estimates/OLIVIA_WORK
```

sschmidt23: Whoops, forgot to modify to the relative path, made the suggested change.

Issue #475, opened 2020-07-18
Allow accessing add-on attributes in photo-z composite catalogs 


yymao:
Some of the photo-z composite catalog configs have the line
```yaml
only_use_master_attr: true
```
However, sometimes it is desirable to access attributes in the add-on catalogs, not just the ones in the main catalog. In those case, this option should be removed from the config files. See e.g., https://github.com/LSSTDESC/DC2-analysis/pull/145#discussion_r456829248.  

@sschmidt23 can you remove this line when appropriate?

yymao:
I think this has been fixed. @sschmidt23 if it's not the case please reopen.

sschmidt23:
I just checked, there are two files that still have this issue (and one with an unnecessary only_use_master_attr: False), making a PR to fix this now.

PR #485, opened 2020-09-03
remove only_use_master_attr from 3 pz files

sschmidt23:
This PR addresses issue #475 and is a simple deletion of the "only_use_master_attr" parameter from three photo-z files.  Setting this prevented the @property photoz_pdf_bin_centers from being accessible in two of the catalogs (third one set it to false, which is the default, so the line is unnecessary).

Tested functionality in the branch, everything works as expected.


yymao (approved):
Thanks @sschmidt23

PR #488, opened 2020-10-05
Set base_dir to Run2.2i-dr6-wfd-v1 instead of dr6d

### LINKED ISSUE #487: Replace metacal reference to Run2.2i dr6d in favor of Run2.2i-dr6-wfd-v1
The base directory for the Run2.2i dr6-wfd-v1 metacal catalogs are pointing to a non-existent dr6d directory. We never made a dr6d release, rather we made a dr6-wfd-v1 release.

As discussed on Slack: https://lsstc.slack.com/archives/C3WLSCNKA/p1601902355012600

heather999:
fix #487 
The metacal base directory was pointing to dr6d, which was never released - we instead moved to dr6-wfd-v1.  The `shared` area at NERSC already reflects this directory structure and to avoid errors when using GCRCatalogs, we need to keep the directories referenced consistent.


yymao (approved):
Thanks, @heather999

PR #490, opened 2020-10-27
Fix unmatched tracts in dc2_object_run2.2i_dr6_wfd_v1_with_metacal

yymao:
@fjaviersanchez noticed this issue. This PR attempts to fix it.

PR #491, opened 2020-10-27
Propose a bug-fix release v1.0.2

yymao:
PR #490 fixed a bug in the DR6 WFD metacal catalog config file. Hence I am proposing a bug fix release so that this fix can propagate downstream.


heather999 (approved):
Looks good - and thanks!

PR #495, opened 2020-11-24
Add dc2 Run2.2i truth catalogs to default list

JoanneBogart:
Run 2.2i truth catalogs belong in the default list.  Up till now only truth from Run 1.2 was showing up there.


yymao (approved):
LGTM! Thanks @JoanneBogart

PR #503, opened 2020-12-06
Utility for setting local site

### LINKED ISSUE #499: Enhance protocols for site, root_dir variables to better support users not at a standard site
Currently a warning is issued if the site variable isn't set, but that is about to be normal for many users.  We need an easy way to disable the warning under some circumstances.

JoanneBogart:
fix #499 
Add script to update value for local_user site


JoanneBogart:
The branch name is now somewhat misleading.  New code will retrieve user's default root_dir from config file if there is one.  From within the package one can optionally write root_dir to a config file or remove it.   A new script may be invoked from the command line to create or update the config file.

GCRCatalogs/register.py
```diff
@@ -7,11 +7,11 @@
...
 import socket
 from GCR import BaseGenericCatalog
-from .utils import is_string_like
+from .utils import is_string_like, get_config_dir
 
 __all__ = [
-    "get_root_dir", "set_root_dir", "reset_root_dir", "get_available_catalogs",
-    "get_reader_list", "get_catalog_config", "has_catalog", "load_catalog", "retrieve_paths", "get_site_list", "set_root_dir_by_site"]
+    "get_root_dir", "set_root_dir", "remove_root_dir_default", "reset_root_dir", "get_available_catalogs",
+    "get_reader_list", "get_catalog_config", "has_catalog", "load_catalog", "retrieve_paths", "get_site_list", "set_root_dir_by_site", "write_to_user_config", "remove_from_user_config", "remove_root_dir_default"]
```
yymao, line 14: `remove_root_dir_default` repeated above.
```suggestion
    "get_reader_list", "get_catalog_config", "has_catalog", "load_catalog", "retrieve_paths", "get_site_list", "set_root_dir_by_site", "write_to_user_config", "remove_from_user_config"]
```

GCRCatalogs/register.py
```diff
@@ -77,6 +79,63 @@ def load_catalog_from_config_dict(catalog_config):
...
+
+    if os.path.exists(config_path):
+        config_dict = load_yaml_local(config_path)
+        if not overwrite:
+            if not config_dict.keys().isdisjoint(d.keys()):
+                warnings.warn("Overwrite condition violated; config file not updated")
+                return None
+
+    for k,v in d.items():
+        config_dict[k] = v
```
yymao, line 109:
```suggestion
    config_dict = dict(d)
```

GCRCatalogs/register.py
```diff
@@ -101,17 +160,30 @@ class RootDirManager:
...
         self._site_config = {}
         self._site_info = self._get_site_info()
 
+        # User config has highest priority
+        if self._user_config_path:
+            if os.path.exists(_USER_CONFIG_PATH):
+                d = load_yaml_local(_USER_CONFIG_PATH)
+                if 'root_dir' in d.keys():
+                    self._default_root_dir = d['root_dir']
+                    return
```
yymao, line 185:
```suggestion
        if self._user_config_path and os.path.isfile(self._user_config_path):
            d = load_yaml_local(_USER_CONFIG_PATH)
            if 'root_dir' in d:
                self._default_root_dir = d['root_dir']
                return
```

GCRCatalogs/register.py
```diff
@@ -558,20 +631,29 @@ def get_root_dir():
     return _config_register.root_dir
 
 
-def set_root_dir(path):
+def set_root_dir(path, write_to_config=False):
```
yymao, line 634:
```suggestion
def set_root_dir(path, write_to_config=False):
```

GCRCatalogs/utils.py
```diff
@@ -34,6 +34,22 @@ def first(iterable, default=None):
     """
     return next(iter(iterable), default)
 
+  
+def get_config_dir(create=False):
+    if os.getenv("XDG_CONFIG_HOME"):                   # Unix
+        user_config_dir = os.getenv("XDG_CONFIG_HOME")
```
yymao, line 40:
```suggestion
        user_config_dir = os.getenv("XDG_CONFIG_HOME")
```

yymao (changes requested):
@JoanneBogart I haven't finished the full review, but made some inline comments on coding style first.

Generally I think I like this direction. I need to think about the script name and also whether to put it in `scripts` or directly in `GCRCatalogs`. I'll come back to this later.

In the meantime, I wonder what you think about moving `write_to_user_config`, `remove_from_user_config`, and `get_config_dir` into another module, say, `user_config.py` -- the first two are currently in `register.py` but they are really more generic function. The last one is in `utils.py` but it seems more suitable to be grouped with the other too. 

I actually these three functions probably should be a class, so that we don't need to keep passing `_USER_CONFIG_PATH` or `_USER_CONFIG_NAME` around. 

```python
class UserConfigManager():
    def __init__(self, config_name):
        ...
	def _get_config_dir(...): 
        ...
	def write_key():
        ...
    def remove_key():
        ....
```

I'll be happy to help refactor this if you agree with the design.

GCRCatalogs/register.py
JoanneBogart: config_dict might already have contents if there was an old file.  In that case I want to add the values from d, not discard old contents of config_dict

GCRCatalogs/register.py
JoanneBogart: I'd like to keep the nested `if` but I'll incorporate the other suggestions in my next push.

GCRCatalogs/register.py
yymao: Ah, in that case, use `config_dict.update(d)`, which would make your intend very clear.

GCRCatalogs/register.py
yymao: OK. BTW, the `_USER_CONFIG_PATH` within the call `load_yaml_local` should probably be `self._user_config_path` as well? 

But I think this would be easier once we implement the user config class.

JoanneBogart: Yes, I noticed that and changed it for the user config class.

JoanneBogart:
@yymao I've committed a new version of UserConfigManager which uses MutableMapping as a base class.  The implementations of MutableMapping methods are nearly identical to yours.  There are also _load_config and _write_config routines which are similar but not identical to yours, and I added several routines.  Here is a more detailed list of differences, followed by reasons.

__init__
* I made a different choice of what to save as instance variables, keeping exactly those things which were needed in other routines.
* The restriction on config_dirname to a single level seemed arbitrary.  My version takes the input argument as is, except it must not be an absolute path.  I don't feel strongly about this.  If it's standard practice that there is at most one subdirectory level under ~/.config I'll change my implementation

_load_config
Instead of using try...except I check first to see if the file exists 

_write_config
My version always writes something.  If the input argument is None or an empty dict it writes a file with contents
{}
In all cases it will overwrite an existing file.  I used yaml.dump rather than yaml.safe_dump because the data to be written is always a dict; I don't think the extra check is necessary.

Extra methods:
_user_config_exists
Convenience

get
like dict get. Returns None if key is not found

pop
similar to dict pop. Remove a single item from config and return value if it was there. Otherwise return sentinel value (default None)

setitems
update config with user-supplied dict

deleteitems
update dict by removing specified items
--------------
Many of the differences have to do with avoiding raising exceptions under "normal" circumstances. I added get and pop because __getitem__ and __delitem__ didn't have the behavior I wanted for a client like RootDirManager.  In particular, don't raise exceptions if
* nothing is actually wrong (no user error; no transient system problem) 
      and
* the situation is not especially uncommon

I added setitems and deleteitems (maybe should change that to delitems?) because, if there were several items to be added or deleted, it would be very inefficient to repeatedly call the corresponding single-item routines, reading in and writing out the config file each time.

JoanneBogart:
@yymao  I will need to make some further small changes in order to support the script --overwrite option (in particular when it's false), as well as modifying and renaming the script itself. You might want to wait for a new commit before taking a look.

scripts/manage_root_dir_config.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+
+"""
+Add, modify or remove value for root_dir in config file.
+root_dir
+"""
+import os
+from argparse import ArgumentParser, RawTextHelpFormatter
+from GCRCatalogs import write_to_user_config, remove_root_dir_default
```
yymao, line 9: `write_to_user_config` no longer exists I think
```suggestion
from GCRCatalogs import set_root_dir, remove_root_dir_default
```

GCRCatalogs/register.py
```diff
@@ -558,20 +584,29 @@ def get_root_dir():
...
     Sets runtime root_dir to *path*.
     """
     _config_register.root_dir = path
 
+    if write_to_config:
+        _config_register.persist_root_dir()
+
 
-def set_root_dir_by_site(site):
+def set_root_dir_by_site(site, write_to_config=False):
```
yymao, line 597: `write_to_config` not used
```suggestion
def set_root_dir_by_site(site):
```

scripts/manage_root_dir_config.py
```diff
@@ -0,0 +1,38 @@
...
+    args = parser.parse_args()
+
+    if args.remove_path:
+        if args.root_dir is not None:
+            print("Conflicting options --root-dir and --remove-path")
+            return
+        remove_root_dir_default()
+    else:
+        write_to_user_config({'root_dir' : os.path.abspath(args.root_dir)},
+                             overwrite=args.override)
```
yymao, line 32:
```suggestion
        set_root_dir(os.path.abspath(args.root_dir), write_to_config=True)
```

See also my general comment on this script.

yymao (changes requested):
Thanks @JoanneBogart --- I made a bunch suggestions. Mostly minor, but some was about new methods that were introduced but I am uncertain about their use cases. 

I have some comment on the script.  I would try to avoid adding things to users' path if possible. One option is to just use module-level script, and the user can do something like:

```bash
python -m GCRCatalogs.user_config set root_dir [new_path]
python -m GCRCatalogs.user_config get root_dir
python -m GCRCatalogs.user_config del root_dir
```
Using `set`, `get`, `del` also make it more general in case of future expansion. Thoughts?

yymao:
@JoanneBogart --- Sorry, I think I have missed your comment above! I went into file review mode directly. So please ignored some of my inline comments asking for use cases, etc. 

I agree that `get` and `pop` are useful. The nice thing of using `collections.abc` is that these standard methods are already implemented once you implement `__getitem__`, `__setitem__`, and `__delitem__` so you don't need to reimplement them. 

I am not sure about `setitems` and `deleteitems`. Regular python dictionary does not have these methods, and I don't see a compelling use case that these methods will be called so often that having them is useful. In fact, we didn't call these methods at all in our own use of `UserConfigManager`. 

Your choices on other things sounds fine to me.  I did make some comments on the script API.

yymao:
Sorry, some message crossing! 

I actually would not ask user to supply a `--override` flag to the config setting script, as it is not a common practice with most config setting commands. 

My suggestion would be to print out both old (existing) value and new value when the config setting script is called.

GCRCatalogs/register.py
```diff
@@ -176,6 +189,19 @@ def set_root_dir_by_site(self, site):
             site_string = ' '.join(_config_register.site_list)
             warnings.warn(f"Unknown site '{site}'.\nAvailable sites are: {site_string}\nroot_dir is unchanged")
 
+    def persist_root_dir(self):
+        """
+        Write current root_dir to user config
+        """
+        self._user_config_manager.__setitem__(_ROOT_DIR_KEY, os.path.abspath(self.root_dir))
```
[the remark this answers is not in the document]
JoanneBogart: Thanks for this.  I didn't realize the [ ] syntax was automatically available.

GCRCatalogs/utils.py
```diff
@@ -34,6 +34,22 @@ def first(iterable, default=None):
...
+    else:
+        user_config_dir = os.path.join(os.path.expanduser("~"), ".config")
+
+    desc_config_dir = os.path.join(user_config_dir, "lsstdesc")
+
+    if create:
+        os.makedirs(desc_config_dir, exist_ok=True)
+
+    return desc_config_dir
+
```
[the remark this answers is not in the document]
JoanneBogart: Yep.  Forgot to do this.

JoanneBogart:
@yymao Concerning delitems, setitems:  it's true they're not needed currently because there is only one item to deal with, but if we did expand what's in the config file, they would be an optimization - better to load and write the file just once rather than once per item to be modified.   They aren't helpful for regular dictionaries because there is no extra cost associated with loading and writing.    

Since we don't need them now, we could put off implementing them in this release, though I don't see any harm in leaving them in.

I like the idea of a module-level script.  My only goal is to be able to update the config at the command line, which this does perfectly well, and without the associated messiness of a separate script which has to be kept in a separate place included in PATH.

Since you had said the interface for the script I had looked ok, I had left --override in.  If we don't need it that will simplify some of the lower level stuff it depended upon.  I agree about returning the old value.  That will make it easy for the user to put back the old value if they want to.

yymao:
Good point about reducing the number of file writing. In that case I'd still suggest we overwrite the `update` method (and remove `setitems`). We can keep `delitems`. 

I think it's ok to leave out `--override`. Print out old value is more useful in practice, and serves the need to protecting accidental overwrite.

JoanneBogart:
Yes, re-implementing `update` instead of `setitems` sounds good. 
In the collections.abc documentation https://docs.python.org/3.8/library/collections.abc.html the table shows that MutableMapping inherits pop, etc., from Mapping, but I haven't been able to find
documentation of the call interface.   I looked at dict as an example, but surely it's described somewhere for Mapping. Do you happen to know where?

yymao:
Not that I know of on the official python website. If you use ipython you can run

```python
from collections.abc import MutableMapping
MutableMapping.pop??
```

JoanneBogart:
Actually, I misread.  It says MutualMapping gets Mixin Methods from Mapping  and *also* pop, etc.   But I have the same question: where is the interface for these methods documented?   Similarly for the [ ] syntax.  How would I learn about that (other than you pointing it out :-)  )?

yymao:
So `d.__setitem__(k, v)` means `d[k] = v` can be found at https://docs.python.org/3/reference/datamodel.html#object.__setitem__

In the opening paragraph of the link above ([here](https://docs.python.org/3/reference/datamodel.html#emulating-container-types)) it also mentioned `pop`, `update`, etc, but not in great details... I myself learned those by watching other people use them too!

JoanneBogart:
> So d.__setitem__(k, v) means d[k] = v can be found here

thanks, @yymao  !

JoanneBogart:
This is more or less what I was looking for:
https://docs.python.org/3.8/reference/datamodel.html#emulating-container-types

yymao:
Yep (that's the second link I posted above I think?)

JoanneBogart:
> Yep (that's the second link I posted above I think?)
Sure is!

I'm more or less done with revising user_config.py and register.py; module-level script is the task for today.

JoanneBogart:
There seems to be a problem with a module-level script which is not in the top-level module.  The script does the right thing, but there is a RuntimeWarning:

> $ python -m GCRCatalogs.user_config get root_dir
/Users/jrb/miniconda3/lib/python3.8/runpy.py:127: RuntimeWarning: 'GCRCatalogs.user_config' found in sys.modules after import of package 'GCRCatalogs', but prior to execution of 'GCRCatalogs.user_config'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
Value of root_dir: /Users/jrb/desc_share

The final line is the expected output.   
I don't think there is any way to turn off that warning inside the code since the problem happens before it's run (and even if it were possible it doesn't seem like a good idea).

This works:

> $ python -W ignore -m GCRCatalogs.user_config get root_dir
Value of root_dir: /Users/jrb/desc_share

but somewhat defeats the purpose of providing a simple interface.

GCRCatalogs/user_config_mgr.py
```diff
@@ -0,0 +1,127 @@
...
+        Returns
+        -------
+        dict written to the file.  
+        If overwrite condition is violated, raise ValueError
+
+        """
+        config_dict = {}
+
+        if self._user_config_exists():
+            config_dict = self._load_config()
```
yymao, line 62: `_user_config_exists` is already checked in `_load_config1
```suggestion
        config_dict = self._load_config()
```

GCRCatalogs/user_config_mgr.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+
+        config_dict = self._load_config()
+
+        keys_to_try = keys
+        if absent_ok:
+            keys_to_try = set(keys).intersection(config_dict.keys())
+
+        for k in keys_to_try:
+            del config_dict[k]
```
yymao, line 86: Right now when `absent_ok` is set to false, some keys may be deleted (the ones before the first missing key), while other will not. I am not sure if this is what a user would expected, or if it has a common use case. I would suggest that we do no action if `absent_ok` is set to false and not all keys are in `config_dict`.

```suggestion
        if not absent_ok and not all(key in config_dict for key in keys):
        	raise KeyError("Some keys do not exist")

        for key in keys:
            config_dict.pop(key, None)
```

GCRCatalogs/utils.py
```diff
@@ -34,7 +34,7 @@ def first(iterable, default=None):
     """
     return next(iter(iterable), default)
 
-
+
```
yymao, line 37:
```suggestion

```

yymao (commented):
@JoanneBogart Thanks for getting this close to ready! I've added some more comments to your recent updates.

GCRCatalogs/user_config.py
```diff
@@ -0,0 +1,40 @@
...
+    args = parser.parse_args()
+
+    umgr = UserConfigManager()
+    
+    if args.operation == 'get':
+        old = umgr.get(args.key)
+        if old:
+            print(f"Value of {args.key}: {old}")
+        else:
+            print(f"{args.key} is not in the user config")
```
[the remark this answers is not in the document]
JoanneBogart: The reason I used the less idiomatic form is because it avoids reading in the config file an extra time. Alternatively UserConfigManager could keep a cached copy of the file, but that seems risky since a session lasts an indeterminate time and some other process can meanwhile modify the file.
I don't have an opinion about supporting Python < 3.6.    I can change those lines if you think it's advisable, but I see in setup.py

> 'Programming Language :: Python :: 3.6',

so maybe it's not an issue.

GCRCatalogs/user_config.py
```diff
@@ -0,0 +1,40 @@
...
+    if args.operation == 'del':
+        old = umgr.pop(args.key, None)
+        if old:
+            print(f"{args.key} deleted from user config")
+        else:
+            print(f"{args.key} was not in user config")
```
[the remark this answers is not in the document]
JoanneBogart: I meant to print out the old value but it seems I forgot. I'll add it.

GCRCatalogs/user_config.py
```diff
@@ -0,0 +1,40 @@
...
+if __name__ == "__main__":
+    usage="""Directly manipulate items in user config"""
+
+    parser = ArgumentParser(description=usage,
+                            formatter_class=RawTextHelpFormatter)
+    parser.add_argument("operation", help="Operation to be performed for a particular key",
+                        choices=("set", "get", "del"))
+    parser.add_argument("key", help="Key on which operation is to be performed")
+    parser.add_argument("value", help="Value to use when operation is 'set'",
+                        default=None, nargs="?")
```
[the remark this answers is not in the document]
JoanneBogart: Thanks for the tip!  I'll look into it.

GCRCatalogs/user_config.py
```diff
@@ -0,0 +1,40 @@
...
+            print(f"{args.key} is not in the user config")
+    if args.operation == 'del':
+        old = umgr.pop(args.key, None)
+        if old:
+            print(f"{args.key} deleted from user config")
+        else:
+            print(f"{args.key} was not in user config")
+    if args.operation == 'set':
+        if not args.value:
+            raise ValueError("Must supply value for 'set' operation")
```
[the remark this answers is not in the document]
JoanneBogart: Commit this for now.  May change to subparser implementation.

GCRCatalogs/user_config.py
```diff
@@ -0,0 +1,40 @@
...
+        else:
+            old = umgr.get(args.key)
+            umgr[args.key] = args.value
+            if old:
+                print(f"New value set. Old value was {old}")
+            else:
+                print(f"New value set. No old value")
```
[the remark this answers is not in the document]
JoanneBogart: I'll a print for the new value.

GCRCatalogs/user_config_mgr.py
JoanneBogart: Thanks for catching this.  Indeed that behavior is not what anyone would expect.

JoanneBogart:
I tried using a subparser.  It's better in that the value argument needed by set is handled automatically, but the command-line help is less informative.  Just calling it with --help doesn't print out anything about subcommand arguments and I don't see any other way from the command line to get that help.  I could include description of usage, etc. in the help option to add_parser but that doesn't seem like a win to me.

GCRCatalogs/user_config_mgr.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Parameters
+        ----------
+        A collection of keys
+        """
+
+        config_dict = self._load_config()
+
+        if not absent_ok and not all(key in config_dict for key in keys):
+        	raise KeyError("Some keys do not exist")
```
yymao, line 79: Always use spaces
```suggestion
            raise KeyError("Some keys do not exist")
```

GCRCatalogs/user_config_mgr.py
```diff
@@ -0,0 +1,123 @@
...
+        reldir = config_reldir or self._default_config_reldir
+        if os.path.isabs(reldir):
+            raise ValueError(f'{reldir} must be a relative path')
+        self._user_config_dir = os.path.join(self._get_config_home(), reldir)
+        self._user_config_path = os.path.join(self._user_config_dir, config_name)
+
+    @staticmethod
+    def _get_config_home():
+        if os.getenv("XDG_CONFIG_HOME"):                   # Unix
+            return(os.getenv("XDG_CONFIG_HOME"))
```
yymao, line 28: Removing trailing tabs
```suggestion
            return(os.getenv("XDG_CONFIG_HOME"))
```

GCRCatalogs/register.py
```diff
@@ -558,12 +590,22 @@ def get_root_dir():
...
+    -------
+    If write_to_config, return old persisted value (which may be None);
+    else None.
     """
     _config_register.root_dir = path
 
+    if write_to_config:
+        return _config_register.persist_root_dir()
+    else:
+        return None
```
yymao, line 607: A function that does not encounter any return statement will automatically returns None

```suggestion
```

GCRCatalogs/register.py
```diff
@@ -176,6 +186,28 @@ def set_root_dir_by_site(self, site):
             site_string = ' '.join(_config_register.site_list)
             warnings.warn(f"Unknown site '{site}'.\nAvailable sites are: {site_string}\nroot_dir is unchanged")
 
+    def persist_root_dir(self, value=None):
+        """
+        Write root_dir to user config. By default write current session value
+        Returns
+        -------
+        old value (may be None)
```
yymao, line 194: Returning old value here is a somewhat unexpected behavior. Usually a function that sets something would either return nothing, or returns the new value that is being set. 

Is there a particular reason that this function must return the old value? If not, or if it can be achieved differently, I would suggest that this function either returns `None` or `value`.

GCRCatalogs/register.py
```diff
@@ -558,12 +590,22 @@ def get_root_dir():
...
 
-def set_root_dir(path):
+def set_root_dir(path, write_to_config=False):
     """
     Sets runtime root_dir to *path*.
+
+    Returns
+    -------
+    If write_to_config, return old persisted value (which may be None);
+    else None.
```
yymao, line 600: See my earlier comment on `persist_root_dir` -- I think a setter function should return None or the new value. Returning the old value would be a rather unexpected behavior.

GCRCatalogs/user_config.py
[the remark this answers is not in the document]
yymao: Re: using `in` --- I am ok with the current implementation as it serves the common use cases, but perhaps we should be accurate in the print out message and say:

```python
     f"{args.key} is not in the user config or its value is None"
```

Re: f-string --- I am ok with dropping support for Python < 3.6. I just want to make sure we are making a deliberate decision rather than making it an implied consequence.

GCRCatalogs/user_config.py
```diff
@@ -0,0 +1,42 @@
...
+            print(f"Value of {args.key}: {old}")
+        else:
+            print(f"{args.key} is not in the user config")
+    if args.operation == 'del':
+        old = umgr.pop(args.key, None)
+        if old:
+            print(f"{args.key} deleted from user config")
+            print(f"Old value was {old}")
+        else:
+            print(f"{args.key} was not in user config")
```
yymao, line 32: Same reason as above
```suggestion
            print(f"{args.key} was not in user config or its value was None.")
```

GCRCatalogs/user_config.py
```diff
@@ -0,0 +1,42 @@
...
+    if args.operation == 'set':
+        if not args.value:
+            raise parser.error("Must supply value for 'set' operation")
+        else:
+            old = umgr.get(args.key)
+            umgr[args.key] = args.value
+            if old:
+                print(f"New value {args.value} was set. Old value was {old}")
+            else:
+                print(f"New value {args.value} was set. No old value")
```
yymao, line 42: Same reason as above
```suggestion
                print(f"New value {args.value} was set. No old value or old value was None")
```

GCRCatalogs/user_config.py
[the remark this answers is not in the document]
yymao: It's ok to drop it for now.

yymao:
@JoanneBogart I think this is really close to ready. I made some minor comments on the print out messages, and more comments on spaces/tabs :slightly_smiling_face: 

I think the only major comment I still have is the behavior of `persist_root_dir` that returns old value. I think this behavior is rather unexpected and not actually used in other parts of the code. I'd suggest removing the return.

GCRCatalogs/register.py
JoanneBogart: I think it might have been left over from the need to display the old value for script users.  Since that is now handled by a separate module, there is no need for routines inside register to do this.

GCRCatalogs/register.py
```diff
@@ -176,6 +185,23 @@ def set_root_dir_by_site(self, site):
             site_string = ' '.join(_config_register.site_list)
             warnings.warn(f"Unknown site '{site}'.\nAvailable sites are: {site_string}\nroot_dir is unchanged")
 
+    def persist_root_dir(self, value=None):
+        """
+        Write root_dir to user config. By default write current session value
+        """
+        old = self._user_config_manager.get(self._ROOT_DIR_KEY)
+        to_write = value or self.root_dir
+        self._user_config_manager[self._ROOT_DIR_KEY] = os.path.abspath(to_write)
```
yymao, line 194: `old` is no longer needed. Add one blank line between class methods. 

One question --- should we allow the `value` argument? And if so, should we also set `self.root_dir = value` when `value` is not `None`? Given the function name `persist_root_dir`, I feel that allowing one to set the config to a value that is different from current `self.root_dir` seems a bit confusing?

yymao: Given how this function is used in `set_root_dir`, I would suggest removing the `value` argument.

JoanneBogart: The behavior is consistent with `unpersist_root_dir`, which only modifies the user config but leaves the session value alone.  I don't see harm in leaving the `value` argument there, even if it's not currently used, but if we change `persist_root_dir` I think we also need to reconsider `unpersist_root_dir`

yymao: I think `unpersist_root_dir`'s behavior is expected as the function name suggests --- i.e., keep the current running session intact but remove the persistent effect, and it does not result in any conflicting settings. 

When `persist_root_dir` is called without `value`, its behavior is also expected (i.e., write current `root_dir` to config). But when `value` is supplied, I would expect its behavior to be similar to when `value` is not set (i.e, write current `root_dir` to config), and hence would expect it to first set current `root_dir`. If it doesn't, the the current session's `root_dir` would be in conflict with the value in config, which seems an odd situation. 

In particular, since `persist_root_dir` is a method of `RootDirManager` --- what is the use case when the user would want to use `RootDirManager` to set config but not the current session? 

If there is a clear use case, I would suggest making a different method for it, rather than overloading `persist_root_dir`.

JoanneBogart: > I think unpersist_root_dir's behavior is expected as the function name suggests --- i.e., keep the current running session intact but remove the persistent effect, and it does not result in any conflicting settings.

I would say that is a conflict:  the persisted value is None and the session value isn't.  But as a practical matter for now, the `value` argument is unused so I will make the change.  We can revisit it when and if there is a need for it.

GCRCatalogs/utils.py
```diff
@@ -5,7 +5,6 @@
 
 __all__ = ['md5', 'is_string_like', 'first', 'decode']
 
-
```
yymao, line 8: restore this blank line

GCRCatalogs/utils.py
```diff
@@ -34,7 +33,6 @@ def first(iterable, default=None):
     """
     return next(iter(iterable), default)
 
-
```
yymao, line 37: restore this blank line

GCRCatalogs/register.py
```diff
@@ -22,7 +22,6 @@
 _CONFIG_DIRPATH = os.path.join(_HERE, _CONFIG_DIRNAME)
 _SITE_CONFIG_PATH = os.path.join(_HERE, "site_config", "site_rootdir.yaml")
 
-
```
yymao, line 25: restore this blank line

GCRCatalogs/register.py
```diff
@@ -76,11 +74,11 @@ def load_catalog_from_config_dict(catalog_config):
         catalog_config[Config.READER_KEY], __package__, BaseGenericCatalog
     )(**catalog_config)
 
-
```
yymao, line 79: restore this blank line

GCRCatalogs/register.py
```diff
@@ -176,6 +185,21 @@ def set_root_dir_by_site(self, site):
             site_string = ' '.join(_config_register.site_list)
             warnings.warn(f"Unknown site '{site}'.\nAvailable sites are: {site_string}\nroot_dir is unchanged")
 
+    def persist_root_dir(self):
+        """
+        Write current session value of root_dir to user config.
+        """
+        self._user_config_manager[self._ROOT_DIR_KEY] = os.path.abspath(self.root_dir)
+    def unpersist_root_dir(self):
```
yymao, line 193: Add a blank line
```suggestion

    def unpersist_root_dir(self):
```

GCRCatalogs/user_config_mgr.py
```diff
@@ -0,0 +1,123 @@
+import os
+import warnings
+import yaml
+from collections.abc import MutableMapping
```
yymao, line 4: Remove unused import (`warnings`).  Import standard modules before external packages.

```suggestion
from collections.abc import MutableMapping

import yaml
```

GCRCatalogs/register.py
```diff
@@ -680,6 +714,5 @@ def retrieve_paths(name_startswith=None, name_contains=None, **kwargs):
     The format would be [(catalog_name, original_path, resolved_path), ...]
     """
     return _config_register.retrieve_paths(name_startswith=name_startswith, name_contains=name_contains, **kwargs)
-
-
 _config_register = ConfigRegister(_CONFIG_DIRPATH, _SITE_CONFIG_PATH)
+
```
yymao, line 718: Two blank lines after module-level function definition. 
Leave only one blank line at the end of file (currently there's two; GitHub hides the last one automatically.)

```suggestion


_config_register = ConfigRegister(_CONFIG_DIRPATH, _SITE_CONFIG_PATH)
```

GCRCatalogs/register.py
```diff
@@ -558,12 +582,14 @@ def get_root_dir():
...
 
-def set_root_dir(path):
+def set_root_dir(path, write_to_config=False):
     """
     Sets runtime root_dir to *path*.
     """
     _config_register.root_dir = path
 
+    if write_to_config:
+        _config_register.persist_root_dir()
```
yymao, line 592: Two blank lines between module-level functions.

```suggestion
        _config_register.persist_root_dir()

```

GCRCatalogs/user_config.py
```diff
@@ -0,0 +1,42 @@
+from argparse import ArgumentParser, RawTextHelpFormatter
+from .user_config_mgr import UserConfigManager
+
+if __name__ == "__main__":
+    desc="""Directly manipulate items in user config"""
+
+    parser = ArgumentParser(description=desc, prog='GCRCatalogs.user_config',
```
yymao, line 7: spaces around `=`. `desc` is fine but we also happen to be in the DESC. To avoid confusion, switch to `usage`

```suggestion
    usage = """Directly manipulate items in user config"""

    parser = ArgumentParser(description=usage, prog='GCRCatalogs.user_config',
```

GCRCatalogs/user_config_mgr.py
```diff
@@ -0,0 +1,123 @@
...
+
+"""
+Utility class for managing user config (a dict) persisted to a yaml file.
```
yymao, line 7: No trailing space
```suggestion

"""
Utility class for managing user config (a dict) persisted to a yaml file.
```

GCRCatalogs/user_config_mgr.py
```diff
@@ -0,0 +1,123 @@
...
+Utility class for managing user config (a dict) persisted to a yaml file. 
+"""
+
+__all__ = ['UserConfigManager']
+
+
+class UserConfigManager(MutableMapping):
+    _default_config_filename = "gcr-catalogs.yaml"
+    _default_config_reldir = "lsstdesc"
+
```
yymao, line 16: no space in a blank line
```suggestion

```

yymao (changes requested):
@JoanneBogart I made several style comments. Generally, please have

- One blank line between class methods
- Two blank lines between module-level functions and classes
- One blank line at the end of file
- No trailing spaces, or spaces in blank lines
- One space around assignment operator (except in function argument definition)

I'm sending these in first. In the meantime I'll also test the script.

yymao:
I have tested the config setting script and it works well. However, I find the warning message is still confusing for non-DESC users. There are two issues:

(1) Warning message is current printed at GCRCatalogs import time. So it shows up when I using the user_config script, which does not require a preset roor dir. I think we should print out the warning at `root_dir` access time, i.e., here:

```python
class RootDirManager:

    @property
    def root_dir(self):
         # print warning here.
```

(2) The warning message is still aiming at DESC users. We should clearly indicate the parts that are for DESC users and the parts for non-DESC users in the message.

GCRCatalogs/user_config.py
JoanneBogart: `usage` has a different meaning to ArgumentParser so I'd rather not use that.   How about `descrip` or, if that's not acceptalbe, `description`?

yymao: let's just use `description` then.

JoanneBogart:
Concerning your style preferences - I understand that a uniform appearance (e.g. always one blank line between class methods) is helpful.   Why do you care about extra blanks at the end of a line or an extra blank line at the end of a file?

yymao:
The main purpose is to reduce the potential changes in future PRs.  Most editors, including the one I use, automatically remove trailing spaces when saving the file. So if we leave trailing space in, next time when I or someone uses such an editor to edit this file, the new PR will include changes on lines that are actually not relevant to the new PR. It's still ok but I'd like to reduce that as much as possible.

yymao (changes requested):
Thanks, @JoanneBogart --- hopefully the last review!

- In `user_config_mgr.py`, please remove `import warnings` and also all trailing spaces.
- In `register.py`, please add two blank line before Lines 728 `_config_register = ConfigRegister(_CONFIG_DIRPATH, _SITE_CONFIG_PATH)`

yymao (approved):
Thank you, @JoanneBogart! All looks good!

PR #504, opened 2020-12-06
Truth-match catalog reader

### LINKED ISSUE #500: Reader and config for DC2 "truth-match" catalog
As part of the DC2 public release, we are generate a "truth-match" catalog that is partitioned in tracts (see https://github.com/LSSTDESC/DC2-production/pull/403 for details about this catalog). 

Since the "truth-match" catalog is in parquet and split by tract, its reader should be a subclass of [`dc2_dm_catalog.DC2DMCatalog`](https://github.com/LSSTDESC/gcr-catalogs/blob/v1.0.4/GCRCatalogs/dc2_dm_catalog.py#L93).

For each tract, the "truth-match" catalog will first have *N* rows that are in the same row order as the object catalog of that tract (*N* = number of objects in the object catalog), and will follow by *M* rows of truth entries that do not have match in the object catalog. 

We want to implement a new reader option (e.g., `for_composite`) so that, when set, the reader will always discard the last *M* rows. 
When the option is not set, the reader will discard any non-unique truth entries in the first *N* rows and return the rest plus the last *M* rows.

yymao:
This PR add a truth-match catalog reader and relevant catalog configs. It fixes #500. 

PR is ready for review, but the path in catalog configs will need to be updated once we generate all the truth-match files and move then to the share space. #509 is opened to keep track of this.

In the composite config, I have added several options that will be available once we upgrade to GCR v0.9.0. Right now those options are ignored by GCR v0.8.x.

The quantity modifiers in the truth-match reader are a bit hack-y. They work fine, but I welcome better ideas.

### DIFF
--- a/GCRCatalogs/dc2_truth_match.py
+++ b/GCRCatalogs/dc2_truth_match.py
@@ -44,7 +44,7 @@ class DC2TruthMatchCatalog(DC2DMTractCatalog):
         if self._as_object_addon:
             no_postfix = ("is_unique_truth_entry", "match_sep", "match_objectId")
             self._quantity_modifiers = {
-                (k + ("" if k in no_postfix else "_truth"): v for k, v in self._quantity_modifiers.items()
+                (k + ("" if k in no_postfix else "_truth")): v for k, v in self._quantity_modifiers.items()
             }
 
     def _obtain_native_data_dict(self, native_quantities_needed, native_quantity_getter):
--- a/GCRCatalogs/dc2_truth_match.py
+++ b/GCRCatalogs/dc2_truth_match.py
@@ -1,0 +1,83 @@
+"""
+Reader for truth-match catalogs persisted as parquet files and partitioned in tracts.
+"""
+import numpy as np
+import astropy.units as u
+
+from .dc2_dm_catalog import DC2DMTractCatalog
+
+__all__ = ["DC2TruthMatchCatalog"]
+
+
+def _flux_to_mag(flux):
+    return (flux * u.nJy).to_value(u.ABmag)  # pylint: disable=no-member
+
+
+class DC2TruthMatchCatalog(DC2DMTractCatalog):
+    r"""
+    DC2 Truth-Match (parquet) Catalog reader
+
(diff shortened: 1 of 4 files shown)


GCRCatalogs/dc2_truth_match.py
```diff
@@ -0,0 +1,87 @@
...
+def _flux_to_mag(flux):
+    with np.errstate(divide="ignore"):
+        return (flux * u.nJy).to_value(u.ABmag)  # pylint: disable=no-member
+
+
+class DC2TruthMatchCatalog(DC2DMTractCatalog):
+    r"""
+    DC2 Truth-Match (parquet) Catalog reader
+
+    Presents tables exactly as they are defined in the files (no aliases,
```
JoanneBogart, line 21: What about the added magnitude columns?
And, for addon case, the name changes to avoid collisions with names in object?

GCRCatalogs/dc2_truth_match.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if self._as_object_addon:
+            no_postfix = ("truth_type", "is_unique_truth_entry", "match_sep", "match_objectId")
+            self._quantity_modifiers = {
+                (k + ("" if k in no_postfix else "_truth")): (v or k) for k, v in self._quantity_modifiers.items()
+            }
+
+    def _obtain_native_data_dict(self, native_quantities_needed, native_quantity_getter):
+        """
+        Overloading this so that we can query the database backend
```
JoanneBogart, line 53: docstring (left over from original truth reader for sqlite dbs?) needs to be updated, to say something about adding other quantities and special handling as appropriate for the mode (as addon or as truth table) of use.

JoanneBogart (changes requested):
A couple of the docstrings need revision.  Everything else (actual code; config files) LGTM

JoanneBogart:
Another way to characterize columns in `no_postfix` would be to say the column names containing "truth" or "match".  But I'm not sure that's any better than just listing them.

JoanneBogart (approved):
Looks fine.

yymao:
Thanks, @JoanneBogart

PR #505, opened 2020-12-06
Use the improved composite in GCR v0.9.0

### LINKED ISSUE #398: Composite Catalog return_iterator
It looks like when you do `get_quantities` with `return_iterator=True` on a composite catalog then GCR loads all the columns for one catalog into memory and then only really iterates on the second one.   I'm not sure if this is generally true or just for the DC2 catalogs I'm using (Parquet versions, joining the metacal and object ones).

This is okay for small catalogs but for large ones will start to cause problems with larger ones.  I'm not sure how easy this is, but there are a few ways one could avoid it.

### LINKED ISSUE #410: get_quantity_info() does not return information for composite catalogs
I was attempting to add some documentation to the photo-z columns for the add-on catalogs via adding some entries to the _subclass_init to be returned by get_quantity_info().  This works fine when the standalone catalog is loaded, however, for composite catalogs it looks like get_quantity_info() only shows entries for the first of two composites.  The photo-z catalogs will be mostly used as composite catalogs matched to a larger photometry catalog, so solving this so that people can get the column descriptions may be important (along with other composite catalogs).

yymao:
I am improving the composite catalog class in https://github.com/yymao/generic-catalog-reader/pull/32. Once GCR is updated to 0.9.0, we should merge this PR to make use the new composite class. 

This PR will fix #398 and fix #410.

### DIFF
--- a/GCRCatalogs/composite.py
+++ b/GCRCatalogs/composite.py
@@ -2,8 +2,17 @@
 composite reader
 """
 import warnings
-from GCR import CompositeCatalog, MATCHING_FORMAT, MATCHING_ORDER
+
+from GCR import CompositeCatalog
+_HAS_NEW_COMPOSITE = True
+try:
+    from GCR import CompositeSpecs
+except ImportError:
+    from GCR.composite import MATCHING_FORMAT, MATCHING_ORDER
+    _HAS_NEW_COMPOSITE = False
+
 from .register import load_catalog, load_catalog_from_config_dict, has_catalog
+
 
 class CompositeReader(CompositeCatalog):
     def __init__(self, **kwargs):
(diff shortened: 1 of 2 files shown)


yymao:
I have just released GCR v0.9.0. This PR is now ready for review. The new features would only be available in GCR v0.9.0.

joezuntz:
I don't think I know enough to review this, but many thanks for the feature!

JoanneBogart (approved):
Most of the work appears to be in GCR.  The small changes to GCRCatalogs code look fine.  As a minimal test, after pulling this branch to my local repo and installing GCR 0.9.0,  I was able to load a composite catalog.  LGTM

yymao:
Thanks @JoanneBogart!

PR #507, opened 2020-12-07
Change decode to be compatible with v2 and v3 of h5py

### LINKED ISSUE #497: Some unicode issue breaks cosmoDC2 reader in desc-python-bleed
In the latest `desc-python-bleed`, one would encounter the follow error when loading the cosmoDC2 catalog via `GCRCatalogs.load_catalog("cosmoDC2_v1.1.4_image")`: 

```python-traceback
/usr/local/py3.7/envs/desc/lib/python3.6/site-packages/GCRCatalogs/cosmodc2.py in <lambda>(k, v)
    233         if collect_info_dict:
    234             quantity_info_dict = dict()
--> 235             modifier = lambda k, v: None if k == 'description' and v == b'None given' else v.decode()
    236             for quantity in native_quantities:
    237                 quantity_info_dict[quantity] = {k: modifier(k, v) for k, v in fh[group_name][quantity].attrs.items()}

AttributeError: 'str' object has no attribute 'decode'
```

In `desc-python`, this issue does not exist. Given the error message and the different behaviors in the two environment, most likely the issue is due to some updated package changes the string/basestring type when reading the HDF5 file. Need to track down which package causes this and to understand the changed behavior, and then we can update the reader so that it works with both versions.

yymao:
This PR fixes #497. 

In `desc-python` the `h5py` package is in v2.10.0, while in `desc-python-bleed` it is in v3.1.0. They process bytestring differently. In v2, bytestring remains bytestring while in v3, it is converted to unicode string. 

In our reader we have assumed v2 behavior, and hence call `.decode()` to convert to unicode string. This  action will fail with h5py v3. Hence, a new utility function `decode` is added to decode if input is bytestring, otherwise just return. By using this utility function we can be compatible with both v2 and v3 of `h5py`. 

I have tested this implementation on cosmoDC2 with both ``desc-python` and `desc-python-bleed`. Both cases work.

### DIFF
--- a/GCRCatalogs/utils.py
+++ b/GCRCatalogs/utils.py
@@ -37,7 +37,10 @@ def first(iterable, default=None):
 
 def decode(bytestring):
     """
-    decode a bytestring is possible
+    Decode a bytestring is possible, return str if not.
+
+    Only bytestring has .decode() method. This utility provides a way to homogenize values
+    that may be bytestring or unicode string  (such as the attributes in h5py v2.x and v3.x).
     """
     try:
         return bytestring.decode()
--- a/GCRCatalogs/utils.py
+++ b/GCRCatalogs/utils.py
@@ -37,7 +37,7 @@ def first(iterable, default=None):
 
 def decode(bytestring):
     """
-    Decode a bytestring is possible, return str if not.
+    Decode a bytestring if possible, return str if not.
 
     Only bytestring has .decode() method. This utility provides a way to homogenize values
     that may be bytestring or unicode string  (such as the attributes in h5py v2.x and v3.x).
--- a/GCRCatalogs/utils.py
+++ b/GCRCatalogs/utils.py
@@ -3,7 +3,8 @@ utility module
 """
 import hashlib
 
-__all__ = ['md5', 'is_string_like']
+__all__ = ['md5', 'is_string_like', 'first', 'decode']
+
 
 def md5(fname, chunk_size=65536):
     """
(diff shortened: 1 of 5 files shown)


GCRCatalogs/utils.py
```diff
@@ -32,3 +33,13 @@ def first(iterable, default=None):
     returns the first element of `iterable`
     """
     return next(iter(iterable), default)
+
+
+def decode(bytestring):
+    """
+    decode a bytestring is possible
```
JoanneBogart, line 40: Suggest  docstring text include the justification as in the pull request, e.g. something like

decode presumed bytestring.  This utility smooths over differences in return values between, in particular, v2 and v3 of h5py

JoanneBogart (approved):
Seems fine.  I made one minor suggestion.

GCRCatalogs/utils.py
yymao: Thanks! I expanded the docstring!

GCRCatalogs/utils.py
```diff
@@ -32,3 +33,16 @@ def first(iterable, default=None):
     returns the first element of `iterable`
     """
     return next(iter(iterable), default)
+
+
+def decode(bytestring):
+    """
+    Decode a bytestring is possible, return str if not.
```
JoanneBogart, line 40: New docstring is nice and clear, except I believe "is" in line 40 should be "if"

JoanneBogart (commented):
Approved with one minor request

GCRCatalogs/utils.py
yymao: Ah, thanks for catching this!

JoanneBogart (approved):
LGTM

yymao:
Thanks @JoanneBogart !

PR #508, opened 2020-12-07
Use MATCHING_FORMAT for Javi's matched add-ons

yymao:
@fjaviersanchez I just noticed that you have been using the `MATCHING_ORDER` option for your matched add-ons. While `MATCHING_ORDER` would work, the performance is not as good as `MATCHING_FORMAT`. The former requires the row ordering to be the same, while the latter requires both row ordering *and* partition methods to be the same across catalogs. 

I believe in your mached catalog, the partition method is indeed by tract, and hence these add-ons should qualify using `MATCHING_FORMAT`. Looking at your reader, I noticed that you have not implemented native_filters for tracts, which is also required for `MATCHING_FORMAT` to work. So I implemented the native filter and change the `matching_method` to `MATCHING_FORMAT` in several configs. 

I have also further edited `dc2_object_run2.2i_dr6_wfd_matched_addon` --- composite of composite will slow down the performance. But it's totally fine to have more than 2 catalogs in one composite. So I changed your composite of composite into 3 catalogs in one composite. 

Side note: I have to admit that name `MATCHING_FORMAT` is not precise. It seems to suggest the underlying format needs to match, but that's not what it means. In GCR v0.9.0 (https://github.com/yymao/generic-catalog-reader/pull/32), I introduced `matching_partition` and `matching_row_order` to be more precise.

### DIFF
--- a/GCRCatalogs/catalog_configs/dc2_object_run2.1.1i_matched_addon.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_object_run2.1.1i_matched_addon.yaml
@@ -5,7 +5,7 @@ description: DC2 Run 2.1.1i Object Catalog with Matched Table addon
 creators: ['Javi Sanchez', 'Eve Kovacs']
 catalogs:
   - catalog_name: dc2_object_run2.1.1i
-    matching_method: MATCHING_ORDER
+    matching_method: MATCHING_FORMAT
   - catalog_name: dc2_object_run2.1.1i_matched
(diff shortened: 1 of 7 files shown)


fjaviersanchez (approved):
Thank you so much @yymao! This is very useful (at least it is to me). Yes, the `MATCHING_FORMAT` bit was a little bit confusing to me but I don't have any better term for it (`MATCHING_NATIVE_FILTERS` maybe?)

PR #511, opened 2020-12-07
Update workflows

yymao:
This PR updates the github actions workflows. 

In `desc-env-test.yml`, while the stack build has been commented out, the workflow can use the `matrix` strategy to be more concise. There is also no need to check out the repo for this workflow to run. 

Using this opportunity I also updated the python ci test to give it a better name and add python 3.9 to the test.

### DIFF
--- a/.github/workflows/desc-env-test.yml
+++ b/.github/workflows/desc-env-test.yml
@@ -3,8 +3,6 @@ name: DESC Env Docker Build Test
 on:
   push:
     branches: [ master ]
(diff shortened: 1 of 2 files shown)


yymao:
`GCRCatalogs` depends on pyarrow, whose support for python 3.9 is still work in progress. https://github.com/apache/arrow/pull/8386

heather999 (approved):
LGTM - took a peek at the recent Actions runs.

PR #516, opened 2020-12-12
Use MutableMapping for UserConfigManager

yymao:
@JoanneBogart --- I am reviewing #503 and wanted to suggest that we use the abstract class `MutableMapping` for `UserConfigManager`. I made this PR (into your branch) to make my suggestion more specific. 

In particular, `UserConfigManager` behaves like a python dictionary, so we should keep the API as similar to that of a python dictionary as possible. The custom methods (`get_value`, `write_entries`,  `remove_keys`), although well-documented, are new names we are introducing for existing python syntax (and with slightly different behaviors, requiring the users to learn how to use these custom methods). Use `MutableMapping` as the base class resolves this issue by allowing users to use syntax that they are already familiar with the native python dictionary. 

The remaining changes in the PR are just space conventions.


JoanneBogart:
Much of the content here has been incorporated in u/jrbogart/site-root_dir-issue499. See PR #503

PR #517, opened 2020-12-14
Add Code of Couduct

yymao:
In anticipation of possible future participation of non-DESC members in this repo, I propose that we add a Code of Conduct to this repo, which can simply refer to [DESC Code of Conduct](https://lsstdesc.org/assets/pdf/policies/LSST_DESC_Professional_Conduct.pdf).


wmwv (dismissed):
LGTM

PR #518, opened 2020-12-17
Update "meta" (quantity info) files for object and truth-match catalogs

yymao:
This PR updates/adds the "meta" (quantity info) files for the object and truth-match catalogs. There are two major changes:

1. **Allows dictionary style in the `*_meta.yaml`** --- this format is more self-contained and can be used for other purpose (e.g. generating a latex table). The code to read these meta files is revised accordingly (`DC2DMCatalog._generate_info_dict`) and it works with both the original (list) format and the new (dictionary) format. 
2. **Enable `<band>` replacement** --- Previously, calling `cat.get_quantity_info("mag_r_cModel")` would not return any info, because in the meta file, the column name was specified as `mag_<band>_cModel`, and hence `mag_r_cModel` would not match. This update modifies `DC2DMCatalog._generate_info_dict` so that `<band>` is properly replaced with available bands. As part of this change, edits are made to `DC2ObjectParquetCatalog._detect_available_bands` and `DC2DMTractCatalog._detect_available_bands` so that they can properly detect the available bands. 

I have tested this PR locally.  This PR is ready for review.


GCRCatalogs/dc2_object.py
```diff
@@ -678,7 +655,10 @@ def _subclass_init(self, **kwargs):
         super()._subclass_init(**kwargs)
 
     def _detect_available_bands(self):
-        return [col.rpartition('_')[0] for col in self._columns if col.endswith('_FLUXMAG0')]
+        return (
+            [col.rpartition('_')[0] for col in self._columns if col.endswith('_FLUXMAG0')] or
+            [col.rpartition('_')[2] for col in self._columns if col.startswith('psFlux_')]
+        )
```
wmwv, line 661: I would suggest using a couple of lines to write out the logic here and maybe even insert a sanity check.  Clarify that the two different column names possible are from different versions and  then do something different based on the version.
In theory, this isn't robust against there being mixed columns, but in practice if that happens perhaps something should be flagged.

wmwv (dismissed):
LGTM
One optional suggestion.
I haven't run it myself.

yymao:
Thanks @wmwv I've addressed your comment and also fixed a bug.

GCRCatalogs/dc2_dm_catalog.py
```diff
@@ -169,7 +171,7 @@ def _generate_modifiers(**kwargs):  # pylint: disable=unused-argument
         return dict()
 
     @staticmethod
-    def _generate_info_dict(meta_path):
+    def _generate_info_dict(meta_path, bands=None):
         """Creates a 2d dictionary with information for each homogenized quantity
 
         Args:
```
JoanneBogart, line 177: Add something about bands argument

GCRCatalogs/dc2_object.py
```diff
@@ -678,10 +655,23 @@ def _subclass_init(self, **kwargs):
...
-        return [col.rpartition('_')[0] for col in self._columns if col.endswith('_FLUXMAG0')]
+        """
+        This method should return available bands in the catalog file.
+        For object catalog files generated by DC2-production/scripts/make_object_catalog.py,
+        columns `<band>_FLUXMAG0` would exist.
+        For DPDD-only object catalog files generated by DC2-production/scripts/write_gcr_to_parquet.py,
+        columns `psFlux_<band>` would exist.
+        This function first check `<band>_FLUXMAG0`, and if none exist, check `psFlux_<band>`.
+        If neither exists, returns an empty list.
+        Note that band name may contain underscroll.
```
JoanneBogart, line 666: """ suggestion
This function first checks for `<band>_FLUXMAG0` columns, and if none exists, then checks for `psFlux_<band>`.
If neither exists, it returns an empty list.
Note that band name may contain underscores
"""

GCRCatalogs/dc2_truth_match.py
```diff
@@ -1,13 +1,18 @@
...
+
 import numpy as np
 import astropy.units as u
 
 from .dc2_dm_catalog import DC2DMTractCatalog
 
 __all__ = ["DC2TruthMatchCatalog"]
 
+FILE_DIR = os.path.dirname(os.path.abspath(__file__))
+META_PATH = os.path.join(FILE_DIR, 'catalog_configs/_dc2_truth_match_meta.yaml')
```
JoanneBogart, line 14: Just an observation:  there seems to be no way to deal with schema versioning, but in this way truth_match is no different from other meta descriptions. Maybe this possibility is handled somewhere else?

JoanneBogart (approved):
I left a couple minor suggestions about documentation; otherwise it looks fine. 
One other comment concerns versioning of meta descriptions. It's not meant to be a request for any changes; I'm just curious.

GCRCatalogs/dc2_dm_catalog.py
yymao: Added!

GCRCatalogs/dc2_object.py
yymao: Thanks for the suggestion! Fixed. 

(BTW, use ``` to open and close suggestion box will enable GitHub's suggestion mode.)

GCRCatalogs/dc2_truth_match.py
yymao: Because of the design philosophy of "maximum flexibility", it is entirely left to each reader to decide to how to implement the quantity info. 

The quantity info should be embedded could be embedded in the data files themselves, in which case they are automatically versioned (like the case of cosmoDC2 --- it reads in quantity info directly from the hdf5 file instead of a meta file).  

For object catalog and alike, we opted to use a meta yaml file checked in the github repo. If the same reader wants to access different meta files, we can define multiple `META_PATH` and the reader can pick on based on, say, an option in the catalog config.

JoanneBogart (approved):
LGTM

yymao:
Thanks, @JoanneBogart!

GCRCatalogs/dc2_truth_match.py
JoanneBogart: Ok.  I guess what I would have expected for those readers which use an external metafile would be something like your last remark: there could be a recognized option in the the config file, call it "meta_path" (ideally the same string for any such reader).   If it was there, the reader would use its value; otherwise it would use a default.   Currently it's not an issue since there is no need for an alternate.

yymao: Agreed! Implemented now in #522.

PR #519, opened 2020-12-17
Improve parquet reading performance/memory usage for arrow 0.16+

yymao:
This PR improves parquet reading performance/memory usage for arrow 0.16.0+. 

In arrow 0.16.0+, two new options `split_blocks` and `self_destruct` are introduced to `to_pandas` -- the first option allows better performance by introducing zero-copy opportunity, and the second option ensures unused references are deleted for garbage collection to reclaim memory (cf. https://github.com/apache/arrow/pull/6067).

Adding these option to our parquet wrapper improves performance/memory usage. Backward (for arrow <0.16.0) compatibility is maintained. 

This PR has been tested locally and is ready for review.

### DIFF
--- a/GCRCatalogs/parquet.py
+++ b/GCRCatalogs/parquet.py
@@ -4,7 +4,7 @@ import re
 __all__ = ['ParquetFileWrapper']
 
 
-def _retrive_data_from_arrow_table(table, as_dict=False):
+def _retrieve_data_from_arrow_table(table, as_dict=False):
     try:
         # Options introdcued in arrow 0.16+ to improve speed and memory usage
         df = table.to_pandas(split_blocks=True, self_destruct=True)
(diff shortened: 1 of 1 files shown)


GCRCatalogs/parquet.py
```diff
@@ -3,6 +3,20 @@
 
 __all__ = ['ParquetFileWrapper']
 
+
+def _retrive_data_from_arrow_table(table, as_dict=False):
```
JoanneBogart, line 7: "retrive" is not an English word.  I think you meant "retrieve" so method should be named

_retrieve_data_from_arrow_table

JoanneBogart (changes requested):
There is just one naming thing which should be addressed.  Otherwise LGTM.

GCRCatalogs/parquet.py
yymao: Thanks for catching this! Fixed now.

JoanneBogart (approved):
Good to go

yymao:
Thanks, @JoanneBogart!

PR #513, opened 2020-12-09
add dr2 truth-match catalog config

yymao:
This PR adds add the truth-match catalog config for Run2.21 DR2 (v1). 

@evevkovacs you may be able to use this catalog for validation. ~but keep in mind that:~
~1. The column names may differ slightly from @fjaviersanchez's version~
~2. This version of truth-match only use nearest neighbor, without magnitude cut. There may be more mismatches especially given that DR2 is shallower.~

Updated Dec 12: The two points I noted above no longer apply with the latest set of commits (which fixed point 1) and with https://github.com/LSSTDESC/DC2-production/pull/409 (which fixed point 2).

PR #520, opened 2020-12-20
Bug fix in truth match catalog

yymao:
A bug was introduced in #513 when I added the `as_matchdc2_schema` option:

When `as_matchdc2_schema` is set to True in the config, both `self._as_matchdc2_schema` and `self._as_object_addon` should be set to True. 

However, when `as_matchdc2_schema` is set to False or not set in the config, only `self._as_matchdc2_schema` should be set to False, and `self._as_object_addon` should be left unchanged. In #513 I accidentally set `self._as_object_addon` to False too.

This PR fixes this logical bug.


JoanneBogart (approved):
This logic looks right.

yymao:
Thanks, @JoanneBogart!

PR #521, opened 2020-12-20
Make get_quantity_info work with "_truth" postfix in truth-match add-on

yymao:
In the truth-match add-on catalog (i.e., when `as_object_addon` set to True), most columns are postfixed with `_truth` to distinguish them from the object columns ([code](https://github.com/LSSTDESC/gcr-catalogs/blob/d21ca7c66e95119f685faac9499585cdb386d7eb/GCRCatalogs/dc2_truth_match.py#L69)). However, this would resulting in code like:

```python
cat.get_quantity_info("mag_r_noMW_truth")
```

not working, as `mag_r_noMW_truth` is not a known quantity in the quantity info dictionary (which, in this case, was obtained from the meta yaml file). 

This PR overwrites the `_get_quantity_info_dict` method by adding a check before calling the parent method, and removes "_truth" postfix when the quantity does not appear in the info dictionary but does appear in quantity modifier.

### DIFF
--- a/GCRCatalogs/dc2_dm_catalog.py
+++ b/GCRCatalogs/dc2_dm_catalog.py
@@ -143,6 +143,8 @@ class DC2DMCatalog(BaseGenericCatalog):
 
         if self.META_PATH:
             self._quantity_info_dict = self._generate_info_dict(self.META_PATH, bands)
+        else:
+            self._quantity_info_dict = dict()
 
         self._len = None
 
--- a/GCRCatalogs/dc2_truth_match.py
+++ b/GCRCatalogs/dc2_truth_match.py
@@ -158,3 +158,16 @@ class DC2TruthMatchCatalog(DC2DMTractCatalog):
                     col,
                     "{}_match_mask".format(t),
                 )
+
+    def _get_quantity_info_dict(self, quantity, default=None):
+        """
+        Befere calling the parent method, check if `quantity` has an added "_truth" postfix
+        due to the `if self._as_object_addon:...` part in _subclass_init. If so, remove the postfix.
+        """
+        if (
+            quantity not in self._quantity_info_dict and
+            quantity in self._quantity_modifiers and
+            quantity.endswith("_truth")
+        ):
+            quantity = quantity[:-6]
+        return super()._get_quantity_info_dict(quantity, default)
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
     ],
     keywords='GCR',
     packages=['GCRCatalogs'],
-    install_requires=['requests', 'pyyaml', 'numpy', 'astropy', 'GCR>=0.9.1'],
+    install_requires=['requests', 'pyyaml', 'numpy', 'astropy', 'GCR>=0.9.2'],
     extras_require={
         'full': ['h5py', 'healpy', 'pandas', 'pyarrow', 'tables'],
     },


JoanneBogart:
Probably there is something very basic I'm missing here, but I don't see any place where self.META_PATH gets set for DC2TruthMatchCatalog or its superclasses and so `get_quantity_info` fails .  When, for simplicity, I just load the truth match catalog by itself and try, e.g., `truth_match.get_quantity_info('is_variable')` the result is
```
>>> truth_match.get_quantity_info('is_variable')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/global/common/software/lsst/common/miniconda/py3.7-4.7.12.1-v2/envs/desc/lib/python3.7/site-packages/GCR/base.py", line 188, in get_quantity_info
    d = self._get_quantity_info_dict(quantity, default if key is None else dict())
  File "/global/u2/j/jrbogart/desc_git/gcr-catalogs/GCRCatalogs/dc2_truth_match.py", line 157, in _get_quantity_info_dict
    quantity not in self._quantity_info_dict and
AttributeError: 'DC2TruthMatchCatalog' object has no attribute '_quantity_info_dict'
```

yymao:
@JoanneBogart -- `META_PATH` is set for `DC2TruthMatchCatalog` in #518 that we merged earlier. So I have to rebase this branch to master, and force-pushed to this PR. Please re-fetch this branch. 

The `AttributeError` you got; however, is another subtle bug that I have not noticed. When `META_PATH` is (incorrectly) not set, `_quantity_info_dict` should still be set to an empty dictionary. I have added a fix.

JoanneBogart:
Looks like setup should be modified to require GCR>=0.9.2.

JoanneBogart:
I did `$ pip install --local --upgrade GCR`  and set PYTHONPATH to my .local/lib... directory. I see
```
>>> import GCR
>>> GCR.__version__
'0.9.2'
```
but I still wonder if somehow GCRCatalogs is not using the right version. I don't see how to access the quantities with postfix '_truth'.  You can see various attempts:
```
>>> objmatch = GCRCatalogs.load_catalog('dc2_object_run2.2i_dr6_wfd_with_truth_match.yaml')
>>> objmatch.get_quantity_modifier('ra')
('dc2_object_run2.2i_dr6_wfd_v1', 'ra')
>>> objmatch.get_quantity_modifier('ra_truth')
>>> objmatch.get_quantity_modifier('is_pointsource')
('dc2_truth_match_run2.2i_dr6_wfd', 'is_pointsource')
>>> objmatch.get_quantity_modifier('flux_r')
('dc2_truth_match_run2.2i_dr6_wfd', 'flux_r')
>>> objmatch.get_quantity_modifier('mag_r')
('dc2_object_run2.2i_dr6_wfd_v1', 'mag_r')
>>> objmatch.get_quantity_modifier('mag_r_truth')
>>> q_list = objmatch.list_all_quantities()
>>> len(q_list)
4698
>>> for q in q_list:
...     if q.find('truth') > -1:
...         print(q)
... 
truth_type
is_unique_truth_entry
>>> truths = objmatch.get_quantities(['ra', 'ra_truth'], native_filters=['tract=3256'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/global/common/software/lsst/common/miniconda/py3.7-4.7.12.1-v2/envs/desc/lib/python3.7/site-packages/GCR/base.py", line 65, in get_quantities
    quantities = self._preprocess_requested_quantities(quantities)
  File "/global/common/software/lsst/common/miniconda/py3.7-4.7.12.1-v2/envs/desc/lib/python3.7/site-packages/GCR/base.py", line 386, in _preprocess_requested_quantities
    self._check_quantities_exist(quantities, raise_exception=True)
  File "/global/common/software/lsst/common/miniconda/py3.7-4.7.12.1-v2/envs/desc/lib/python3.7/site-packages/GCR/base.py", line 372, in _check_quantities_exist
    raise ValueError(msg)
ValueError: Native quantity `ra_truth` does not exist (required by `ra_truth`)
```

yymao:
Ah, that's due to the bug that I fixed in #520. Let me rebase this branch again.

yymao:
OK, just force-pushed. Hopefully it would work now.

JoanneBogart:
Yes!   All the responses look reasonable now.

JoanneBogart (changes requested):
If, as it appears, GCR version 0.9.2 is required, setup.py should be changed to reflect that. 

Not directly related to this PR, I recommend that in  `_dc2_truth_match_meta.yaml` the `the` in the description for `is_good_match` be changed to `all`, that is
```
"True if this object--truth matching pair satisfies all matching criteria"
```
I think it makes it a little clearer what the distinction between a not-so-good match and a good match is without going into a lot of detail.

Finally, I'm very slightly nervous about the convention of adding `_truth` to the end to form aliases.  It's not very likely that any quantities already ending in `_truth` will be added to the catalog, but if they were it would be a problem.  Maybe use `__truth` for the suffix?  Just a thought, not something you have to address.

yymao:
Thanks @JoanneBogart 

Re: GCR version --- This PR doesn't actually _require_ GCR >=0.9.2. It fixes a different issue than what GCR 0.9.2 fixes. Have both make the `get_quantity_info` works better, but they are not dependent of each other. Anyway, I updated the setup.py file. 

Re: `is_good_match` description: sounds good. I've opened #527 to fix that. 

Re: `_truth`  -- can you explain a bit more about your concerns? If I am guessing your concern right (which I am not certain), I think it's ok because if a quantity that ends in `_truth` appears in the catalog, that quantity should appear in `self._quantity_info_dict` and hence the if statement in `_get_quantity_info_dict` won't be fulfilled.

JoanneBogart:
@yymao  Thanks for the explanation.   If there were such a column it would make it hard by eye to tell which quantities were aliases (which shouldn't matter to anyone), but I see that the code would work correctly.

JoanneBogart (approved):
All my comments have been addressed. Thanks!

yymao:
From a user's perspective, would they want/need to know which column is an alias?

PR #523, opened 2020-12-20
add helper functions for tract catalogs

yymao:
This PR adds two helper functions for tract catalogs -- `tract_filter` and `sample_filter`. They are used in the tutorial notebooks for public release. Since they will be used in multiple notebooks and the users are likely to use them in their own script (if they follow the notebook), it seems that putting them into `GCRCatalogs` would be a better choice. 

Here, `tract_filter` and `sample_filter` return GCRQuery objects that can be used in the `native_filters` and `filters` arguments, respectively, of `get_quantities()`. They can help select specific tracts or down sample the rows at random.

### DIFF
--- a/GCRCatalogs/helpers/base_filters.py
+++ b/GCRCatalogs/helpers/base_filters.py
@@ -10,10 +10,14 @@ from GCR import GCRQuery
 __all__ = ["partition_filter", "sample_filter"]
 
 
-def partition_filter(name, ids, id_high=None):
+def partition_filter(partition_name, ids, id_high=None):
     """
     Returns a GCRQuery object to be used in the `native_filters` argument of get_quantities(),
-    to select a subset of partitions. Partitions must have integer IDs.
+    to select a subset of partitions.
+
+    *partition_name* must be a "native filter quantity" in GCR,
+    and the partitions must be specified by integer IDs.
+    Existing examples include "tract" for object catalogs and "healpix_pixel" for cosmoDC2.
 
     If *ids* is a single integer, select only that partition.
     If *ids* and *id_high* are both given as single integers, select [ids, id_high]
@@ -22,19 +26,19 @@ def partition_filter(name, ids, id_high=None):
     """
     if isinstance(ids, int):
         if id_high is None:
-            return GCRQuery(f"{name} == {ids}")
+            return GCRQuery(f"{partition_name} == {ids}")
         elif isinstance(id_high, int):
-            return GCRQuery(f"{name} >= {ids}", f"{name} <= {id_high}")
-        raise ValueError(f"When `{name}s` is an integer, `{name}_high` must be an integer or None.")
+            return GCRQuery(f"{partition_name} >= {ids}", f"{partition_name} <= {id_high}")
+        raise ValueError(f"When `{partition_name}s` is an integer, `{partition_name}_high` must be an integer or None.")
 
     ids = np.unique(np.asarray(ids, dtype=np.int))
     if not ids.size:
-        raise ValueError(f"Must select at least one {name}.")
+        raise ValueError(f"Must select at least one {partition_name}.")
 
     def _partition_selector(partition_ids, ids_to_select=ids):
         return np.isin(partition_ids, ids_to_select, assume_unique=True)
 
-    return GCRQuery((_partition_selector, name))
+    return GCRQuery((_partition_selector, partition_name))
 
 
 def sample_filter(ref_col_name, frac, random_state=None):
@@ -54,7 +58,7 @@ def sample_filter(ref_col_name, frac, random_state=None):
     if frac == 1:
         return GCRQuery()
     if frac == 0:
-        return GCRQuery((lambda a: np.zeros_like(a, dtype=np.bool), "tract"))
+        return GCRQuery((lambda a: np.zeros_like(a, dtype=np.bool), ref_col_name))
 
     if not isinstance(random_state, np.random.RandomState):
         random_state = np.random.RandomState(random_state)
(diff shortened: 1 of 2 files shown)


GCRCatalogs/helpers/tract_catalogs.py
```diff
@@ -0,0 +1,65 @@
...
+        return GCRQuery()
+    if frac == 0:
+        return GCRQuery((lambda a: np.zeros_like(a, dtype=np.bool), "tract"))
+
+    if not isinstance(random_state, np.random.RandomState):
+        random_state = np.random.RandomState(random_state)
+    seed = random_state.randint(65536)
+
+    def _sampler(tract, frac=frac, seed=seed):
+        size = len(tract)
```
wmwv, line 60: I don't have in my mind clearly how this will be called.

But if a user were directly calling this then I would be concerned about this assumption about the type of `tract`.
1. If `tract` is an int, then `len` will raise `TypeError`
2. If `tract` is a string, then `len` will give the length of the string
3. If `tract` is an array, then `len` will give the length of the array
Sorry, you know all this.  I'm half just writing this out for myself.  But I was a bit concerned about the type safety here.

wmwv: Because the next line seems to be implying that if `size!=0` then we return `....rand(size)` but if `size==0` then we return `np.zeros`.  But we only get `size==0` if passed an empty list (empty collection).  When would that happen?

wmwv (approved):
LGTM.

One naming question.

And a second a bit of confusion on my part about how the `_sampler` gets called that probably can be best addressed by a bit of code comment/documentation rather than any actual change to functionality.

GCRCatalogs/helpers/tract_catalogs.py
yymao: Thanks for raising this, @wmwv. `_sampler` is not a user facing function, and it's only used in `GCRQuery` (see the return line). Recall that for `GCRQuery`, one can supply a tuple that consists of a callable + columns to feed into that callable. Here, `_sampler` is that callable, and `tract` is the column of tract IDs that will be sent to `_sampler`. The `_sampler` must return a boolean mask array that have the same length as the input array. 

Hence, `tract`, which I now rename to `tract_arr`, will always be a numpy array of tract IDs. For a given tract, it'll be the same value. In fact, for the purpose of sampling, I could choose the `objectId` column as well. I am choosing `tract` because this function can them be applied to any catalog that has `tract` (and may not have `objectId`).

JoanneBogart (approved):
I'm approving the merge as is, but see my comment.  You may want to address it (which would probably involve changing the filename as well to, e.g., `partitioned_catalogs.py`)

GCRCatalogs/helpers/tract_catalogs.py
```diff
@@ -0,0 +1,65 @@
...
+"""
+
+import numpy as np
+
+from GCR import GCRQuery
+
+__all__ = ["tract_filter", "sample_filter"]
+
+
+def tract_filter(tracts, tract_high=None):
```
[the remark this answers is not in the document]
yymao: Both `tract_filter` and `sample_filter` are just convenience functions. If the users want to filter on, say healpix in cosmoDC2, they don't need to use `tract_filter`, and should just do `healpix_pixel == 1`, etc. 

It's true that in the future we may want to create a `healpix_filter` convenience function. To that end, I've created a new `base_filters` module that contains `partition_filter` and `sample_filter` without specifying the columns. However, I do not expect the end users to use these directly, it kind of defeats the purpose of using a convenience function if you have to specify the column name every time. 

The functions in `base_filters` are to be called by high-level functions (like the one in `tract_catalogs`). These high-level functions are written for specific catalog types (and hence the module name).

GCRCatalogs/helpers/base_filters.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+
+import numpy as np
+
+from GCR import GCRQuery
+
+__all__ = ["partition_filter", "sample_filter"]
+
+
+def partition_filter(name, ids, id_high=None):
```
JoanneBogart, line 13: Would like to see either a more descriptive first argument name (`column_name`? `partition_name`?) or something in the docstring about it, e.g
``` Returns a GCRQuery object to be used in the `native_filters` argument of get_quantities with native_filter *name*```

File: GCRCatalogs/helpers/base_filters.py [hunk 4]
```diff
@@ -0,0 +1,69 @@
...
+    """
+    # pylint: disable=no-member
+
+    frac = float(frac)
+    if frac > 1 or frac < 0:
+        raise ValueError("`frac` must be a float number in [0, 1].")
+    if frac == 1:
+        return GCRQuery()
+    if frac == 0:
+        return GCRQuery((lambda a: np.zeros_like(a, dtype=np.bool), "tract"))
```
JoanneBogart, line 57:
```suggestion
       return GCRQuery((lambda a: np.zeros_like(a, dtype=np.bool), ref_col_name))
```

JoanneBogart (changes requested):
I've suggested a couple minor changes.  Otherwise it looks fine.

GCRCatalogs/helpers/base_filters.py
yymao: Thanks for catching this. I've fixed it (didn't use the commit feature because it missed a space)

GCRCatalogs/helpers/base_filters.py
yymao: `partition_name` is better. Changed.

JoanneBogart (approved):
LGTM
Thanks, @yymao

Issue #501, opened 2020-12-05
A new reserved config keyword to indicate catalogs for public release


yymao:
We have defined a few [reserved keyword](https://github.com/LSSTDESC/gcr-catalogs/blob/master/CONTRIBUTING.md#reserved-keywords) in catalog configs to indicate various catalog status. One important function these keywords have is to allow `GCRCatalogs.get_available_catalogs` to hide certain catalogs (e.g., deprecated) from end users' view ([see implementation here](https://github.com/LSSTDESC/gcr-catalogs/blob/v1.0.4/GCRCatalogs/register.py#L439)).

We should add a new reserved config keyword to indicate catalogs that we make public available, so that when a user calls `GCRCatalogs.get_available_catalogs`, they will be able to filter out catalogs that are for DESC internal use only. 

This feature can be coupled with #499 --- if `DESC_GCR_SITE` is not set or not known, then `GCRCatalogs.get_available_catalogs` will only return public catalogs.

yymao:
I plan to work on this after #503 is merged.

PR #524, opened 2020-12-22
Add public-release catalog configs

### LINKED ISSUE #502: Add configs for catalogs that are part of DC2 public release
(Edited by @yymao Dec 22)

Catalogs in the public release will have different paths than those for the DESC internal use. Hence we will need to add new catalog configs for them. We need to add three configs:

- [x] DPDD-only object catalog
- [x] Truth-match catalog (with only unique truth entries)
- [x] Composite catalog of Object + Truth-match (as an add-on)

The DPDD-only object catalog catalog will use [dc2_object.DC2ObjectParquetCatalog](https://github.com/LSSTDESC/gcr-catalogs/blob/v1.0.4/GCRCatalogs/dc2_object.py#L656) as its reader with the reader option `is_dpdd` set to true to disable translation. 

The paths for these catalogs will be the ones we use for globus integration.

yymao:
This PR add the config files for the three catalogs (one is a composite) that we will include in the public release. This PR fixes #502. 

There are two things that may require some further discussion:

- Catalog names: The catalog names need to be different from the internal ones. Here I opted to use `desc_dc2_dr6` as the prefix for all public-release catalogs, and dropped `run2.2i` and `wfd` in the names. Other suggestions/opinions are welcome. 
- Relative paths: The relative paths are still subject to change depending on @heather999's Globus setup. 

In addition, I've added a `public_release` tag in anticipation of #501.


yymao:
Actually, one additional thing to consider here is that the current design would prevent one from access DESC-internal catalogs and public-release catalogs simultaneously during the same session, as `register.py` only allows one `root_dir` during one session. Hence, when `^/` is replaced with the public-release `root_dir`, all DESC-internal catalog configs won't work, and vice versa.

rmandelb:
I'm certainly in favor of dropping `run_2.2i` from catalog names that we provide publicly (it's DESC-internal jargon).  I'm less sure about dropping `wfd`.  If we can imagine in future serving some 5-year-depth DDF products, then having `wfd` now and `ddf` for the later ones would make sense to me.  And `wfd` vs `ddf` includes not just different observing conditions but also different populations of astrophysical objects in them, so I feel like the reminder in the name may be A Good Thing regardless?

heather999:
I think retaining as much useful information is helpful - so yes, include `wfd` or `ddf`, but skip `run2.2i`.  We may want to include some versioning which follows our own version conventions such as dr6-v1, to allow for the possibility that there is an update to the data.

yymao:
Based on this slack thread, I am adding `run2.2i` to the catalog names but not `wfd`. I realize this is somewhat in conflict with the opinions raised above. Further thoughts are welcome. 

As of now, the proposed names are:

- `desc_dc2_run2.2i_dr6_object`
- `desc_dc2_run2.2i_dr6_object_with_truth_match`
- `desc_dc2_run2.2i_dr6_truth`

The two main differences from our internal catalog names are:
1. The addition of `desc_` prefix
2. Run/DR go before catalog type

JoanneBogart:
Other than the catagories (`object_dpdd`, `truth_match`) I only see identifying information in the catalog names, not in the paths.
How is this supposed to work after we have another public release?   Will there be a separate Globus endpoint?  Do we expect non-DESC users to set up a separate local root_dir for each release?   That seems awkward.

heather999:
Let's review what we have.  The gsharing directory at NERSC contains:
```
/global/cfs/cdirs/lsst/gsharing/dc2_public/lsstdesc-dc2-wfd-dr6-v1> ls
object_dpdd/  truth_match/
```
We should go back over that naming and come up with something that is consistent to the naming mentioned in [Yao's comment above](https://github.com/LSSTDESC/gcr-catalogs/pull/524#issuecomment-750712092). 

The [draft portal](https://lsstdesc-portal.nersc.gov/), can download files and enforce the creation of a directory structure (You need to login with your Globus credentials, and then click Transfer)...and it does not, using the current names, but that needs to be changed.  Perhaps the area under `gsharing` could be changed to:
`lsstdesc/dc2-run2.2i-dr6-v1/object_dpdd`
`lsstdesc/dc2-run2.2i-dr6-v1/truth_match`
The portal's globus downloads will create that directory structure on the user's machine and they can define their ROOT_DIR to point to that `desc` directory. Would that work?  Do we need a deeper structure?  a subdirectory for dc2 to allow for a separate area for let's say DP0.1 and ultimately real data?

heather999:
Just a note, for now, transfers will only work for specific people I have granted read access to (right now that's Yao and Joanne) - for testing. Later today, I plan to set this up and allow public read access to allow everyone to test.

yymao:
When downloading files from the portal, the files will be put in like:

`<directory I chose>/lsstdesc-dc2-wfd-dr6-v1/object_dpdd/object_dpdd_tract2723.parquet`

So the question for the configs is whether we use `<directory I chose>` as the root dir, or something else.

heather999:
Well..  it may be a matter of opinion.  If users anticipate running GCRCatalogs for non DESC data as well as DESC data - then I can imagine using `<directory I choose>` for root dir is better.  We can dictate everything under `<directory I choose>` and I would still like to consider modifying the directory structure we provide for download so that it is more hierarchical - which we had originally, I just wasn't sure the downloads could replicate that structure.
So what I'd like to see might be `<directory I choose>/lsstdesc/dc2/run2.2i-dr6-v1` and we suggest users set their root dir to point to `<directory I choose>`

yymao:
`<directory I choose>/lsstdesc/dc2/run2.2i-dr6-v1` sounds fine, though it seems arbitrary that `dc2` is a subdirectory while `run2.2i` and `dr6` are not. So IMO `<directory I choose>/lsstdesc/dc2-run2.2i-dr6-v1` would work equally well. 

In either case, it seems the root dir the end users should set in GCRCatalogs would be `<directory I choose>/lsstdesc` rather than `<directory I choose>`. For example, `<directory I choose>` could be just their home directory, and it seems to make sense to use `<directory I choose>/lsstdesc` in GCRCatalogs?

JoanneBogart:
I prefer setting <directory I choose> as root dir.  For one thing, the instructions are simpler: people specify the same directory to the portal and to GCRCatalogs.   For another, it makes it possible for DESC people  to use the same root dir for public release data and other catalogs on their laptops if they want to. 
Maybe instead of lsstdesc that directory should be named something like lsstdesc-public?

yymao:
That sounds a reasonable proposal too. 

We should also keep in mind that some people may be accessing the files directly without GCRCatalogs. With that in mind I would vote for not making `dc2` a subdirectory.

yymao:
So we have settled on the following structures:

- `<directory_user_chooses>/lsstdesc/dc2-run2.2i-dr6-v1/object_dpdd`
- `<directory_user_chooses>/lsstdesc/dc2-run2.2i-dr6-v1/truth_match`

where `<directory_user_chooses>` will be the directory the user chooses during the Globus download step and it will also need to be set as the `root_dir` for GCRCatalogs.

yymao:
As for the catalog names (in GCRCatalogs), they will be:

- `desc_dc2_run2.2i_dr6_object`
- `desc_dc2_run2.2i_dr6_object_with_truth_match`
- `desc_dc2_run2.2i_dr6_truth`

Under the hood they are in fact aliases to 

- `desc_dc2_run2.2i_dr6_v1_object`
- `desc_dc2_run2.2i_dr6_v1_object_with_truth_match`
- `desc_dc2_run2.2i_dr6_v1_truth`

Please comment today if you have further thoughts on the catalog names. (cc @rmandelb: earlier you had comments on catalog names, so I'm pinging you to see if you have further comments before we finalize.)

rmandelb:
I have no further comments before you proceed to finalize this - thanks!

JoanneBogart (approved):
LGTM.   This is what we agreed upon.

PR #529, opened 2020-12-24
Add `public_release` keyword; adjust `get_available_catalogs` behavior accordingly

### LINKED ISSUE #501: A new reserved config keyword to indicate catalogs for public release

yymao:
This PR tries to fix #501. In particular it adds the following features:

- Recognize the `public_release` keyword in catalog configs.
- `get_available_catalogs` now accepts `is_public_release` argument; it has three operation modes:

  ```python
  # returns all catalogs that have `public_release`
  get_available_catalogs(is_public_release=True)
  
  # excludes all catalogs that have `public_release` 
  get_available_catalogs(is_public_release=False) 

  # returns all catalogs that have `public_release` and its value *contains* "v1"
  get_available_catalogs(is_public_release="v1")  
  ``` 

- Add `get_available_catalog_names`, which is a shorthand for `get_available_catalogs(names_only=True)`.

- If the user does not have `root_dir` set by site, `get_available_catalogs()` defaults to `get_available_catalogs(is_public_release=True)`. Otherwise, it defaults to `is_public_release=False`. Users can supply `is_public_release` keyword argument to overwrite.

In order to implement the last bullet point, I had to make some adjustment to `RootDirManager`. Specifically, I now distinguish:

```python
self._root_dir_from_runtime
self._root_dir_from_config
self._root_dir_from_site
```

This order is the order in which `root_dir` will be determined. In this way, we can use `self._root_dir_from_site` to determined if the user is a potential DESC member. A `has_root_dir_from_site` property was added `RootDirManager` for this purpose. 

During tests I notices a few places where an error may be raise due to None / empty string confusion. I've fixed those in this PR too. 

@JoanneBogart you may have some thoughts/opinions on the implementation details and I welcome your comments.

### DIFF
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -127,14 +127,18 @@ class RootDirManager:
         # User config has highest priority
         user_root_dir = self._user_config_manager.get(self._ROOT_DIR_KEY)
         if user_root_dir:
-            self._default_root_dir = user_root_dir
-            return
+            self._root_dir_from_config = user_root_dir
 
-        if self._site_config_path:
+        # Try to set self._root_dir_from_site
+        if self._site_config_path and os.path.isfile(self._site_config_path):
             self._site_config = load_yaml_local(self._site_config_path)
+
+        if self._site_config:
+            site_info = self._get_site_info()
+            if site_info:
             for k, v in self._site_config.items():
-                if k in self._site_info:
-                    self._default_root_dir = v
+                    if k in site_info:
+                        self._root_dir_from_site = v
                     break
 
     def _get_site_info(self):
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -245,7 +245,7 @@ class RootDirManager:
                     try:
                         resolved_path = os.path.join(self.root_dir, orig_path[len(self._ROOT_DIR_SIGNAL):])
                     except TypeError:
-                        warnings.warn("Root dir has not been set!")
+                        pass
                     else:
                         config_dict[k] = resolved_path
                 if record is not None:
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -214,11 +214,8 @@ class RootDirManager:
         """
         Remove root_dir item from user config.  root_dir for the current
         session is unchanged, however.
-        Returns
-        -------
-        old value (may be None)
         """
-        return self._user_config_manager.pop(self._ROOT_DIR_KEY, None)
+        self._user_config_manager.pop(self._ROOT_DIR_KEY, None)
 
     def reset_root_dir(self):
         self._custom_root_dir = None
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -734,7 +734,7 @@ def get_public_catalog_names(
     **kwargs
 ):
     """
-    Returns a list of all available catalog names.
+    Returns a list of names of all available catalog satisfying any constraints specified in parameters.
 
     Parameters
     ----------
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -740,7 +740,7 @@ def get_public_catalog_names(
     ----------
     include_default_only: bool, optional (default: True)
         When set to False, returned list will include catalogs that are not in the default listing
-        (i.e., those may not be suitable for general comsumption)
+        (i.e., those which may not be suitable for general consumption)
     name_startswith: str, optional (default: None)
         If set, only return catalogs whose name starts with *name_startswith*
     name_contains: str, optional (default: None)
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -711,7 +711,7 @@ def get_available_catalog_names(
     ----------
     include_default_only: bool, optional (default: True)
         When set to False, returned list will include catalogs that are not in the default listing
-        (i.e., those may not be suitable for general comsumption)
+        (i.e., those which may not be suitable for general consumption)
     name_startswith: str, optional (default: None)
         If set, only return catalogs whose name starts with *name_startswith*
     name_contains: str, optional (default: None)
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -428,7 +428,6 @@ class ConfigManager(Mapping):
     def __init__(self, config_dir):
         self._config_dir = config_dir
         self._configs = dict()
-        self.default_to_public_releases_only = False
         for config_file in sorted(os.listdir(self._config_dir)):
             config = Config(config_file, self._config_dir)
             if not config.ignore:
@@ -532,9 +531,7 @@ class ConfigManager(Mapping):
                 get_content = lambda config: (config.rootname, config.content)  # noqa: E731
 
         conditions = list()
-        if include_default_only and self.default_to_public_releases_only:
-            conditions.append(lambda config: config.is_public_release)
-        elif include_default_only:
+        if include_default_only:
             conditions.append(lambda config: config.is_default)
         if not include_addons:
             conditions.append(lambda config: not config.is_addon)
@@ -593,7 +590,6 @@ class ConfigRegister(RootDirManager, ConfigManager):
         RootDirManager.__init__(self, site_config_path, user_config_name)
         for config in self.configs:
             config.set_resolvers(self.resolve_reference, self.resolve_root_dir)
-        self.default_to_public_releases_only = not self.has_root_dir_from_site
 
     @property
     def root_dir(self):
@@ -687,6 +683,8 @@ def get_available_catalogs(
     """
     if "resolve_content" not in kwargs:
         kwargs["resolve_content"] = (not names_only)
+    if "is_public_release" not in kwargs and not _config_register.has_root_dir_from_site:
+        kwargs["is_public_release"] = True
     return _config_register.get_configs(
         names_only=names_only,
         include_default_only=include_default_only,
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -142,12 +142,15 @@ class RootDirManager:
         Return a string which, when executing at a recognized site with
         well-known name, will include the name for that site
         """
-        site_from_env = os.getenv(self._DESC_SITE_ENV)
+        site_from_env = os.getenv(self._DESC_SITE_ENV, "")
         site_from_socket = socket.getfqdn()
+
+        # hack for nersc batch nodes
+        if site_from_socket.startswith("nid") and site_from_socket[3:].isdigit():
+            site_from_socket += ".nersc.gov"
+
         if site_from_env:
-            if site_from_socket and site_from_env not in site_from_socket and not (
-                site_from_env == "nersc" and site_from_socket.startswith("nid")
-            ):
+            if site_from_socket and site_from_env not in site_from_socket:
                 warnings.warn("Site determined from env variable {} = {}, which differs from node name {}".format(
                     self._DESC_SITE_ENV, site_from_env, site_from_socket
                 ))
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -684,9 +684,9 @@ def get_available_catalogs(
         If set, only return catalogs whose name contains with *name_contains*
     """
     if not kwargs.get("is_public_release") and not _config_register.has_valid_root_dir_in_site_config:
-        warnings.warn("""It appears that you do not have access to DESC default root dir, or you are using a customized root dir.
+        warnings.warn("""It appears that you do not have access to the default root dir at a recognized DESC site, or you are using a customized root dir.
 As such, the returned catalogs may not all be available to you.
-Use get_public_catalog_names to see catalogs to see a list of catalogs from public releases.
+Use get_public_catalog_names to see a list of catalogs from public releases only.
 If you are a DESC member and believe you are getting this warning by mistake, please contact DESC help.""")
     kwargs.setdefault("resolve_content", (not names_only))
     return _config_register.get_configs(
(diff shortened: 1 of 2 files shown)


JoanneBogart:
@yymao I haven't looked at the implementation and so don't (yet) have opinions concerning most of what you propose. But from your description I worry that get_available_catalogs is getting rather complex.   I assume your goal is to make things easy for  non-DESC users - I'm all for that.  But I'm leery of your last point:

> If the user does not have root_dir set by site, get_available_catalogs() defaults to get_available_catalogs(is_public_release=True). Otherwise, it defaults to is_public_release=False. Users can supply is_public_release keyword argument to overwrite.

The complicated behavior could be annoying to DESC users who work both at a site and on laptops.   Instead how about providing a separate routine aimed at non-DESC users, say `get_released_catalogs(names_only=True, names_startswith=None, name_contains=None, **kwargs)` 
and maybe dispense with `get_available_catalog_names` ?  It might still be useful to  have an `is_public_release` option for `get_available_catalogs`, but with a default of False under all circumstances

JoanneBogart:
Looking into this a little more has confirmed my initial impression.  I don't think we want to use "root dir set by site" as a proxy for "is a DESC member"; it's not always accurate.  I think a clearer and more reliable approach is to have a separate, easy-to-use and document method for non-DESC users (or for anyone who prefers to look at just public release catalogs in a particular session).  If we proceed this way the changes to RootDirManager, which further complicate its behavior, would not be necessary.

yymao:
Providing a separate function is a good idea and I'll implement it. 

But we do have extensive public tutorials and README that use `get_available_catalogs`. Unless we'll edit all of those, I think `get_available_catalogs` should print out some sort of warnings for non-DESC members. In fact, I have been asked by non-DESC cosmoDC2 users why they can't load catalogs from `get_available_catalogs`. 

We could keep the behavior of `get_available_catalogs` the same but print out a warning for non-DESC members. 

I also agree that the proxy for "non-DESC members" is not great. Maybe a better way is to just test if the current `root_dir` is one of the paths in `site_configs`.

JoanneBogart:
> I also agree that the proxy for "non-DESC members" is not great. Maybe a better way is to just test if the current root_dir is one of the paths in site_configs.

I think this has the same problem: if someone is running on their laptop, this will not tell you if they're a DESC member or not.  All we can do is document - probably in several places - what definitely works for external users and what is "use at your own risk".   One place such information should go is the GCRCatalogs README.  I'd add a section immediately before or after the **NERSC Users** section with this kind of information.

yymao:
If someone is running on their laptop, then I think it makes sense that `get_available_catalogs` raises a warning saying not all catalogs that it returns would be available, no?

JoanneBogart:
> If someone is running on their laptop, then I think it makes sense that get_available_catalogs raises a warning saying not all catalogs that it returns would be available, no?

Perhaps; we've not had such a warning before.

> But we do have extensive public tutorials and README that use get_available_catalogs. Unless we'll edit all of those...

I think we should. Perhaps at the top of each publicly-available tutorial we add a sentence about intended audience and a disclaimer about  `get_available_catalogs` if it's used in the tutorial.

yymao:
@JoanneBogart I've reverted unused changes to `RootDirManager`, added `get_public_catalog_names`, and added a warning text in `get_available_catalogs`. Thoughts welcome.

JoanneBogart:
I'm not sure about the warning.  It's comprehensive, which is good, but that makes it quite long.  I'm aware of two situations where it might get triggered:
* non-DESC user who should really be calling get_public_catalog_names instead
* DESC user working with restricted set of catalogs on their laptop

In the latter case, often the warning isn't necessary because the user understands that only the catalogs they've downloaded are truly available.   Will it be annoying? Will they think they're being warned by mistake?  Hard to say.   What might be useful to such a user would be a method listing those catalogs which are installed, say 
list_accessible_catalogs(...)
which would behave mostly like get_available_catalogs but would only report those catalogs where `base_dir` (or similar keyword) resolved to an existing path that the user has access to.   But that could be involved to write.

The number of public GCRCatalogs routines is already quite large.  I would like to provide a documentation page including more or less the information in the docstrings for those routines.  Perhaps it could live in the repo and also be inserted into the Data Release Note?

yymao:
Actually, in GCRCatalogs v1.0.4, users who work on their laptops do receive a warning, but at GCRCatalogs import time. We have now suppressed the import time warning. So I am not sure if the current situation is more or less confusing. I think it's better in the sense that the warning is trigger at the time that it is relevant. But suggestions to how to improve the warning is welcome.

I agree that it'd be great to have a way to list actually available catalogs. I've been thinking about it, but it's not as simple as checking whether those directories exist (which is already a bit involved if we want to implement that). For most catalogs, the files in that directory need to match certain filename patterns. Also, would this new feature also check for completeness of catalog files (the set of tracts, for example)? If not, I worried that it might be more confusing than helpful. 

I think we should *not* write a page that mostly repeats the docstrings. This will result in high maintenance cost in the future as we would always need to update many places at once (and we are already doing that for things like schema tables, which I really don't think that's a good idea, but alas). 

But it's true we need some high-level documentation. This can show users examples and use cases, and refer them to the docstrings if they need details. I think this can go into the release note, and/or maybe a readme in the public release repo?

JoanneBogart:
I know it's a maintenance headache, but calling sequences and return values need to be documented.  Examples are extremely helpful when you're getting started, but after that a compact reference is much more useful.  I was thinking that a single GCRCatalogs tag would be associated with the public release so there would be no harm in copying the API documentation into the Note.  If that could get out of date, then a reference in the Note to a documentation file in the GCRCatalogs repo would do. 

If you don't mind, I'll make a separate PR to create such a file and assign myself to it.

yymao:
If existing docstrings are not clear, or if calling sequences and return values are missing, they should be expanded/added to the docstrings that live with the code. We can also add "Notes", "Examples", "See also" sections to the docstrings of public functions. (Note: for longer docstrings we should follow [Numpy's style](https://numpydoc.readthedocs.io/en/latest/format.html#sections).)

If we want to duplicate some parts of the docstrings in the release note that's fine by me, hopefully with integration with examples. What I don't think we should do is to create a separate docstring file that lives in this repo. We expect the users to know how to find docstrings.

GCRCatalogs/register.py
```diff
@@ -654,8 +683,12 @@ def get_available_catalogs(
     name_contains: str, optional (default: None)
         If set, only return catalogs whose name contains with *name_contains*
     """
-    if "resolve_content" not in kwargs:
-        kwargs["resolve_content"] = (not names_only)
+    if not kwargs.get("is_public_release") and not _config_register.has_valid_root_dir_in_site_config:
+        warnings.warn("""It appears that you do not have access to DESC default root dir, or you are using a customized root dir.
+As such, the returned catalogs may not all be available to you.
+Use get_public_catalog_names to see catalogs to see a list of catalogs from public releases.
+If you are a DESC member and believe you are getting this warning by mistake, please contact DESC help.""")
```
JoanneBogart, line 690:
```suggestion
       warnings.warn("""It appears that you do not have access to the default root dir at a recognized DESC site, or you are using a customized root dir.
As such, the returned catalogs may not all be available to you.
Use get_public_catalog_names  to see a list of catalogs from public releases only.
If you are a DESC member and believe you are getting this warning by mistake, please contact DESC help.""")
```

GCRCatalogs/register.py
```diff
@@ -665,6 +698,66 @@ def get_available_catalogs(
...
+    **kwargs
+):
+    """
+    Returns a list of all available catalog names.
+
+    Parameters
+    ----------
+    include_default_only: bool, optional (default: True)
+        When set to False, returned list will include catalogs that are not in the default listing
+        (i.e., those may not be suitable for general comsumption)
```
JoanneBogart, line 714:
```suggestion
       (i.e., those which may not be suitable for general comsumption)
```

GCRCatalogs/register.py
```diff
@@ -665,6 +698,66 @@ def get_available_catalogs(
...
+    name_startswith: str, optional (default: None)
+        If set, only return catalogs whose name starts with *name_startswith*
+    name_contains: str, optional (default: None)
+        If set, only return catalogs whose name contains with *name_contains*
```
JoanneBogart, line 718:
```suggestion
      If set, only return catalogs whose name contains  *name_contains*
```

GCRCatalogs/register.py
```diff
@@ -665,6 +698,66 @@ def get_available_catalogs(
...
+
+def get_public_catalog_names(
+    include_default_only=True,
+    name_startswith=None,
+    name_contains=None,
+    public_release_name=None,
+    **kwargs
+):
+    """
+    Returns a list of all available catalog names.
```
JoanneBogart, line 737:
```suggestion
  Returns a list of names of all available catalog satisfying any constraints specified in parameters
```

GCRCatalogs/register.py
```diff
@@ -665,6 +698,66 @@ def get_available_catalogs(
...
+
+    Parameters
+    ----------
+    include_default_only: bool, optional (default: True)
+        When set to False, returned list will include catalogs that are not in the default listing
+        (i.e., those may not be suitable for general comsumption)
```
JoanneBogart, line 743:
```suggestion
       (i.e., those which may not be suitable for general comsumption)
```

GCRCatalogs/register.py
```diff
@@ -665,6 +698,66 @@ def get_available_catalogs(
...
+    name_startswith: str, optional (default: None)
+        If set, only return catalogs whose name starts with *name_startswith*
+    name_contains: str, optional (default: None)
+        If set, only return catalogs whose name contains with *name_contains*
```
JoanneBogart, line 747:
```suggestion
      If set, only return catalogs whose name contains *name_contains*
```

JoanneBogart (dismissed):
I've made some minor suggestions to docstrings. Merge in if you see fit. Other than that, LGTM

JoanneBogart:
I guess we just have different styles (and I probably have a significantly worse memory).   I do end up consulting docstrings in source files when I have no other options and it's quite inefficient.  Once I've had some experience with a package, what is most useful to me is a list of public routines in some logical order (either grouped by topic or alphabetical by name).  For each routine there should typically be a sentence describing its purpose, minimal descriptions of arguments (including type information if there is any ambiguity) and returns. Anything which is already clear from well-chosen names does not need to be repeated.  In other words, content is just what is in a well-written docstring.  But I don't want to have to first find the source file and then hunt through it to find what I need.   Ideally one could at least partially automate the process of extracting this information from the source file.   I'm aware it's going to be a significant amount of effort but I think it's warranted for a package which has a wide user community.

yymao:
It seems we are talking about different types of documentation. Take Numpy as an example. There are high-level, descriptive notes like [this one](https://numpy.org/doc/stable/reference/arrays.ndarray.html) that give users overviews and general structures of routines/code. There's also lists of functions/methods, such as [this one](https://numpy.org/doc/stable/reference/routines.array-manipulation.html), grouped in some logical ways, and finally there's docstrings for individual functions like [this one](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html). 

All of them are useful and serve different purposes, and I was not saying that we only need docstrings. One of my points was that these different types of documentation should not be merely repeating each other, but reference each other. 

Now, there's a separate issue of limited human power that we are facing. If we can only maintain one type of documentation, I think we should do docstrings for two reasons: (1) Docstrings are more easily maintained, mostly because they got noticed during code revision, while a separated page can easily get outdated. (2) They are supported by standard tools so that users can easily browse them without looking into source code (e.g, `help(GCRCatalogs.get_available_catalogs)`, or, in Ipython and jupyter, simply `GCRCatalogs.get_available_catalogs?`). 

So, I think we should first make sure our docstrings are accurate and sufficient, and if we do have spare time, we can certainly add further documentation.

JoanneBogart (approved):
Looks fine

JoanneBogart:
I agree the top priority is good docstrings. Anything else should be derived from that.
> They are supported by standard tools ...
But I didn't know about them and I wouldn't be surprised if a significant fraction of our users don't either.  However, even so they don't quite serve the function I had in mind because you don't always know what you're looking for, and you might want to do your planning outside of a particular python, ipython or jupyter session.   What **would** serve, even if a bit crude, would be a web page which is a concatenation of that output for all public functions.  It might need some formatting, but the content would not have to be re-entered.   And it would allow someone to see the full scope of the capabilities of the package all at once in about a single page.

yymao:
> But I didn't know about them and I wouldn't be surprised if a significant fraction of our users don't either. 

Right, and our work here (in this `GCRCatalogs` repo) is not to teach users how to use standard tools. We can give people some pointers, but much would need to be done during discussion (among users or when users ask us questions). 

> What would serve, even if a bit crude, would be a web page which is a concatenation of that output for all public functions. 

I think this serves the need?

```python
help(GCRCatalogs.register)
```

But if we do want to create such a page, we should use existing tools to automatically generate one. For example, I made one for [https://yymao.github.io/generic-catalog-reader/index.html](GCR). The source of this doc site can be found [here](https://github.com/yymao/generic-catalog-reader/tree/master/docs-src). We can open another issue for this.

PR #530, opened 2020-12-24
Update schema/meta files

### LINKED ISSUE #514: Update SCHEMA.md
When writing the DESC Note for the Public Release, I noticed a few things that are inconsistent/missing in the current `SCHEMA.md` for the object catalog. The up-to-date version is being produced in the DESC Note, and we should pull that back into `SCHEMA.md` when ready.

yymao:
This PR fixes #514. During my attempt to update `SCHEMA.md`, I found more inconsistencies that I originally expected. I've tried to fix as much as I can in this PR. A few recurring issues are:

- `I{xx,xy,yy}` and `I{xx,xy,yy}PSF` all need a `_pixel` postfix, and their units are pixels.
- `psf_fwhm_pixels` should be called `psf_fwhm_pixel` to be consistent with above.
- There's no `ra_err` or `dec_err`
- `I_flag` should have no unit
- Repeated entries in meta files should be removed.

In addition, I noticed that `_dc2_metacal_meta.yaml` does not actually contain any information about the metacal catalog. I have removed it to avoid confusion. I also noticed some readers defined `SCHEMA_FILENAME` -- this is no longer used for any reader that is a subclass of `DC2DMCatalog` and should be removed.


wmwv (approved):
LGTM.

Thanks for taking the thorough and detailed trawl through all of this.

yymao:
Thanks, @wmwv

PR #532, opened 2020-12-29
Bug fix: allow non-string values in `info_dict`

yymao:
In the updated `_generate_info_dict` method, all values in the `info_dict` that contains `<band>` will be replaced by the actually band name. However, not all values in `info_dict` are string. Hence, a check should be done before invoking the `.replace` method. 

This bug was discovered when I ran the full test suite for the upcoming v1.1.0 tag.


JoanneBogart (approved):
LGTM

PR #533, opened 2021-01-01
Bug fix: `names_only` should be set to True in `get_*catalog_names`

yymao:
This PR fixes a bug in the newly introduced functions: `get_available_catalog_names` and `get_public_catalog_names`. These two functions are shorthands for `get_available_catalogs(names_only=True, ...)` and `get_available_catalogs(names_only=True, is_public_release=True)`. But somehow when I wrote these functions I mistakenly set `names_only=False`. This PR fixes the issue.


JoanneBogart (approved):
LGTM

yymao:
Thanks @JoanneBogart

PR #535, opened 2021-01-05
Add dc2_object_run2.2i_dr6_v2pre

yymao:
This PR adds the catalog config for `dc2_object_run2.2i_dr6_v2pre` and resolves #534.

PR #536, opened 2021-01-05
Tag v1.1.1

yymao:
We have just released v1.1.0 but we have made a bug fix (#533) and added DR6 v2pre (#535). It seems that remaining PRs for v1.2.0 (public release) may take longer to review, so I am proposing that we make a tag of v1.1.1 first.


JoanneBogart (approved):
There is now only one other PR outstanding, but if it seems useful to make this release now, it's ok with me.

yymao:
Let's see where we are after today's DM/DA telecon and then decide.

yymao:
Since we are push DR2 release to next week and the v1.2.0 tag (for public release) should take place this week, the v1.1.1 seems unnecessary. Closing this now.

PR #537, opened 2021-01-06
Tag v1.2.0 release candidate

yymao:
v1.2.0 should be tagged when we are ready for the public release. See also https://github.com/LSSTDESC/gcr-catalogs/milestone/11


yymao:
Actually, let's make a v1.2.0rc1 and invite DESC members to test things out.

JoanneBogart (approved):
LGTM

PR #538, opened 2021-01-07
Exclude public catalogs by default when calling get_available_catalogs

yymao:
When a user call `get_available_catalogs` or `get_available_catalog_names`, public catalogs will now be excluded by default, unless the user passes `include_public_release=True` keyword. 

The user can also pass `include_public_release_only=True` or `include_public_release_only="v1"`, which will return *only* the public catalogs. This has the same behavior as calling `get_public_catalog_names`.


JoanneBogart (approved):
LGTM

PR #540, opened 2021-01-08
Use find_packages in setup.py

### LINKED ISSUE #539: `GCRCatalogs.helpers` not importable in pip installation
The `GCRCatalogs.helpers` module is not importable in pip installation because it is not added to the `packages` option in `setup.py`.

Thanks @JoanneBogart for noticing this bug.

yymao:
This PR fixes #539. By using `find_packages`, all submodules will be automatically added during pip installation, including `GCRCatalogs.helpers`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ http://opensource.org/licenses/MIT
 """
 
 import os
-from setuptools import setup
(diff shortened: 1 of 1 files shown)


JoanneBogart (approved):
This should do the right thing

PR #541, opened 2021-01-08
Improve user config prompt

yymao:
When using `python -m GCRCatalogs.user_config set root_dir /path/`, the prompt would said `New value /path/ was set` which is a bit unclear since it is `root_dir` being set. This PR changes the prompt to `root_dir is now set to /path/`.


JoanneBogart (approved):
This is definitely an improvement.  Thanks @yymao

PR #543, opened 2021-01-09
Add cosmoDC2 public configs

yymao:
This PR adds public configs for the cosmoDC2 catalog. @heather999, can you verify that `catalog_root_dir` matches the endpoint setting? Thanks!


heather999 (approved):
Looks good and the copy to gsharing is going on right now!

PR #544, opened 2021-01-12
Update README.md to point to DESC data portal

yymao:
Update readme to point public users to the DESC data portal. 

Also, advice DESC members use the new `get_available_catalog_names(...)` instead of the lengthy `get_available_catalogs(names_only=True, ...)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ validation framework, and the concept and description of this reader interface
 can be found in the [DESCQA paper](https://doi.org/10.3847/1538-4365/aaa6c3)
 and also the [GCR repo](https://github.com/yymao/generic-catalog-reader).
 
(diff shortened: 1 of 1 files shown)


JoanneBogart (approved):
Looks good but maybe you could fix one more little thing in lines just before the ones you changed:   "comsumption"   should be "consumption"

yymao:
Thanks, @JoanneBogart. Fixed the typo. Can you give it a re-approval?

JoanneBogart (changes requested):
There's another one at line 28 (sorry I didn't make that clearer before).  Please fix (and I promise there aren't any others).

JoanneBogart (approved):
LGTM

PR #549, opened 2021-01-15
Allow cosmoDC2 reader to skip processing metadata of all files

yymao:
The cosmoDC2 reader (used for both cosmoDC2 and SkySim5000) processes metadata of all HDF5 files at load time. This was manageable given the smaller number of files for cosmoDC2, but it takes way too long for SkySim5000. 

This PR updates the reader's `_process_metadata` method such that, if `check_md5`, `check_size`, and `ensure_meta_consistent` (new option) are *all* set to `False` in the catalog config, the reader will only parse the metadata from *one* file and assume that all other files have consistent metadata instead of actually checking them. 

However, `_process_metadata` is also responsible for calculating the total sky area, which is obtained from the metadata of all individual files. In order to skip processing metadata of all files, we need to record the total sky area in the catalog config. 

In this PR, I have allowed both cosmoDC2 and SkySim5000 to skip these checks, and it significantly speeds up the initial load time of both cosmoDC2 and SkySim5000, especially the latter. 

We previously implemented these checks for the image simulation --- both to ensure the files are not tempered with and that the reader would raise errors as early as possible, instead of during the middle of a run. Now that the image simulation is done, turning off these checks may be ok? I want to flag this to @jchiang87 to see if he agrees. If not, I can turn back on the checks for cosmoDC2, only turning it off for SkySim5000.


jchiang87:
I think it should be fine to turn off the checking by default.  For the DC2 simulations, it was only relevant during the generation of the instance catalogs, which we are planning to discard.  For future image simulations, we'll use the new Sky Catalog scheme, but that's still in its early stages of development, so we might wait until that gets fleshed out further before deciding to re-enable these checks.   @JoanneBogart may have some thoughts about this.

evevkovacs (approved):
LGTM. I checked the numbers for the area and read through the code. I could not spot any problems.

yymao:
Thanks, @jchiang87 and @evevkovacs.

I'll see if @JoanneBogart has further comments before I merge.

JoanneBogart:
I think it's fine to turn off the checks.  If it seemed necessary to reinstate while generating Sky Catalogs we could always make another catalog based on the new catalog but overriding the values for the options.  Also the fact that these catalogs are now kept in a relatively protected and managed space makes tampering or incompleteness much less likely.

yymao:
Thanks, @JoanneBogart!

PR #548, opened 2021-01-15
add parquet versions of cosmoDC2 and SkySim5000

yymao:
This PR adds two catalog configs for the parquet versions of cosmoDC2 and SkySim5000:

- `cosmoDC2_v1.1.4_parquet.yaml`
- `skysim5000_v1.1.1_parquet.yaml`

Both configs use the newly added `CosmoDC2ParquetCatalog` reader class in `cosmodc2_parquet.py`.

The `CosmoDC2ParquetCatalog` is a subclass of the existing `DC2TruthParquetCatalog` in `dc2_truth_parquet.py`. The only addition to `CosmoDC2ParquetCatalog` is to process metadata present in the catalog configs.

Some code style adjustments were applied to `dc2_truth_parquet.py`, but the content of the code has not changed.

This PR fixes #546

PR #552, opened 2021-01-15
Bump version to v1.2.1

yymao:
In preparation for a wider SkySim5000 announcement, if we merged #548 and #549 to improve skysim5000 loading performance (either by skipping checks or by parquet), I think it's worth making a new release.


evevkovacs (approved):
Yes, this seems fine. It is a significant change and would alert people that there is an update.

PR #554, opened 2021-01-22
Add available_healpix_pixels to cosmoDC2 readers

### LINKED ISSUE #553: Allow users to obtain available healpix pixels for cosmoDC2/SkySim5000
Similar to Object Catalog, where an `available_tracts` property exists, we should add an `available_healpix_pixels` to cosmoDC2/SkySim5000 readers so that users can easily access the available healpix pixels of the catalog.

yymao:
This PR fixes #553, adding `available_healpix_pixels` property to the two cosmoDC2/SkySim5000 readers (one for hdf5 and one for parquet), so that they have the same public API for accessing available healpix pixels. 

Similar to `available_tracts` of the Object Catalog, `available_healpix_pixels` will return a list of integers that correspond to available healpix pixel IDs.

(cc @joezuntz)

### DIFF
--- a/GCRCatalogs/cosmodc2.py
+++ b/GCRCatalogs/cosmodc2.py
@@ -389,6 +389,12 @@ class CosmoDC2ParentClass(BaseGenericCatalog):
             return default
         return self._quantity_info.get(q_mod or quantity, default)
 
+    @property
+    def available_healpix_pixels(self):
+        if not self.lightcone:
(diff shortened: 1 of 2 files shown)


JoanneBogart (approved):
LGTM
The code looks fine. Just to be sure I tried it on both hdf5 and parquet cosmoDC2 catalogs.

PR #559, opened 2021-01-28
Update to v1.2.2

yymao:
In preparation of the DR6 v2 release, I am proposing a new v1.2.2 tag. Given that DR6 v2 is kind of like a "bug fix" of v1, I think increasing the patch version would be ok? If people feel we should increase the minor version instead, let me know.


heather999 (approved):
I think just increasing the patch is fine, esp for DR6-v2.  I'm wondering if adding the DR2 release warrant a new revision, since it moved from pre?

yymao:
We actually already include DR2 in the [v1.1.0 release](https://github.com/LSSTDESC/gcr-catalogs/releases/tag/v1.1.0).

heather999:
ok - carry on

heather999:
@yymao I think we're ready to go - it'd be good to get the new tag out this morning so we can proceed and update all the python environments and put out announcements concerning DR2 and DR6.  I'm ready to go ahead and mv symlinks at NERSC.

PR #567, opened 2021-02-23
Preparing v1.3.0

yymao:
Preparing a new tag.

I am debating between using v1.2.3 and v1.3.0. I started with v1.2.3 but maybe we should go with v1.3.0 instead.  Compared to v1.2.2, we fixed a bug and added Run 3.1i truth catalogs and Run 2.2i DR6 photo-z. According to https://semver.org/,  we should "increment the MINOR version when you add functionality in a backwards compatible manner".

Any preference regarding whether or not to increase minor version?


JoanneBogart:
I think increasing the minor version is appropriate.

JoanneBogart (approved):
LGTM

Issue #571, opened 2021-03-16
Need to change site_config to match the new catalog root directory at CC-IN2P3 


boutigny:
Following the introduction of a new CephFS file system at CC-IN2P3 some reorganization occurred and the GCRCatalogs root directory has been moved from /sps/lsst/data/desc/gcr_rootdir  to /sps/lsst/groups/desc/shared.
I am creating a PR to reflect this change in site_config

yymao:
Fixed by #572

PR #572, opened 2021-03-16
change catalogs root directory at CC-IN2P3

boutigny:
Fix issue #571


yymao (approved):
Thanks @boutigny. I didn't actually try it on CC but I believe you put down the right path :slightly_smiling_face:

PR #573, opened 2021-04-01
Add redmapper v0.7.5 on cosmodc2 1.1.4

erykoff:

### DIFF
--- a/GCRCatalogs/catalog_configs/cosmoDC2_v1.1.4_redmapper_v0.5.7.yaml
+++ b/GCRCatalogs/catalog_configs/cosmoDC2_v1.1.4_redmapper_v0.5.7.yaml
@@ -22,3 +22,4 @@ description: |
   (v0.5.7) run on cosmoDC2 v1.1.4.
 
 include_in_default_catalog_list: false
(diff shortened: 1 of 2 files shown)


GCRCatalogs/catalog_configs/cosmoDC2_v1.1.4_redmapper_v0.5.7.yaml
```diff
@@ -21,4 +21,4 @@ description: |
   Cluster and member catalogs computed from the redMaPPer cluster finder
   (v0.5.7) run on cosmoDC2 v1.1.4.
 
-include_in_default_catalog_list: true
+include_in_default_catalog_list: false
```
yymao, line 24: Should this catalog also be deprecated? If so, add the "deprecated" flag:

```suggestion
include_in_default_catalog_list: false
deprecated: Use cosmoDC2_v1.1.4_redmapper_v0.7.5 instead
```

yymao (changes requested):
@erykoff thanks! I have a question/suggestion, otherwise it looks good.

yymao (approved):
Thanks, Eli

PR #575, opened 2021-04-07
Update desc-env-test GH Action

### LINKED ISSUE #574: Modify desc-env-test.yml to avoid use of explicit python version
As we move away from python 3.7 to python 3.8, the installation directory has been updated in the desc-python Docker image to avoid explicitly including the python version.  Now the install is in `/usr/local/py`.  Updating the GH Action accordingly.

heather999:
Fix #574 
The desc-python image now installs the conda environment in `/usr/local/py` and avoids explicitly mentioning the python version.  This will allow easier upgrades in the future.


yymao (approved):
LGTM, thanks!

PR #576, opened 2021-04-08
Update version to v1.3.2

yymao:
Oops, I forgot to increase version in the v1.3.1 tag. 

Move to v1.3.2 to avoid inconsistency between code and package versions.


heather999:
Seems fine - I'm just a little confused though. I see v1.3.1 was tagged a few minutes ago. How is v1.3.2 different?

heather999 (approved):
ah now I see.. but could this not still be v1.3.1 and retagged?  Either way!

yymao:
It's the same -- I forgot to update `version.py` so in the v1.3.1 package, the code version is still v1.3.0.  Here v1.3.2 is to fix that. 

Shouldn't have done things quickly..

yymao:
I don't think we can "re-tag"? (Well, it makes sense that git doesn't allow that.)

heather999:
It is possible delete a git tag and retag - at least I've done that in the past in GitHub directly.  But truly, whatever works.  v1.3.2 is out the door!  :)

PR #578, opened 2021-05-26
add desc_dc2_run2.2i_dr6_v2* catalogs; update desc_dc2_run2.2i_dr6* a…

JoanneBogart:
…liases to point to v2

GCRCatalogs updates for public release of run2.2i_dr6_v2 data.
See https://github.com/LSSTDESC/desc-data-portal/issues/75 second item.


JoanneBogart:
@yymao what should the new version be?   1.3.3 ?

yymao (changes requested):
Thanks @JoanneBogart -- changes look good to me but we should also deprecate `desc_dc2_run2.2i_dr6_v1_*` -- can you add the `deprecated` flag to those configs?

JoanneBogart:
Ok.  I assume the message should direct people to the generic names, the ones not including v2.
And I will set version to 1.3.3
Please let me know if you want me to do something different.

JoanneBogart:
@yymao would you like to take another look?

yymao (approved):
Oops, sorry, missed the notification. LGTM.

PR #582, opened 2021-07-14
Update to v1.3.4

yymao:
I am proposing a new release so that the updated redmapper catalog can be available on DESC environments.


JoanneBogart (approved):
A one-character change!  Looks fine.

PR #583, opened 2021-08-06
Update redMaPPer catalog for cosmoDC2 and add run2.2i DC2 DR6 catalog.

erykoff:


yymao (changes requested):
Thanks @erykoff.  LGTM. Can you update the README as well?

yymao (approved):
Thanks for the update and sorry for the delay!

PR #587, opened 2022-03-03
1. Add catalog configs for star truth files with int id

### LINKED ISSUE #586: Add new DC2 truth files to NERSC shared area and Data Portal
New versions of three star truth parquet files have been made with `id` field an int rather than a string.  Also the summary table has been trimmed (similar to variability and lc_stats tables) to exclude stars outside the DC2 footprint. We need configs for these files in both installations.

JoanneBogart:
2. Add several truth files for public release
3. Update version: fields for public release files to include all
   release note versions which are applicable
4. Update creator: fields for several catalog configs

fix #586

### DIFF
--- a/GCRCatalogs/catalog_configs/dc2_truth_run2.2i_star_truth_summary.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_truth_run2.2i_star_truth_summary.yaml
@@ -2,7 +2,7 @@
 
 subclass_name: dc2_truth_parquet.DC2TruthParquetCatalog
 base_dir: ^/DC2-prod/Run2.2i/truth/startruth
-filename_pattern: 'star_truth_summary_trimmed.parquet$'
-description: DC2 Run 2.2i  Summary truth table for stars, trimmed to DC2 footprint. For full version see dc2_truth_run2.2i_star_truth_summary_full
+filename_pattern: 'star_truth_summary.parquet$'
+description: 'DC2 Run 2.2i  Summary truth table for stars, extends
+  beyond DC2 footprint.  See also dc2_truth_run2.2i_star_truth_summary.'
 creators: ['DESC DC2 Team', 'Joanne Bogart']
-include_in_default_catalog_list: true
--- a/GCRCatalogs/catalog_configs/dc2_truth_run2.2i_star_truth_summary_full.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_truth_run2.2i_star_truth_summary_full.yaml
@@ -1,8 +1,0 @@
-#  Use ^/ to indicate file path relative to GCR root dir
-
-subclass_name: dc2_truth_parquet.DC2TruthParquetCatalog
-base_dir: ^/DC2-prod/Run2.2i/truth/startruth
-filename_pattern: 'star_truth_summary.parquet$'
-description: 'DC2 Run 2.2i  Summary truth table for stars, extends
-  beyond DC2 footprint.  See also dc2_truth_run2.2i_star_truth_summary.'
-creators: ['DESC DC2 Team', 'Joanne Bogart']


JoanneBogart:
There are still a couple open questions:
* Do the new catalog configs for the Portal describe an acceptable file hierarchy or should more be lumped together?
* Should old (string `id` field) star truth catalogs be marked as deprecated?

JoanneBogart:
@yymao thanks for checking.  I thought about both of the things you mentioned and couldn't quite make up my mind so left things alone when I made the pull request.

I'll go ahead and change the single entries to lists.  I'm still in a quandary about which of the star versions to use.   For someone just looking at stars, the int ID versions are preferable, but for someone looking at SNe as well, maybe not. Do you have an opinion?  And if I take one or the other set out of the default list, should those catalogs also be deprecated?

I've loaded the catalogs for the new entries under the DESC shared area and made queries successfully, but can't try the new `desc_...` entries until the files are copied.

yymao:
Maybe keep the string ones in the default list? I don't feel strongly. In the truth-match catalog and truth ID is still string I think. 

Regardless which one we choose, we don't need to deprecate the other, as it is still a valid catalog that serves a slightly different purpose. We just don't need to recommend it for new users.

JoanneBogart:
@yymao  your suggestion (to take ..int_id configs out of default list) is fine with me.  I've done that, but realized that a better default version for summary_truth is one that is trimmed to the DC2 footprint.  There already was such a file with string `id` field in the shared area but there was no config for it, so I made one and tested that the catalog loads correctly.  That new config is included in the latest commit.

yymao:
Thanks @JoanneBogart -- there's an inconsistency in the naming that I want to double check:

- `dc2_truth_run2.2i_star_lc_stats` is using the trimmed file but doesn't have "trimmed" in the catalog name.
- `dc2_truth_run2.2i_star_truth_summary_trimmed.yaml` is using the trimmed file but has "trimmed" in the catalog name.
- `dc2_truth_run2.2i_star_variability_truth.yaml` is using the non-trimmed file(?). 

Should we just always use trimmed files, but drop "trimmed" in the catalog name (since there's already "dc2" in the name)?

JoanneBogart:
@yymao  the naming inconsistency is historical and, I'm afraid, to some extent unavoidable.  When Jim made the original sqlite files he did not append `trimmed` in the variability file, even though it was trimmed.  He _did_ include it it for the lc_stats file.  The star summary file he created wasn't trimmed. I carried his names over (for file names) when I made the parquet files, but I dropped the `trimmed` from the catalog config for lc_stats since there was no need for it.   Then I made a trimmed version of the summary parquet file.  I need to include `trimmed` (or something) in its name to distinguish it from the untrimmed version.

yymao:
Right, thanks for the background. 

Obviously changing the filenames now is not a great idea. But I think updating the "catalog names" would make sense.  My suggestion would be:

- `dc2_truth_run2.2i_star_lc_stats` --> points to the trimmed version
- `dc2_truth_run2.2i_star_truth_summary` --> points to the trimmed version
- `dc2_truth_run2.2i_star_variability_truth` --> points to the trimmed version (ever though the filename doesn't have "trimmed"; we should add a comment in the yaml file for future reference)

We can then also create (optionally):

- `dc2_truth_run2.2i_star_lc_stats_full` --> points to the untrimmed version
- `dc2_truth_run2.2i_star_truth_summary_full` --> points to the untrimmed version

I think this would be more clear to the end users?

JoanneBogart:
There is no untrimmed lc_stats file.  Only truth summary has an untrimmed version and I'm loathe to get rid of it entirely because up until this coming release it's been the only parquet version of star truth summary.
I already have text like this in the description fields of most of the `star` files where it applies:
"...for stars restricted to the DC2 footprint".   I missed a couple; I'll fix that.
We could rename the configs for truth summary as you suggest, but that does mean that a config file which used to point to one thing would now point to something different.  Is that a problem?

yymao:
I think that'd be ok. We should not get rid of the untrimmed version of summary though. We should add a note in the `description` field stating that if people need the full version please switch to `dc2_truth_run2.2i_star_truth_summary_full.` This way, in the unlikely event that this change breaks anyone's code, it'd be a very easy fix. 

And in this case, `trimmed` can also be removed from the catalog names for catalogs with int IDs.

JoanneBogart:
The only catalog name containing `trimmed` is `dc2_truth_run2.2i_star_truth_summary_trimmed.yaml`

yymao:
@JoanneBogart I just started to work on the tutorial and notice something else. For the other existing public catalogs, we have created a "version-less" alias for each of them. Here's an example:

https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/desc_dc2_run2.2i_dr6_object.yaml

Should we do the same for the newly added truth information?

JoanneBogart:
@yymao would you like to take one last look?

yymao:
Yes, I am testing this branch in my tutorial directly.

GCRCatalogs/catalog_configs/desc_dc2_run2.2i_star_lc_stats_int_id.yaml
```diff
@@ -0,0 +1,8 @@
+#  Use ^/ to indicate file path relative to GCR root dir
+
+subclass_name: dc2_truth_parquet.DC2TruthParquetCatalog
+base_dir: ^/lsstdesc-public/dc2/run2.2i-dr6-v2/truth_variable
+filename_pattern: 'star_lc_stats_trimmed_int_id.parquet$'
```
yymao, line 5:
```suggestion
filename_pattern: 'star_lc_stats_int_id.parquet$'
```

GCRCatalogs/catalog_configs/desc_dc2_run2.2i_star_truth_summary_int_id.yaml
```diff
@@ -0,0 +1,8 @@
+#  Use ^/ to indicate file path relative to GCR root dir
+
+subclass_name: dc2_truth_parquet.DC2TruthParquetCatalog
+base_dir: ^/lsstdesc-public/dc2/run2.2i-dr6-v2/truth_variable
+filename_pattern: 'star_truth_summary_trimmed_int_id.parquet$'
```
yymao, line 5:
```suggestion
filename_pattern: 'star_truth_summary_int_id.parquet$'
```

yymao (changes requested):

JoanneBogart:
@yymao concerning the alias question -
Somewhere towards the beginning of this project (maybe a Slack thread, maybe a Gibhub issues; I haven't been able to find it) I listed three ways to address the organization of the configs for the new catalogs and using aliases was number 3.  My feeling at the time was that the extra machinery was not necessary in the case of the truth catalogs and no one else dissented.  I still think that's the case.  I don't object to going this route, but I'm not eager to implement it myself.   Overall, I think we need to stop tweaking and go with what we've got.

yymao:
@JoanneBogart --- hmm, I am not sure if we are talking about the same thing (but maybe I misunderstood the original slack thread). 

Our current convention is that we have a versioned config, and a "not versioned" config that is just an alias to the version one. This way, when we update to a new version, people who use the not versioned" config will get the update automatically. 

For example, `desc_dc2_run2.2i_dr6_object` is an alias for `desc_dc2_run2.2i_dr6_v2_object`. 

So I was asking if we should do the same for the truth catalogs.

yymao:
Re: merging this -- we should first resolve #589.

yymao:
@JoanneBogart For now, I think we can keep the updates to `creators` and to `public_release` versions, and remove the new catalog configs, and merge this PR? Once we have the new files we can create a PR to add them?

JoanneBogart:
That sounds right.   I took a quick look. There was one change of (data file) filename for [dc2_truth_run2.2i_star_truth_summary.yaml](https://github.com/LSSTDESC/gcr-catalogs/pull/587/files#diff-93b66adf22f094980681d474bd396a785a73f213cfaf4bf8b741f105c565ae8f)(deleted a "trimmed" I think).  Feel free to do with it whatever seems appropriate.

JoanneBogart:
@yymao please take a look when you get the chance.  All configs for the new files in the shared area are in the last commit as well as the non-alias files for the Portal.

yymao:
@JoanneBogart thanks for updating this PR. The newly added configs from your most recent commits 8d1810c and 4d28e36 look good. But I have a question about what we should do with some old configs. 

For example, there's a bunch configs with names starting with `dc2_truth_run2.2i` -- they are not consistent with the new naming scheme and in fact have an older version of schema. Should we deprecate all of those? 

There are also configs with names starting with `desc_dc2_run2.2i_star` and `desc_dc2_run2.2i_sn` -- Should these be removed?

JoanneBogart:
@yymao I haven't had a chance to look yet.  When I do, I suspect I will agree with you that they should be deprecated.
However, would it be ok if we separate that question from this PR?

yymao:
Is there a reason to delay the deprecation of those catalogs?  If so then yes we can separate those from this PR.

The public catalog configs (with `desc_` prefix) in this PR still need to be made consistent with the new tuth catalogs though.

JoanneBogart:
I would like to merge in what we've got, then create another PR (or more if necessary) for any remaining work needed for the public release. It's becoming too complicated to remember and keep track of what's in this PR.
 
 I don't know what you mean by 

> The public catalog configs (with desc_ prefix) in this PR still need to be made consistent with the new tuth catalogs though.
Are you talking about the new ones I made with `desc_` prefix or something else?  Whatever it is, I would prefer to deal with it somewhere other than this PR, which needs to get out the door.

yymao:
These should be removed, right? 

- GCRCatalogs/catalog_configs/desc_dc2_run2.2i_sn_truth_summary.yaml
- GCRCatalogs/catalog_configs/desc_dc2_run2.2i_sn_variability_truth.yaml
- GCRCatalogs/catalog_configs/desc_dc2_run2.2i_star_lc_stats_int_id.yaml
- GCRCatalogs/catalog_configs/desc_dc2_run2.2i_star_truth_summary_int_id.yaml
- GCRCatalogs/catalog_configs/desc_dc2_run2.2i_star_variability_truth_int_id.yaml

JoanneBogart:
I thought we had already gotten rid of those.  Yes, they should be removed.

JoanneBogart:
@yymao The `desc_dc2_...` configs you mentioned above are now removed.

`dc2_truth_run2.2i_star_lc_stats_int_id.yaml`
`dc2_truth_run2.2i_star_truth_summary_int_id.yaml`
`dc2_truth_run2.2i_star_variability_truth_int_id.yaml`

should also go, shouldn't they?  And at some point the files they point to should be removed from the shared area since I believe those configs were only on the development branch.

yymao:
Yes. Those should be removed too. Good catch. 

How about `dc2_truth_run2.2i_star_truth_summary_full.yaml`?  Should we rename it back to `dc2_truth_run2.2i_star_truth_summary.yaml`?

JoanneBogart:
> How about dc2_truth_run2.2i_star_truth_summary_full.yaml? Should we rename it back to dc2_truth_run2.2i_star_truth_summary.yaml?

...because the parquet file the current `dc2_truth_run2.2i_star_truth_summary.yaml` is pointing to (`star_truth_summary_trimmed.parquet`) has been superseded by the newer files?  Maybe.  Or else just remove `dc2_truth_run2.2i_star_truth_summary_full.yaml`.   In either case it should be deprecated.

yymao:
Right, I was thinking that, instead of change the existing config, we can just keep it as is and then deprecate it.

yymao:
Suggestion: Remove `dc2_truth_run2.2i_star_truth_summary.yaml` and then rename `dc2_truth_run2.2i_star_truth_summary_full.yaml` back to `dc2_truth_run2.2i_star_truth_summary.yaml`

yymao (approved):

PR #588, opened 2022-03-10
Add "nersc_public" as a site

yymao:
This allows users on nersc to run 

```python
GCRCatalogs.set_root_dir_by_site("nersc_public")
```
to switch to the public released root dir.

### DIFF
--- a/GCRCatalogs/site_config/site_rootdir.yaml
+++ b/GCRCatalogs/site_config/site_rootdir.yaml
@@ -1,2 +1,3 @@
 nersc: /global/cfs/cdirs/lsst/shared
 in2p3: /sps/lsst/groups/desc/shared
+nersc_public: /global/cfs/cdirs/lsst/gsharing


GCRCatalogs/site_config/site_rootdir.yaml
```diff
@@ -1,2 +1,3 @@
 nersc: /global/cfs/cdirs/lsst/shared
 in2p3: /sps/lsst/groups/desc/shared
+nersc_public: /global/cfs/cdirs/lsst/gsharing
```
heather999, line 3: I guess this is similar to pointing to `shared` in that we're not pointing to the underlying subdirectories but I'm wondering if in this case it should go all the way down to `/global/cfs/cdirs/lsst/gsharing/lsstdesc-public`

yymao: I don't think we can. This path acts like the folder a non-nersc user choose to transfer catalogs from Globus, and Globus creates the `lsstdesc-public` folder. That's why this path is one level up.

JoanneBogart: And we've set it up so that the relative paths in catalog configs look like `^/lsstdesc-public..` so I think what Yao has is right: `lsstdesc-public` should not be included in the definition of the root dir.

JoanneBogart (approved):
Looks fine

PR #589, opened 2022-03-10
Suggestions for #587

yymao:
Here's some suggestions for #587 (It's easier to use a PR to list suggestions). 

1. `star_lc_stats` can be combined with `star_truth_summary` using composite catalog now that they have the same order. The user then doesn't need to know which catalog to load since GCRCatalogs will decide for the user based on the quantities requested. 
2. Remove `int_id` since there's no other options in the public release catalogs. 
3. Rename `variability_truth` to `truth_variability` -- I don't feel very strongly about this and we can revert it back. It looks nicer to be aligned with `truth_summary`, but it also creates misalignment between the filename and catalog name, and also with the internal catalog name... (Lesson here is that we need some overall guidelines for naming that can we check against everytime we create new catalogs..)


JoanneBogart:
The composite catalog is a good idea; let's do it.
I'd prefer to keep the int_id.  It does serve a purpose: only these three catalogs do in fact have an integer id. The SNe catalogs and truth match do not.
I also prefer to leave the other names as is, annoying as they are.  If we change only the catalog names of the public release files and not the NERSC-only ones it will be confusing.   But the NERSC ones have been around for a while; I wouldn't want to change those.   Also there is to my ear a difference between "truth variability" and "variability truth".   The first implies that the truth is varying; the second is talking about the truth of the (source's) variability.   Similarly, "truth summary" means "summary of the truth", which is what we want.

yymao:
I'm ok with not renaming `variability_truth` to `truth_variability` -- I can revert it back in this branch.

I am less certain about including `int_id` in the public catalog names. I don't see how that helps the general users. But I don't want do a fight here. 

I think we need a naming guideline that we can stick to. Otherwise we'll likely need to fight about these things every time we add some new catalogs.

JoanneBogart:
> I am less certain about including int_id in the public catalog names. I don't see how that helps the general users. 

For anyone who uses other truth catalogs (SNe or truth match) it makes it clear that these catalogs are different, a subtle difference which they might not otherwise be aware of, but which could cause them grief.

I strongly agree we need naming guidelines and am quite prepared to spend some time and effort on helping to generte them, but *after* we get this release out the door, please.

yymao:
I am afraid that I disagree with the reasoning -- I don't think we should use filename to highlight all the subtle difference, especially in the case where we didn't provide the alternative.. 

That said, I agree that we should get this out and I'll concede. I've revert the `vaiability_truth` and `int_id` in my last commits. Please merge at your convenience and we can go back to #587

GCRCatalogs/catalog_configs/desc_dc2_run2.2i_star_truth_summary_int_id.yaml
```diff
@@ -1,8 +1,25 @@
 #  Use ^/ to indicate file path relative to GCR root dir
 
-subclass_name: dc2_truth_parquet.DC2TruthParquetCatalog
-base_dir: ^/lsstdesc-public/dc2/run2.2i-dr6-v2/truth_variable
-filename_pattern: 'star_truth_summary_int_id.parquet$'
+subclass_name: composite.CompositeReader
```
JoanneBogart, line 3: There should be something in the config name to indicate stats are include. Could be, e.g., `desc_dc2_run2.2i_star_truth_summary_with_stats_int_id.yaml` or `desc_dc2_run2.2i_star_truth_summary_int_id_with_addons.yaml`. I'm open to other suggestions.

JoanneBogart (changes requested):
This will be ready to go if we can agree on a suitable name can be found for this config.

GCRCatalogs/catalog_configs/desc_dc2_run2.2i_star_truth_summary_int_id.yaml
yymao: Right, again, we need the naming guideline yesterday.  Here we are not providing a "not with addon" option, and this catalog will not result in any confusion if people thought it doesn't have the addon columns... 

So I don't think we with those additions in the name.  But if you want to make a decision here, then I'll follow that decision.

JoanneBogart: The reason I think this is important, even though it makes a long unwieldy name even longer, is that `truth_summary` tables have a well-defined structure https://docs.google.com/document/d/1PnFZqdMmHBk6ePEvZBdKVOt64vwQ1cse2SaSJTVx7GM/edit  which we've respected up to this point.

yymao: Ok in the to-be-written naming guidelines we should also specify how to move forward when people can't reach an agreement 😂

I think we have different philosophies for what catalog names' main purpose is, so it might be difficult for us to agree here.

If we were to add some new columns to an existing catalog, I don't think we will need to change the catalog name, and I think the same logic applies here. Yes, the truth summary has a specific schema, but I don't see why we can't add a few new columns to it and say that those columns are only available for stars, not SNe. To me, changing the catalog name does not help the users, as in both cases they still need to read the release notes to see what's available.

But again, I think our debate here is not likely to be fruitful. Instead of trying to convince each other, one of us can make a decision. Since you are both the person who prepared the catalog and also the co-convener, I think it makes sense for you to make the decision here and I'll just follow it. 

If you want me to make the decision instead, I can do that too. Just let me know.

JoanneBogart: Sorry, I didn't see your last post until now.   
Probably this whole pull request and #587 also have been superseded by events.

yymao: This one is certainly superseded. We won't need composite catalog with the new proposal. Some of the #587 should be kept -- I'll comment there.

yymao:
Obsoleted.

PR #591, opened 2022-04-14
really deprecate this time

JoanneBogart:
This change was intended to be part of the previous PR

### DIFF
--- a/GCRCatalogs/catalog_configs/dc2_run2.2i_truth_galaxy_summary.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_run2.2i_truth_galaxy_summary.yaml
@@ -1,0 +1,2 @@
+alias: dc2_run2.2i_truth_galaxy_summary_v1-0-0.yaml
+include_in_default_catalog_list: true
--- a/GCRCatalogs/catalog_configs/dc2_run2.2i_truth_galaxy_summary_v1-0-0.yaml
+++ b/GCRCatalogs/catalog_configs/dc2_run2.2i_truth_galaxy_summary_v1-0-0.yaml
@@ -1,8 +1,7 @@
 #  Use ^/ to indicate file path relative to GCR root dir
 
 subclass_name: dc2_truth_parquet.DC2TruthParquetCatalog
-base_dir: ^/DC2-prod/Run2.2i/truth/galtruth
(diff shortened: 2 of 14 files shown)


yymao:
Thanks @JoanneBogart -- should we just use this PR to deprecate other truth catalogs that we plan to deprecate? Seems a bit strange to just deprecate this one. 

We should also edit the README to point to the new set of truth catalogs.

JoanneBogart:
@yymao Yes, it would be fine to deprecate whatever else needs deprecating.  Do you have a list?

yymao:
These can be deprecated now:
- dc2_truth_run2.2i_sn_truth_summary.yaml
- dc2_truth_run2.2i_sn_variability_truth.yaml
- dc2_truth_run2.2i_star_lc_stats.yaml
- dc2_truth_run2.2i_star_truth_summary.yaml
- dc2_truth_run2.2i_star_variability_truth.yaml

This should be deprecated when the new galaxy truth table is made:
- dc2_truth_run2.2i_galaxy_truth_summary.yaml

This should be deprecated when the new tract-partitioned merged truth table is made:
- dc2_truth_run2.2i_summary_tract_partition.yaml

JoanneBogart:
The SN and star ones are taken care of.  I could deal with the galaxies or, since that involves adding new files as well as deprecating, that could be in a separate PR.

yymao:
Thank you @JoanneBogart!  We can move galaxy truth, raw truth, and truth match in other PRs indeed.

Can you update the listing in the README file in this PR as well?

JoanneBogart:
I went ahead and added configs for galaxies in this PR.  I also changed portal directory name from `truth_variable` to `truth_unmerged` (maybe `truth_raw` would be better?) because it will include the galaxy truth.  README.md informatio is updated for all the new catalogs to date (everything but truth match).  I'm not sure what to do about the file SCHEMA.md, in particular the subsection "Quantities belonging to all truth summary catalogs".  Preferably, we either go into detail in the package documentation or in the Release Note, but not both.

yymao (approved):
Thanks @JoanneBogart. Looks good to me.  Approving this now. 

I just realize we also need to update the schema, but we can open another PR
https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/SCHEMA.md

PR #594, opened 2022-05-25
Update GH Action for new desc-python:bleed

### LINKED ISSUE #592: docker build test failed due to missing desc env
docker build test now fails due to missing desc env, see
https://github.com/LSSTDESC/gcr-catalogs/runs/6135638876?check_suite_focus=true#step:11:11

Is this due to a rename of the environment in the docker image? I wonder if @heather999 would know.

heather999:
Modified installation directory in desc-python:bleed and now using conda base env

Fix #592


yymao (approved):
Thanks @heather999

PR #597, opened 2022-06-07
Bug fixes to newly added configs

yymao:
I am testing the newly added configs and this PR will contain fixes to bugs that I found.


JoanneBogart (approved):
Yes, this looks right. Thanks!

PR #600, opened 2022-08-31
Nersc ci

stuartmcalpine:
**Added a github actions workflow (nersc-ci.yml):**

This triggers the NERSC CI pipeline. At the moment this is just triggered manually, however you can change this to on push, schedule, etc.

**Added quick catalog test for ci (./tests/test_catalogs_ci-py):**

This gets run with the --cori-ci flag for pytest. It just reads a few value from the cosmoDC2_v1.1.4_small, however you can change this to what you need.
 
**Added a gitlab workflow (.gitlab-ci.yml):**

This is what gets run on Cori, at the moment it just installs gcr-catalogs and runs pytest on the new test file

I don't have permission to run workflows on this repo (I think). The repos the nersc-ci.yml file points to are my NERSC GitLab repos. You'll have to make repos in your NERSC instance and change the values in the nersc-ci.yml file.  Setup for repos is in the docs for CI repo (https://desc-continuous-integration.readthedocs.io/en/latest/desc/ci_at_nersc.html)

### DIFF
--- a/.github/workflows/nersc_ci.yml
+++ b/.github/workflows/nersc_ci.yml
@@ -4,21 +4,21 @@ on: [workflow_dispatch]
 
 env:
   # URL to source repository on GitHub.
-  GITHUB_SOURCE_REPO: https://github.com/LSSTDESC/<source-repo>.git
+  GITHUB_SOURCE_REPO: https://github.com/LSSTDESC/gcr-catalogs.git
   # Branch to work with.
-  GITHUB_SOURCE_BRANCH: main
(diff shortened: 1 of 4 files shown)


stuartmcalpine:
The CI now gets done on Cori and reported back here (see the "NERSC" status above).

The files are ready to be merged in, but the branch name in the workflow will have to be changed from nersc_ci to master. And the NERSC repos will have to be changed to your ones. 

You may also want to add any additional tests to the test_catalogs_ci.py file.

JoanneBogart:
I think events have passed this PR by.

PR #601, opened 2022-09-26
Add WaZP cosmoDC2 v1.1.4 cluster catalogs

m-aguena:
Add WaZP catalogs.


yymao (changes requested):
Thanks, @m-aguena. There are two things -- 

The catalog name `cosmoDC2_v1.1.4_image_with_photozs_flexzboost_v1_wazp_v1.0` should be changed to something like `cosmoDC2_v1.1.4_wazp_flexzboost_v1.0`

The prefix `cosmoDC2_v1.1.4_image_with` is reserved for composite catalogs that include all the information of `cosmoDC2_v1.1.4_image`, which I don't think it's the case here? 

The `### ALL WaZP columns ###` section should be removed from the python file. I understand it's convenient to have that, but it really shouldn't be here. If you want to keep a list of columns, you can create a meta yaml file (like [this one](https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/_dc2_truth_match_meta.yaml))

m-aguena:
Thanks for the review @yymao.

I have updated the names of the wazp catalogs, removed the commented text and created a meta file

README.md
```diff
@@ -68,8 +68,10 @@ Confluence page (*DESC member only*).
 - `cosmoDC2_v1.1.4_redmapper_v0.8.1`: Redmapper catalog (v0.8.1) for `cosmoDC2_v1.1.4_image` (provided by Eli Rykoff).
 - `cosmoDC2_v1.1.4_redmagic_v0.8.1_highdens`: Redmagic catalog (v0.8.1) for `cosmoDC2_v1.1.4_image` for the high density sample (provided by Eli Rykoff).
 - `cosmoDC2_v1.1.4_redmagic_v0.8.1_highlum`: Redmagic catalog (v0.8.1) for `cosmoDC2_v1.1.4_image` for the high luminosity sample (provided by Eli Rykoff).
+- `cosmoDC2_v1.1.4_wazp_v1.0_truez`: WaZP catalog (v1.0) for `cosmoDC2_v1.1.4` using true redshifts (provided by Michel Aguena).
 - `cosmoDC2_v1.1.4_image_with_photozs_v1` and `cosmoDC2_v1.1.4_small_with_photozs_v1`: containing photo-z for cosmoDC2 v1.1.4 (provided by Sam Schmidt)
 - `cosmoDC2_v1.1.4_image_with_photoz_calib` and `cosmoDC2_v1.1.4_small_with_photoz_calib`: containing columns that identify DESI-like QSOs, LRGs, ELGs, or a magnitude limited sample in cosmoDC2 v1.1.4 (provided by Chris Morrison)
+- `cosmoDC2_v1.1.4_wazp_v1.0_flexzboost_v1`: WaZP catalog (v1.0) for `cosmoDC2_v1.1.4_image_with_photozs_flexzboost_v1` using FlexZBoost redshifts (provided by Michel Aguena).
```
yymao, line 74: Move this line to right below `cosmoDC2_v1.1.4_wazp_v1.0_truez`.

Also, is the additional `_v1` at the end intentional? If so, that's ok. Just checking to be sure.

yymao (changes requested):
One minor change request on the README.

README.md
m-aguena: Yes, `_v1`  is there to match the version of the flexzboost catalog

yymao (approved):

PR #603, opened 2022-10-03
Update for new version of SkySim5000 with some bug fixes. 

evevkovacs:


yymao (changes requested):
Thanks @evevkovacs -- can you add `skysim5000_v1.1.2` to README? (Generally speaking, all catalogs with `include_in_default_catalog_list: true` should appear in README).

evevkovacs:
I updated the README to include the new catalog with the caveat that it is only available for 0<z<1

yymao (approved):
Thanks @evevkovacs.

Issue #604, opened 2022-10-26
desc_dc2_run2.2i_dr6_v4_object pointing to v2 directory


nevencaplar:
I am trying to run the tutorial notebook https://github.com/LSSTDESC/desc-data-portal/blob/main/notebooks/dc2/object_table_gcrcatalogs_intro.ipynb

My data, freshly downloaded, gets placed in `/Users/nevencaplar/lsstdesc-public/dc2/run2.2i-dr6-v4/object_dpdd`. 

I am unable to access it via the suggested command `print(GCRCatalogs.load_catalog('desc_dc2_run2.2i_dr6_object').available_tracts)` as that point to v2 object, i.e., fails with `ValueError: 'base_dir' /Users/nevencaplar/lsstdesc-public/dc2/run2.2i-dr6-v2/object_dpdd is not a valid directory`.

After some googling, I try `print(GCRCatalogs.load_catalog('desc_dc2_run2.2i_dr6_v4_object').available_tracts)`, but that is still pointing to v2 directory, i.e., fails with the same error. 
 
Some more brief investigation perhaps suggest that the issue might be that [lsstdesc-public/dc2/run2.2i-dr6-v2/object_dpdd ](https://github.com/LSSTDESC/gcr-catalogs/blob/9335c78dc71ef8f548c89bdb19871b32cfadef26/GCRCatalogs/catalog_configs/desc_dc2_run2.2i_dr6_v4_object.yaml#L1) is still connected to v2?

PR #605, opened 2022-10-27
Point desc_dc2_run2.2i_dr6_v4_object to v4 directory

yymao:
As reported #604, `desc_dc2_run2.2i_dr6_v4_object` was an alias of `desc_dc2_run2.2i_dr6_v2_object`. Indeed these two catalogs are identical; however, for public users, if they downloaded the v4 object catalogs, the newly downloaded files will be stored in a directory with `v4` in its path, not `v2`. As a result, `desc_dc2_run2.2i_dr6_v4_object` cannot be set as an alias.


JoanneBogart:
Just to be sure I understand - the issue is that the downloading of, e.g., v4, is done by physical directory, so the file being aliased to would not be downloaded.  Is that it?

yymao:
The physical files are not aliased and will be downloaded from the portal, but the downloaded files will be stored under `^/lsstdesc-public/dc2/run2.2i-dr6-v4/object_dpdd` (notice the `v4` in the path). 

The original config uses `alias` so its path will be the same as `v2`, but in reality the path needs to has `v4`, and that's the issue here. 

I just made another commit to switch to `based_on`, which allows us to use the v2 config but just update the path.

JoanneBogart (approved):
LGTM

PR #606, opened 2022-10-27
Prepare v1.4.2 release

yymao:
Even though we have only made a small change (#605), I think it's worth making a new release because that was a bug fix, and the fixed bug affects public users who want to use our v4 public data set.


JoanneBogart (approved):
I agree; it merits a new version tag.  Looks good.

PR #613, opened 2022-12-21
Use builtin types in place of numpy aliases

### LINKED ISSUE #612: Numpy bool is deprecated
`np.bool`, which is being used in a lot of places in the code, has been deprecated since  version 1.20.0. They recommend to just use `bool`.

yymao:
This PR fixes #612.  

This PR replaces numpy builtin type aliases (e.g. `np.bool`) with builtin types (e.g. `bool`), because the former has been deprecated since numpy 1.20.  See "[Using the aliases of builtin types like np.int is deprecated](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated)" for detail.


m-aguena (approved):
It all looks good

JoanneBogart (approved):
LGTM!

PR #616, opened 2023-01-21
Check env variable NERSC_HOST to determine site

### LINKED ISSUE #610: Automatic detection of site fails for Perlmutter
Existing code tries to determine if it's running at NERSC by looking for a particular pattern in the node name. This worked for cori but Perlmutter node names are undistinguished (e.g. `login34`) and don't satisfy the pattern used. One can establish site manually by means of the environment variable `DESC_GCR_SITE`. gcr-catalogs will use this value but still puts out a warning since the bogus "site" determined by the code doesn't match the value in the environment variable.       Since our standard kernels at NERSC set `DESC_GCR_SITE` appropriately, most users don't need to be concerned with it beyond knowing to ignore the warning.  
I propose we eliminate the code which tries to deduce site or, at the very least, eliminate the warning message.

JoanneBogart:
fix #610

### DIFF
--- a/GCRCatalogs/register.py
+++ b/GCRCatalogs/register.py
@@ -148,13 +148,13 @@ class RootDirManager:
         well-known name, will include the name for that site
         """
         site_from_env = os.getenv(self._DESC_SITE_ENV, "")
+        if site_from_env:
+            return site_from_env
+
         if  os.getenv("NERSC_HOST", ""):
             site_from_node = 'nersc'
         else:
             site_from_node = None
-
-        if site_from_env:
-            return site_from_env
         return site_from_node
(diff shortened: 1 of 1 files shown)


yymao (approved):
Thanks!  

One minor comment: I think Lines 156-157 can be moved to right after 150, given that these two methods are not disentangled?

JoanneBogart:
You're saying that, when the user has set the site environment variable it takes priority. Therefore there is no point in bothering with evaluating site_from_node if the site environment variable has been set.  Agreed!  I will make this change.

yymao (approved):
LGTM!

PR #608, opened 2022-11-28
Changes for switching to treecorr convention for position_angle

evevkovacs:
Legacy quantities have been renamed with _dc2. position_angle_true now uses morphology/position_angle and flips the sign to be consistent with the treecorr convention.
ellipticity_1* and ellipticity_2* now use the supplied values of morphology/position_angle and convert to the treecorrr convention before computing the components.

PR #622, opened 2023-03-01
Changes and config files for skysim_v3.1

evevkovacs:
Here are the changes to cosmodc2.py and the necessary config files to make the new skysim_v3.1 available to the collaboration


yymao:
@evevkovacs thanks for the PR and sorry for the delay. The config addition looks good. Can you update the readme accordingly?  

Did you also add a new reader class `SkySimGalaxyCatalog`? Can you document the difference between `SkySimGalaxyCatalog` and the existing `SkySim5000GalaxyCatalog`?

Also, I noticed that the changes we made in #608 have been reverted back in this PR? I don't think that's intentional?

evevkovacs:
Hmm, regarding the changes for #608, no that wasn't intentional, but I am a bit puzzled because I pulled from the latest gcr-catalogs before adding my changes.  Let me look into this.
 I did add a new class. I will add the documentation for the differences.  The new class is appropriate for the new generation of mocks. It seemed better to start a fresh class than try to massage the old one, since the underlying native quantities are quite different. Maybe I should change the name to avoid confusion? We could call it something like "DiffSkyGalaxyCatalog".  What do you think?

yymao:
Changing the class name sounds a good idea.  I am not sure if DiffSky refers to some specific features of this catalog?

evevkovacs:
@yymao Sorry for the long delay time in replying.  I have added a couple of sentences of documentation. Please let me know if you would like more. I changed the new class name to DiffSky, because our new catalogs will most likely be called that. I also reinstated the position_angle changes. That was my oversight, caused by copying an older version of cosmodc2.py into the repository. If you are satisfied with the documentation, it is ready to merge. Thanks.

evevkovacs:
@yymao I also just added the configuration files and updated README for the bug-fixed version of skysim5000_v1.2, which I uploaded for the collaboration meeting. Patricia and I had a miscommunication about the version number, so internally it is called v1.2  Hence, I needed to update the configuration files to make it consistent. I decided to add the changes to this branch because there were also additions to the README for v3.1 and I didn't want any of those to get lost.

GCRCatalogs/cosmodc2.py
```diff
@@ -430,7 +432,7 @@ def _generate_quantity_modifiers(self):
             'magnification': (lambda mag: np.where(mag < 0.2, 1.0, mag), 'magnification'),
             'halo_id':       'uniqueHaloID',
             'halo_mass':     (lambda x: x/self.cosmology.h, 'hostHaloMass'),
-            'is_central':    (lambda x: x.astype(bool), 'isCentral'),
+            'is_central':    (lambda x: x.astype(np.bool), 'isCentral'),
```
yymao, line 435: `np.bool` has been deprecated
```suggestion
            'is_central':    (lambda x: x.astype(bool), 'isCentral'),
```

README.md
```diff
@@ -50,11 +50,19 @@ Confluence page (*DESC member only*).
 
 ### Extragalactic Catalogs and Add-ons
 
+#### SkySim/DiffSky
+*by Andrew Hearin, Eve Kovacs, Patricia Larsen, Esteban Rangel, Katrin Heitmann et al.*
+
+- `skysim5000_v3.1`: prototype of new SkySim catalog, produced with differentiable, forward modeling techniques, sky area (~400 sq. deg.)
```
yymao, line 56: I believe the catalog name is `skysim_v3.1` and there's no 5000. Suggest to also make it clear that this is not SkySim5000. 

```suggestion
- `skysim_v3.1`: prototype of new SkySim (also known as DiffSky) catalog, produced with differentiable, forward modeling techniques, sky area (~400 sq. deg.). Not to be confused with SkySim5000.
```

GCRCatalogs/cosmodc2.py
```diff
@@ -600,6 +602,74 @@ def _generate_quantity_modifiers(self):
         return quantity_modifiers
 
 
+class DiffSkyGalaxyCatalog(CosmoDC2ParentClass):
+    """
+    DiffSky galaxy catalog reader, inherited from CosmoDC2ParentClass
+    Class for new generation of catalogs generated with JAX-based 
+    forward modeling techniques.
```
yymao, line 609: Suggest adding what catalogs use this reader
```suggestion
    forward modeling techniques.
    This reader is used by SkySim v3+ (not SkySim5000) catalogs.
```

yymao (changes requested):
Thanks @evevkovacs and sorry for the delay on my part too. I've added a few suggestions to make it more clear.

evevkovacs:
@yymao Thanks for the suggestions. I added very similar comments. Thanks for catching my (bad) typo of leaving 5000 in the catalog name. Everything should be addressed now.

yymao (approved):
Thanks, @evevkovacs!

PR #623, opened 2023-05-05
Prepare release 1.6.0

yymao:
We should make a new release so that recent merged PRs can propagate to our shared environments.


evevkovacs:
@yymao @JoanneBogart I just went to fix the bug in cosmodc2 that was reported in dc2_users regarding the use of np.bool.  But it has already been fixed! Thanks to whoever did that.

yymao:
Yes, it has been fixed but not released! So I wanted to get this out.

JoanneBogart (approved):
LGTM

PR #625, opened 2023-09-26
Diffsky v0.1

evevkovacs:
Two catalog configuration files for test catalogs for the roman-desc sims.
The diffsky catalog is stored in the CS directory. It is not critical for this catalog to be generally available, but the special named version of the cosmoDC2 catalog should be pushed up to the main branch for more general use
cosmodc2.py has also been modified for the diffsky catalog and this should also go in the main branch.


yymao:
@evevkovacs thanks. 

Two questions: 
1. Are we sure we don't want to copy the data to the general shared place?  I'm slightly concerned that we are adding catalogs to the registry that don't live in the general shared place. Pining also @heather999 @JoanneBogart about this.
2. Wouldn't the edits in the reader break backward compatibility?

JoanneBogart:
The understanding has been that everything with a config file in the master branch is kept in the shared area. For one thing, this makes it easy to maintain the catalogs at another installation.  Is there a reason _not_ to copy the data there in this case?

evevkovacs:
I will copy to the shared directory (if I can, or ping Heather) and update the configuration file. The diffsky catalogs are evolving rapidly, so this one will soon become obsolete, but that happened with cosmoDC2 also.

evevkovacs:
Oops, I forgot that I had made those edits in the reader. I will have to edit those and make a version check or do something similar to ensure that I am not breaking the reader for the earlier catalogs.
When perlmutter comes back up I will make those changes, test and push the new version.

evevkovacs:
@yymao Here is the updated PR for the roman-rubin reader. I copied the latest catalog to the shared area, deleted the previous obsolete versions and updated the cosmodc2 reader so that it works on the new catalog and is backwards compatible.

GCRCatalogs/catalog_configs/diffsky_v0.1.1_p39991_elais.yaml
```diff
@@ -0,0 +1,32 @@
+#  Use ^/ to indicate file path relative to GCR root dir
+
+subclass_name: cosmodc2.DiffSkyGalaxyCatalog
+
+catalog_root_dir:  /global/cfs/cdirs/lsst/groups/CS/diffsky_v0.1.1/pscan_39991_nersc
```
yymao, line 5: The catalog data should be moved to the shared space.

GCRCatalogs/catalog_configs/diffsky_v0.1.1_p39991_image.yaml
```diff
@@ -0,0 +1,39 @@
+#  Use ^/ to indicate file path relative to GCR root dir
+
+subclass_name: cosmodc2.DiffSkyGalaxyCatalog
+
+catalog_root_dir:  /global/cfs/cdirs/lsst/groups/CS/diffsky_v0.1.1/pscan_39991
```
yymao, line 5: The catalog data should be moved to the shared space.

GCRCatalogs/catalog_configs/roman_rubin_2023_v1.1.1_elais_unlensed.yaml
```diff
@@ -0,0 +1,32 @@
+#  Use ^/ to indicate file path relative to GCR root dir
+
+subclass_name: cosmodc2.DiffSkyGalaxyCatalog
+
+catalog_root_dir:  /global/cfs/cdirs/lsst/groups/CS/roman_rubin_2023_v1.1.1_unlensed
```
yymao, line 5: The catalog data should be moved to the shared space.

GCRCatalogs/cosmodc2.py
```diff
@@ -436,8 +436,8 @@ def _generate_quantity_modifiers(self):
             'stellar_mass':  'totalMassStellar',
             'stellar_mass_disk':        'diskMassStellar',
             'stellar_mass_bulge':       'spheroidMassStellar',
-            'size_disk_true':           'morphology/diskMajorAxisArcsec',
-            'size_bulge_true':          'morphology/spheroidMajorAxisArcsec',
+            'size_disk_true':           'morphology/diskHalfLightRadiusArcsec',
+            'size_bulge_true':          'morphology/spheroidHalfLightRadiusArcsec',
```
yymao, line 440: Are these backward compatible?  How about the older files that still use `morphology/diskMajorAxisArcsec` and `morphology/spheroidMajorAxisArcsec`

yymao (commented):
Thanks, @evevkovacs. Two questions: 

1. There are still a few configs that point to files not in the shared space. Do you intend to add those?
2. It's unclear to me that the change to size definition in the cosmoDC2 reader would be backward compatible. Can you elaborate how it is backward compatible?

GCRCatalogs/cosmodc2.py
evevkovacs: I decided to remove the 2 configuration files that corresponded to catalogs that were not in the shared space. 
The changes for size_disk_true and size_bulge_true are actually a bug fix, although due to the bug, the numerical value will not change. Let me explain: previously diskMajorAxisArcsec and spheroidMajorAxisArcsec were set to the half light radius, which was incorrect. ie the major axes and the sizes were all set to the same value.  The size should have been set to the half light radius.  This PR corrects that bug.  The native major and minor axes are still incorrect, but that is another issue that needs to be more widely discussed.  I hope that clarifies the situation.  Please let me know if you have any questions.

yymao: Thanks for the explanation. That sounds alright and we should just include a brief summary of this bug fix in the upcoming release so that people are aware of it.

yymao:
@evevkovacs -- are you planning to remove the configs that have files not in the shared space?

yymao (approved):
Thanks @evevkovacs

PR #628, opened 2023-11-28
Update version to v1.7.0

yymao:
@evevkovacs -- we should make a release with the new catalogs that you added in. Are the two new catalogs meant for general use? If so, I think we can bump the version to v1.7.0. But if they are just for testing purpose, v1.6.1 may be a better version. 

In the case of general use, we should also add them to README.


evevkovacs:
roman_rubin_v1.1.1_elais is for general use. It is being used for the roman-desc image sims. 
cosmoDC2_v1.1.4_test_roman_desc was just a test catalog that was used by the image sim folks to compare with the new catalogs.  It was only for their test area and does not overlap with the area used for the final image simulations. We could actually remove this catalog.  I can update the README  in this branch and delete that config file.  Please let me know if I should go ahead with that.

yymao:
Sounds good. Thank you!

evevkovacs:
OK, I made those updates.

yymao:
Thanks @evevkovacs  -- I noticed that you removed `skysim_v3.1` from README, but `skysim_v3.1` is still available in catalog config. It the intention here to discourage people from using `skysim_v3.1`?

yymao:
Based on @evevkovacs's comment, let's bump this version to v1.7.0.  

Once we sort out the `skysim_v3.1` question we can merge this and tag v1.7.0

yymao:
Got you. In this case, for the 3 skysim configs, we should remove 
```yaml
include_in_default_catalog_list: true
```

and also add 
```yaml
deprecated: Use roman_rubin_2023_v1.1.1_elais instead.
```

evevkovacs:
I just pushed the changes to the skysim_v3.1.yaml config file. I also changed the roman_rubin config file to inclue the catalog in the default list. I had to do a git merge to push my changes. I hope all is well.

yymao:
Thanks @evevkovacs  -- how about `skysim_v3.1.0_image` and `skysim_v3.1.0_small`? Should these be deprecated as well?

yymao:
OK I've added the deprecation flags. @evevkovacs can you take a look and approve this PR if they look good?

evevkovacs (approved):

evevkovacs (approved):

PR #629, opened 2023-12-01
Remove tables from dependencies 

yymao:
`tables` (or `pytables`) is not actually used in `GCRCatalogs` and should be removed from dependencies. 

Also updated version to 1.7.1 for a new tag.


JoanneBogart (approved):
I agree; it doesn't appear to be used.   Let's get rid of it!
However CI is failing so maybe some further changes are needed.

yymao:
ah, apparently `tables` is needed by `pd.read_hdf`.

PR #630, opened 2024-01-22
Config files for new catalogs

evevkovacs:
Here are the configuration files for the catalog used for the roman-desc simulations  (v_1.1.2) and a newer variant with improved model parameters tuned against Cosmos2020 data.
These catalogs are in the shared catalogs area and the configuration files have been tested with DESCQA. See [ dn/dmag test](https://portal.nersc.gov/cfs/lsst/descqa/v2/?run=2024-01-19)


yymao (changes requested):
Thanks @evevkovacs. If `roman_rubin_v1.1.1_elais` is deprecated (as indicated by README), can you add the `deprecated` tag to its config file?

evevkovacs:
@yymao I added the deprecated tag as requested and updated the information in the skysim_v3.1* yaml files to point to the newer roman_rubin catalog.

yymao (approved):
Thanks!

JoanneBogart:
Is it ok to merge, now that it's approved?

PR #636, opened 2024-06-18
Extend astropy's FlatLambdaCDM

### LINKED ISSUE #635: Incompatibility with astropy 6.1.0
The code in the cosmodc2 reader which attempts to add attributes to the object returned by `astropy.cosmology.FlatLambdaCDM` fails with astropy 6.1.0 (the version in current LSST Pipelines builds) because that object is not modifiable. In particular, attributes for quantities `n_s` and `sigma8` cannot be added.  If there are users needing those quantities, they should be made accessible some other way.

yymao:
This PR introduces a new subclass from astropy's FlatLambdaCDM. The new subclass is identical to FlatLambdaCDM except that it can accept additional attributes, maintaining the original behavior. 

Fix #635

### DIFF
--- a/GCRCatalogs/alphaq.py
+++ b/GCRCatalogs/alphaq.py
@@ -8,8 +8,8 @@ import warnings
 from distutils.version import StrictVersion # pylint: disable=no-name-in-module,import-error
 import numpy as np
 import h5py
-from astropy.cosmology import FlatLambdaCDM
 from GCR import BaseGenericCatalog
+from .cosmology import FlatLambdaCDM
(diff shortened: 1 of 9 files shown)


JoanneBogart (approved):
Thanks, Yao!   This looks fine and I confirmed it does fix the problem I saw.
Perhaps the CI should be updated to exercise Python 3.11.
I see there is some apparently unrelated problem causing the 3.10 build to fail.

yymao:
Thanks @JoanneBogart. Going to ignore the CI for now. Need to fix that indeed..

PR #638, opened 2024-08-13
Integrate use of dataregistry into gcr-catalogs

### LINKED ISSUE #637: Make use of dataregistry when present and requested
Traditionally gcr-catalogs has used the collection of config files to determine which catalogs are available and how to read them. It should also be possible to register  catalogs in the dataregistry, where "register" includes saving the content of the config files in the dataregistry database.  Then users with access to the dataregistry would be able to use it to find and read catalogs.

JoanneBogart:
If the dataregistry module is available in the user's environment, they can choose to use it to find and load catalogs instead of the traditional config file mechanism.

fix #637

### DIFF
--- a/GCRCatalogs/__init__.py
+++ b/GCRCatalogs/__init__.py
@@ -2,4 +2,5 @@ import GCR
 from GCR import *
 
 from .register import *
+from .base_config import *
 from .version import __version__
--- a/GCRCatalogs/base_config.py
+++ b/GCRCatalogs/base_config.py
@@ -22,6 +22,13 @@ class BaseConfig(Mapping):
     ADDON_KEY = "addon_for"
     PUBLIC_RELEASE_KEY = "public_release"
 
+    def __init__(self, name=None, rootname=None):
+        self.name = name
+        self.rootname = rootname
+        self._resolved_content_ = None
+        self._resolvers = None
+        self._content_ = None
+
     def set_resolvers(self, *resolvers):
         """
         Each resolver will be called with two arguments:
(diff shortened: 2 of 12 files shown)


yymao (approved):
Thanks for preparing this @JoanneBogart.  I think the proposed change makes sense, and the high-level implementation (such as creating `base_config`) is very reasonable, so I am approving this PR. However, I have **not** read the code line-by-line (especially the new `dr_register.py`) and have **not** tested this PR either. If a detailed review is needed, I can help but it will need to wait until I have more time to attend to it.

JoanneBogart:
Thank you, @yymao !
One change I made after the discussion some weeks ago was to automatically activate the cfg-file mode of access in case the dataregistry code is not available. This makes it possible to at least test that, in this case, everything behaves as expected for existing applications. If you have code lying around which exercises anything beyond the CI tests and which involves minimal effort on your part, please run it and let me know if you see any glitches.
 However users who _can_ successfully import dataregistry will have to make a call to  `set_config_source` before they can do anything else.

yymao:
I don't have other tests beyond the existing CI tests. 

I wonder if it's useful to add a CI test to check the functionality of `dr_register.py`?

JoanneBogart:
> I wonder if it's useful to add a CI test to check the functionality of dr_register.py?

It would be useful but would take some work since the CI workflow has no access to NERSC resources, including the data registry Postgres database.  We would have to create a database as part of the workflow and add entries.  This kind of thing is done for the dataregistry CI.  
I have a test script which I used to debug after seeding a Postgres database at NERSC with some entries.     
dataregistry does also support sqlite to some degree.  I might be able to use that for a more self-contained test.

PR #639, opened 2024-09-13
Allow user to select config source  via environment variable

JoanneBogart:
Allow user to specify config source via environment variable  `GCR_CONFIG_SOURCE` (acceptable values are `dataregistry` or `files`) so that they don't have to modify code to make the choice.  
Also use value of `GCR_DR_SCHEMA` if set to specify dataregistry schema to use; otherwise default to "production".


yymao (approved):
LGTM. thanks!

JoanneBogart:
Thank you, @yymao, for attending to this so quickly!

PR #645, opened 2025-04-15
updates for compatibility with changed DataRegistry API, making use o…

JoanneBogart:
The API for connecting to the dataregistry has changed.  This PR updates gcr-catalogs to be compatible with the new protocol.


GCRCatalogs/dr_register.py
```diff
@@ -157,7 +158,10 @@ class DrConfigRegister(RootDirManager, DrConfigManager):
         def __init__(self, site_config_path=None, user_config_name=None,
                      dr_root=None, dr_schema=None, dr_site=None,
                      owner_type=None, owner=None):
-            if dr_schema.find("production") > -1:
+            if dr_schema is None:
+                owner_type = "production"
+                owner = "production"
+            elif dr_schema.find("production") > -1:
```
yymao, line 164: Is there a reason that these two conditions are not combined in one if statement? 
```suggestion
            if dr_schema is None or dr_schema.find("production") > -1:
```

yymao (commented):
Thanks @JoanneBogart.  I have one minor comment inline.

Given that there are compatibility-breaking changes (although not user facing), we should at least increment the minor version number.

GCRCatalogs/dr_register.py
JoanneBogart: If `dr_schema` is None the second condition would throw an exception if executed first.   I was being cautious. In a compiled language it's possible that the order of execution could change but, after some googling, I'm more or less convinced that's not a concern for Python. I will accept your suggestion.

JoanneBogart:
@yymao I agree about the version change, just wasn't sure at what stage this should be done.
Do I just change the value in `version.py` to, e.g., 1.10.0, or is there more to it than that?

yymao (approved):
Thanks @JoanneBogart . LGTM

